PR #8, opened 2015-02-02
Enhanced checkpointing for MCSCF

wirawan0:
Qiming,

I created an enhanced checkpointing for MCSCF calculations so we could have a copy of the last CI vector at the end of a CASSCF macro iteration. It is not enabled by default; to use it, one must pass `dump_chk_ci=True' to the CASSCF/CASCI kernel() function or mc1step() method. Right now, only writing is supported, I did not implement the reading yet.

Comment for the last commit: in the SCF module, I passed the mf.mol as mol argument in get_init_guess, since it should perhaps be passed on in this way to be safe. Let me know if there is a reason against passing this on by default.

Wirawan


sunqm:
Hi Wirawan,

Thank you for adding the new module to write the ci vector.  I think
the reading function is not that essential.

In the SCF kernel function,  the defalut environment of
mf.get_init_guess is mf.mol.  Either passing mf.mol or not is OK, and
they're actually the same.  It's not bad you explicitly add it so that
it's more clear which environment is used.

BTW, I noticed you add a new key "mcscf/mo_coeff_new" in
mc1step.CASSCF.save_mo_coeff.  I think it's unnecessary to distinguish
the final solution and the intemediate results.  If there's no
particular consideration, I will change it back to "mcscf/mo_coeff"  in
the next commit.

Best,
Qiming

On Sun, Feb 01, 2015 at 08:55:57PM -0800, Wirawan Purwanto wrote:
> Qiming,
> 
> I created an enhanced checkpointing for MCSCF calculations so we could have a
> copy of the last CI vector at the end of a CASSCF macro iteration. It is not
> enabled by default; to use it, one must pass `dump_chk_ci=True' to the CASSCF/
> CASCI kernel() function or mc1step() method. Right now, only writing is
> supported, I did not implement the reading yet.
> 
> Comment for the last commit: in the SCF module, I passed the mf.mol as mol
> argument in get_init_guess, since it should perhaps be passed on in this way to
> be safe. Let me know if there is a reason against passing this on by default.
> 
> Wirawan
> 
> -------------------------------------------------------------------------------
> 
> 
>   https://github.com/sunqm/pyscf/pull/8
> 
> Commit Summary
> 
>   * Merge branch 'master' into wirawan
>   * Merge branch 'master' into wirawan
>   * Merge remote-tracking branch 'origin/master' into wirawan
>   * Added more comprehensive checkpointing for CASSCF calculations.
>   * Forgot to add pyscf.mcscf.chkfile module!
>   * * Don't forget to specify the 'mol' argument for get_init_guess call.
> 
> File Changes
> 
>   * A mcscf/chkfile.py (41)
>   * M mcscf/mc1step.py (22)
>   * M mcscf/mc1step_uhf.py (22)
>   * M scf/hf.py (2)
> 
> Patch Links:
> 
>   * https://github.com/sunqm/pyscf/pull/8.patch
>   * https://github.com/sunqm/pyscf/pull/8.diff
> 
> —
> Reply to this email directly or view it on GitHub.*
> 

-- 
Dr. Qiming Sun
Frick Chemistry Laboratory
Princeton University
www.sunqm.net

wirawan0:
This is the reason I made the name `mcscf/mo_coeff_new` (please refer to one of the MCSCF's kernel() function): `mo_coeff_new` is the newly optimized MO coefficients *before* the new CI vector was calculated via `casscf.casci` method call. This variable is intended as a transient checkpoint only.  In an unfortunate case that the CI failed, we already get the last MCSCF orbitals stored. The '_new' in the `mo_coeff_new` indicates the newest MO coefficients that have not had the updated CI vector calculated. In any case, when `self.dump_chk` is called (after CI vector calculation), the entire `/mcscf` data group would have been erased and rewritten with only `/mcscf/mo_coeff` in it. So it is somewhat unimportant, but I put it there for safety reason. If you don't think it's necessary, then I'm fine with changing it back to `mcscf/mo_coeff`. Any user trying to take advantage of this checkpointing facility should see where the calculation failed and should take the necessary action (in this case, whether to reuse the saved CI vector or not). I guess in common cases, CI vectors should not change much from one iteration to the other -- am I correct?

Wirawan

sunqm:
I see.  You want to keep a copy of the previous orbital coefficients, in
case the CI solver failed in the current step.  Yes,  if the
mcscf/mo_coeff was overwritten, it could happen that the mo_coeff is not
consistent with the ci vector stored in the chkfile.  However, it should
be very rare unless there is state flipping due to the orbital rotation
and the ground state becomes extremely hard for FCI to converge.  If it
happens, I'd change either CAS space or initial guess because such a
state flipping means the CAS space is not proper and the CASSCF
optimization is not reliable.  Actually, for hard system, I prefer to
store the CAS orbital coefficients for every step like
examples/mcscf/save-load did.  Then I can restore any steps by reading
the corresponding coefficients and feed them to the CASSCF.kernel
function as the initial guess.

Qiming

On Mon, Feb 02, 2015 at 11:52:55AM -0800, Wirawan Purwanto wrote:
> This is the reason I made the name mcscf/mo_coeff_new (please refer to one of
> the MCSCF's kernel() function): mo_coeff_new is the newly optimized MO
> coefficients before the new CI vector was calculated via casscf.casci method
> call. This variable is intended as a transient checkpoint only. In an
> unfortunate case that the CI failed, we already get the last MCSCF orbitals
> stored. The '_new' in the mo_coeff_new indicates the newest MO coefficients
> that have not had the updated CI vector calculated. In any case, when
> self.dump_chk is called (after CI vector calculation), the entire /mcscf data
> group would have been erased and rewritten with only /mcscf/mo_coeff in it. So
> it is somewhat unimportant, but I put it there for safety reason. If you don't
> think it's necessary, then I'm fine with changing it back to mcscf/mo_coeff.
> Any user trying to take advantage of this checkpoi nting facility should see
> where the calculation failed and should take the necessary action (in this
> case, whether to reuse the saved CI vector or not). I guess in common cases, CI
> vectors should not change much from one iteration to the other -- am I correct?
> 
> Wirawan
> 
> —
> Reply to this email directly or view it on GitHub.*
> 

-- 
Dr. Qiming Sun
Frick Chemistry Laboratory
Princeton University
www.sunqm.net

ndattani:
So that future users don't get as confused as I did, much of the code that was merged in Pull Request #8 was modified since then, so `dump_chk_ci` will not be found in the current state of PySCF. Files from this pull request such as `mcscf/chkfile.py` still exist (e.g. see [this](https://github.com/pyscf/pyscf/blob/master/pyscf/mcscf/chkfile.py), which was [renamed](https://github.com/pyscf/pyscf/commits/25b30e616d59f3c807e5a2173f7ec21525240648/mcscf/chkfile.py?browsing_rename_history=true&new_path=pyscf/mcscf/chkfile.py&original_branch=master) in [2017](https://github.com/pyscf/pyscf/commit/25b30e616d59f3c807e5a2173f7ec21525240648#diff-05c02a9e367bdf1cedec3adfc33528f9b1a19fb53b85c8cdf1c50b79480fe313)).

PR #53, opened 2017-01-24
Fixed print statements and import of linalg_helper.py in PBC.

lkwagner:
Feel free to reject this if it doesn't work for your codebase. 

This works correctly under python3 and fixes some imports.


tberkel:
Thanks Lucas.  Just in case you are using it, pyscf/pbc/lib/linalg_helper.py is quickly becoming deprecated in favor of the asymmetric Davidson routine eig() in pyscf/lib/linalg_helper.py.

lkwagner:
That's totally fine. I'm just trying to get PBC calculations to work!

PR #64, opened 2017-03-28
Improved PBC performance

pulkin:
My plan is to reduce the number of calls to `numint.eval_ao` as the performance really suffers from re-calculation of numerical orbitals. I realized that someone left a work-around which caches calls to `numint.eval_ao`, however, I strongly recommend removing this piece of code (`@memory_cache`). There are several reasons for that:

1. First, it does not work with a clean install since `joblib` is not installed by default
2. Second, it actually fails on importing `from pyscf import pbc`. It is weird to me.
3. Third, it wants to cache to `'./tmp/'`. Not user-friendly. Also, magic constants should be avoided in public codes.
4. Finally, it is lazy, i.e., it does not recognize that `kpt` and `kpts` keywords are equivalent by fact. Also, it is not immediately clear whether the cache lookup compares objects (`a is b`), uses `__eq__` (`a == b`) and how it works with mutable objects.

This is the first part of the implementation related to `fft.FFTDF(aoR_loop)` and I will follow-up with the second pull request related to `numint._NumInt(block_loop)`. With this patch, I reduce the calculation time of some Gamma-point model by a factor of two.


pulkin:
Follow-up questions:

1. What does `pyscf.pbc.dft.numint._KNumInt.block_loop` actually do? Is it a parallel or distributed memory thing?
2. Is there any specific reason why this 'macro' is a public function of `_KNumInt`? The only usage of `self` inside this function is `self.eval_ao` which indicates only a loose connection with the class (for example, `self.eval_ao` may be passed instead of `self`).
3. May I safely move this function to another place? Do I need to worry about its compatibility with `pyscf.dft.numint._NumInt.block_loop`? Or maybe `pyscf.dft.numint._NumInt.block_loop` may also be moved outside?

Currently I want to outsource calculations of atomic orbitals to some other place and pass them to `block_loop` as an argument. This will result in a fact that `self` is no longer needed in this function.

sunqm:
Hi Pulkin,

I agree the recalculation of eval_ao causes unnecessary cost and caching AO
value can save a lot of computational time.  joblib is not a good solution and
actually this is in plan to remove it in future.

You may notice in an old commit I removed the cache_ao code because it
potentially causes many problems, including

1. AO derivative level.  Higher order derivatives is required for GGA, meta-GGA,
   TDDFT, gradients, hessian.  Since computation tasks can be changed in
   computation, the cached AO needs to be updated accordingly.
  
2. A late change of .kpts and .gs in class _NumInt and _KNumInt needs to be
   considered in cache_ao.
   
3. _NumInt and _KNumInt also support the use of Becke grids.  A check to the 
   grids scheme is needed.  Moreover, cutoff to grids in Becke grids for lower
   accuracy may be called, so it'd be more safe to check the grids as well.

I actually did checks for these problems in the old cache_ao implementation.  If
you switch to git version 16ada5819f3465b20b85016ecf02420efbed7bd0, you may find
some awkward code checking the environment consistency before calling cache_ao

-        if (self._numint.cell is None or id(cell) != id(self._numint.cell) or
-            self._numint._deriv != 0 or
-            self._numint._kpts.shape != kpts.shape or
-            abs(self._numint._kpts - kpts).sum() > 1e-9 or
-            self._numint._coords.shape[0] != ngrids or
-            (gs is not None and numpy.any(gs != self.gs))):

But the checking code is hard to be complete as other environments, eg
attributes in cell, can also be changed (we never require any object being
immutable in the package although immutable object is assumed in some code).
If cache_ao is a useful feature, I suggest to provide a flag in numint class to
switch on the cache, instead of using cache by default in all situations.
In this treatment, I think we can bypass many checks for cache_ao and leave the
requirements of environment consistency to users.

Last, the master branch is close to a release stage.  Would you please close
this pull request and merge your AO cache code to develop branch, which will be
the 1.4-alpha version in Jun.

Best,

Qiming


On Tue, Mar 28, 2017 at 08:06:10AM -0700, pulkin wrote:
> My plan is to reduce the number of calls to numint.eval_ao as the performance
> really suffers from re-calculation of numerical orbitals. I realized that
> someone left a work-around which caches calls to numint.eval_ao, however, I
> strongly recommend removing this piece of code (@memory_cache). There are
> several reasons for that:
> 
>  1. First, it does not work with a clean install since joblib is not installed
>     by default
>  2. Second, it actually fails on importing from pyscf import pbc. It is weird
>     for me.
>  3. Third, it wants to cache to './tmp/'. Not user-friendly. Also, magic
>     constants should be avoided in public codes.
>  4. Finally, it is lazy, i.e., it does not recognize that kpt and kpts keywords
>     are equivalent by fact. Also, it is not immediately clear whether the cache
>     lookup compares objects (a is b), uses __eq__ (a == b) and how it works
>     with mutable objects.
> 
> This is the first part of the implementation related to fft.FFTDF(aoR_loop) and
> I will follow-up with the second pull request related to numint._NumInt
> (block_loop).
> 
> -------------------------------------------------------------------------------
> 
> 
>   https://github.com/sunqm/pyscf/pull/64
> 
> Commit Summary
> 
>   * Small changes in notation
>   * Added unicode support (important for json-restored objects, etc)
>   * Moved band structure data into a separate file
>   * Added cp2k comparison. The test cases are still under construction
>   * Merge branch 'master' of https://github.com/sunqm/pyscf
>   * Switched to explicit keywords (useful for debugging)
>   * Introduced explicit caching of atomic orbitals. Reduced the number of calls
>     of `numint.eval_ao`
>   * Improved performance of PBC for large unit cells
> 
> File Changes
> 
>   * M gto/mole.py (2)
>   * M pbc/df/fft.py (55)
>   * M pbc/dft/numint.py (4)
>   * M pbc/dft/test/test_ks_noimport.py (257)
>   * A pbc/dft/test/test_ks_noimport_hBN_cp2k.json (1)
> 
> Patch Links:
> 
>   * https://github.com/sunqm/pyscf/pull/64.patch
>   * https://github.com/sunqm/pyscf/pull/64.diff
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or mute the thread.*
> 

-- 
Dr. Qiming Sun
Division of Chemistry and Chemical Engineering
California Institute of Technology
www.sunqm.net

sunqm:
block_loop function is an abstract for generating quantities required by
numerical integration in the evaluation of XC functional, potential,
kernel and density and density derivatives.  Although not implemented,
MPI parallelization should be issued through this abstract function.
With this abstract, functions like eval_vxc or eval_rho will not see the
parallel layer.

Basically, all functions and attributes in pyscf are public objects.
This could save a lot work for developers.  Some objects are prefixed
with _ because I don't want them to be shown in ipython completion list
(or to exclude them from keyword list, see lib.misc.sanity_check function).

I regularly move class methods to module level, eg scf.hf.energy_tot
function, so that many classes can share the same methods without
complicated class inherences.  If that's what you plan to do with
block_loop function, just do it.  Otherwise please leave this design as
it is.  If you just need to pass some AO values, I think an kwarg to
block_loop function should be enough.

Best,

Qiming


On Tue, Mar 28, 2017 at 08:51:01AM -0700, pulkin wrote:
> Follow-up questions:
> 
>  1. What does pyscf.pbc.dft.numint._KNumInt.block_loop actually do? Is it a
>     parallel or distributed memory thing?
>  2. Is there any specific reason why this 'macro' is a public function of
>     _KNumInt? The only usage of self inside this function is self.eval_ao which
>     indicates only a loose connection with the class (for example, self.eval_ao
>     may be passed instead of self).
>  3. May I safely move this function to another place? Do I need to worry about
>     its compatibility with pyscf.dft.numint._NumInt.block_loop? Or maybe
>     pyscf.dft.numint._NumInt.block_loop may also be moved outside?
> 
> Currently I want to outsource calculations of atomic orbitals to some other
> place and pass them to block_loop as an argument. This will result in a fact
> that self is no longer needed in this function.
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or mute the thread.*
> 

-- 
Dr. Qiming Sun
Division of Chemistry and Chemical Engineering
California Institute of Technology
www.sunqm.net

pulkin:
> If cache_ao is a useful feature, I suggest to provide a flag in numint class to
switch on the cache, instead of using cache by default in all situations.

Qiming, I have several arguments to use cache by default:

0. It saves computing time for the cost of memory. In my case it spends 99% of the time in `eval_ao`. If the memory is a concern it will be easy to shadow numpy arrays to calculate something on the fly.
1. Other PBC codes do like that: they pre-compute a set of quantities on the grid before running scf loop.
2. Caching is useful in **most** scenarios, right?
3. To make everything safer against changes we may provide setters and getters (or an equivalent python magic), however, this will not prevent objects and classes from 'hacking' as python is not a safe language. For example, there are fields
```python
        self.stdout = cell.stdout
        self.verbose = cell.verbose
        self.max_memory = cell.max_memory
```
which also become stale when `self.cell` is assigned outside or attributes in `self.cell` are changed. However, nobody cares about that. Maybe lets just add a couple of underscores to these fields?

Thank you for extended responses. I think I will try to define `block_loop` outside the class, outsource integral evaluation, check if everything works fine and then remove `block_loop` class method completely.

PR #66, opened 2017-03-29
Improvements in PBC code

pulkin:
By reducing the number of calls to `eval_ao` I was able to shorten CPU time significantly (factor of ten for the [test model](https://github.com/pulkin/pyscf/blob/master/pbc/dft/test/test_ks_noimport.py)) . There is more work to be done:

- Apply same changes to `KUKS`;
- Make `pyscf.pbc.dft.numint._NumInt` to be a subclass of `pyscf.pbc.dft.numint._KNumInt`
- Run tests
- Remove `@memory_cache` (anyone still needs it?)


pulkin:
Outdated.

PR #70, opened 2017-04-28
KUHF fixes

lkwagner:
Hi folks,

I have been working on KUHF now. I would like you to confirm that I did the right thing on get_occ(). Before I touched it, it did not enforce nup-ndown. Now it enforces that on a per-kpoint basis. This is more or less what we want for QMC but maybe not the correct thing to do in general? 

I also specialized analyze() for KUHF. I removed the computation of the dipole since it doesn't make sense in PBCs, at least the way it's done for molecules. :)


gkc1000:
Although that's an option, I'm not sure it's the right thing to do. Put another way, imagine that we were using the analogous technique for get_occ, but just for the total number of particles. Would you just put the same number of particles per k-point? That would be what I think is called the canonical ensemble approach to twist averaging, but really one wants to just fill the N lowest orbitals across all k points together. 

The way to do it, I think, is to take in the total Nup and total Ndown (over all k points) and then fill up the Nup lowest alpha orbitals and the Ndown lowest beta orbitals, taken over all k points. This would then give the lowest determinant, say, for a set of non-interacting electrons in a magnetic field. The "canonical" approach you implemented could be a different option if it is useful for QMC, or it could just be an addon and swapped in at runtime by replacing the .get_occ() function.

lkwagner:
Yes, you're right. It introduces a finite size error in metals. I'll change the behavior as you suggested. 

Some minor notes/questions:
- I expected that if I set the total spin of the cell, it would be enforced. This is how all other DFT codes that I know of behave. Most of the time you also have the ability to let it vary without setting the spin. Was it intended for the spin to be allowed to vary? Is there a preferred way to implement this? Is it already implemented and I just didn't see it?

- I think the 'canonical' approach can be quite useful in accelerating convergence, especially for systems known to be insulators. So it would be nice as an option, even just as a way to get a good starting guess. I can implement this as an addon. Should it go in kuhf.py since it applies only to kuhf or addons.py?

lkwagner:
OK, would you like to take another look at this? This seems to work on some simple tests. 

I can converge FeO by using the canonical occupation decorator and then restart with open occupation. I had a lot of trouble doing that for the general case. As you mentioned, Garnet, the HOMO is greater than the LUMO in PBE if we force nalpha-nbeta = spin on a per-kpoint basis.

PR #81, opened 2017-06-05
Initial attempt at Travis CI support

mjw99:
This is an initial attempt at building the code with Travis and running the unit tests. Currently some tests are failing and also the process times-out due to a 50 minute time limit on a Travis instance. The latter may be addressed in the future with shorter unit tests, but this currently seems like a good starting point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,39 @@
+# More syntax details at: https://docs.travis-ci.com/user/customizing-the-build
+# Workaround present for https://github.com/sunqm/pyscf/issues/71
+
+sudo: required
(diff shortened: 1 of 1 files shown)


mjw99:
Also, one needs to follow the initial steps here: https://docs.travis-ci.com/user/getting-started/ to enable Travis on the repo.

mjw99:
This (#2) is also quite useful for triggering a CI build without a commit: https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit

PR #89, opened 2017-06-20
argparse fixes for nosetests

mjw99:


sunqm:
Migrate to argparse from optparse.

PR #92, opened 2017-06-21
Correct format in cubefile output.

mjw99:
Please see:

http://www.ks.uiuc.edu/Research/vmd/plugins/molfile/cubeplugin.html
http://gaussian.com/cubegen/
http://paulbourke.net/dataformats/cube/

Also, it might be worth looking at using
https://pypi.python.org/pypi/fortranformat
for any fortran formatting.


sunqm:
master branch is frozen for new features.  Please issue a new PR to dev branch.  Thank you.

mjw99:
OK, thanks. I will do this.

PR #94, opened 2017-06-22
RTTDDFT code

v3op01:
the First batch of RT TDDFT code by John Parkhill
As requested, it only uses package/lib in PySCF.


sunqm:
I didn't go to too much details of the implementation.  Some code style and should be changed before merging:

1. replace hard tab (\t) with soft-tab (8 whitespace)

2. please remove .DS_Store  file in tdscf directory

3. Remove many duplicated functions defined in tdscf/func.py.
    * load_library is defined in pyscf.lib module
    * eval_rhoc looks almost the same to the dft.numint.eval_rho2
    * _dot_ao_dm and Z_dot_ao_dm are defined for complex arrays.  But the relevant functions defined in dft/numint.py have the capability to handle complex arrays. It's not needed to reimplement them.  Also, the code added in lib/dft/nr_numint.c is not needed.

v3op01:
cleaned it up. I don't know if I have to start a new pull request or just need to wait.

gkc1000:
Thank you for committing the new code and functionality!

A few additional comments:

1. in tdscf.py, what does InitFockBuild do? It looks like it is performing an scf cycle, and so could reuse the routines in hf.py

2. What does "By default it does" mean in the docstring for class tdscf?

3. There are some functions that seem broken or empty e.g. tdscf.rhodot, tdfields.Update; do we still need them?

4. In the examples, perhaps the tdhf script function should be called TestTDHF

5. In the long run, it would be good to use pyscf's logging mechanism rather than print statements.

6. There are both tddft and tdscf modules now, should rearrange at some point; perhaps this module should be in future/?

gkc1000:
Thanks for committing the code!

A few additional queries:

1. Does InitFockBuild do an SCF cycle?

2. There are some empty/broken functions, do we still need them? (e.g. tdfields.Update)

3. Some of the attributes in the class tdscf duplicate those found in the mol object or scf object. Also, the comment in the class docstring ends on a cliffhanger "by default it does"

4. In the longer run, it would be nice to switch to pyscf's logging rather than print statements

v3op01:
1. InitFOckBuild does SCF cycle, but I am using it as separate routine to have access to some of the matrices in SCF method needed to run MMUT method. Also, I did not have a chance to look at the newest code. If I can replace the routine, I will.

2. I took out some lines of code before submission. I will add them soon (today). Because some of the function of our original code depended on the C++, I took them out. So, tdfields.Update is needed. 

3. I will look into it. The comments on the class is by my advisor. I do not know what "by default  it does" means. 

4. Yes, I will look into it. I think the logging had to do with the fact that we imported our code directly from QChem, so we will clean things up as we go.

PR #108, opened 2017-07-24
Dev

lkwagner:
Some fixes for python3; I also fixed a problem with the memory allocation in density fitting. Please look over and advise.


sunqm:
What is the use of __getitem__ function in pbc/df/df.py?  It looks the same functionality as the __getslice__ function.

lkwagner:
__getslice__ is python2 only as far as I know.

PR #109, opened 2017-07-24
SHCI_tools did not compile on gcc versions later than 6.

lkwagner:
Needed a cast for div(). Note that it was implicitly using ints. I don't know if that will cause problems for high memory? Actually I don't know what SHCI does!


sanshar:
Hi Lucas,
  This implements some functions to interface our Dice code (that performs
HCI calculations) with pyscf. Thanks for the correction.

Sandeep.

sunqm:
Sandeep, do you have time to modify the code to c99 standard.  I prefer to avoid the C++ compiler in cmake.

PR #115, opened 2017-08-20
Pyscf/Pyberny interface

azag0:
As is, there are two suggestions in `pyscf.geomopt.berny_solver`:

- Function `kernel()` that takes a method and optionally a `mol` and optimizes either the given `mol` or the method's assigned `mol` (both side effects). Returns `pyberny.Molecule`.
- Convertors `as_berny_solver()` and `to_berny_geom()` that adapt Pyscf objects to be passed to `pyberny.optimize()`.

My comments:

- I'd suggest that on the highest API level, Pyberny should be invisible from Pyscf. So the convertors should be applied by Pyscf itself. No Pyscf functions should return Pyberny objects.
- I'd argue against any side effects. Also, in all other contexts in Pyscf, a function called "kernel" has no side effects and returns energy. So I wouldn't reuse that name for a function that has side effects and returns a foreign geometry object.

In this pull request, I suggest to wrap `pyberny.optimize()` with `pyscf.geomopt.optimize()` that does all this conversion internally.

Alternatively, we could wrap `pyberny.Berny()` in a similar way and `pyberny.optimize()` could take an optional `klass=` argument, so wrapping it would then by simply

```python
def optimize(method, mol, **kwargs):
    return pyberny.optimize(method, mol, klass=pyscf.geomopt.Berny, **kwargs)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,8 +59,8 @@ CMakeCache.txt
 CMakeFiles
 Makefile
 cmake_install.cmake
-lib/deps
-lib/config.h
+pyscf/lib/deps
+pyscf/lib/config.h
 
 settings.py
 
--- a/pyscf/geomopt/berny_solver.py
+++ b/pyscf/geomopt/berny_solver.py
@@ -4,7 +4,7 @@ Interface to geometry optimizer pyberny
 '''
 from __future__ import absolute_import
 try:
-    from berny import Berny, geomlib, Logger
+    from berny import Berny, geomlib, Logger, optimize as optimize_berny
 except ImportError:
     raise ImportError('Geometry optimizer pyberny not found.\npyberny library '
                       'can be found on github https://github.com/azag0/pyberny')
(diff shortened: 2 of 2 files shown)


azag0:
When Pyscf creates a method object, it takes a `mol` as an input. Is there a guarantee that the method would be equivalent if it was created with the optimized geometry rather than the initial geometry? If not, this should be probably documented in the section on geometry optimization.

azag0:
Ok, I realized only now that the scanners modify the parent method object, or rather, the mole object assigned to the parent method object. I have to admit that's quite unintuitive to me. It seems to be quite easy to lose track of which mole object is assigned to what method (and so can be modified by the method action). I'd also assume that the scanners returned by `.as_scanner()` won't modify the parent method object when used.

If I were to design the API, I'd either make `optimize()` an attribute of a method, without any return value (and hide the in-place-modfying scanners as private API), because then it's clear that method`.mol` is modified in-place. Or I'd make sure that the scanners don't modify their parent method object.

PR #116, opened 2017-08-24
Fixes; rename to rt (Realtime Tdscf)

v3op01:
Fixed by doing some on TODO list. 
Renamed tdscf to rt (Realtime TDSCF) not to be confused with tddft.


v3op01:
Is the fail due to my code?

sunqm:
No. Some tests in pbc code.

PR #117, opened 2017-08-31
Old state

mbarbry:
I'm not sure how to handle this.
This is a branch of the lastest working version (commit 3b0916c)


mbarbry:
Maybe we should come back in the master branch at this state and try again?

mbarbry:
or restart from scratch with a clean repository from Sun's one and add nao?

sunqm:
We probably need to restart from a clean repo of master branch.  Seems there are more than one module you have added in this PR. We should group the changes based on their features. My plan is to do this in several steps and issue PR ASAP when we get one step or two finished.

First, we need to decide a module name.  "nao" is not a good name that reflects the contents of the code.  Seems the code contains at least an implementation of TDDFT, interface to siesta, some support functions and data.  I recommend to put them in different modules.  We can then choose one to start merge.  Some changes to the existing code need to be revised. More unit tests should be added. We can discuss them later.

mbarbry:
Sorry, This merge request was not intended for this repository but for the one we are actually developing this code. We got problems after some update of the code here but it is now fixed. We will do a proper merge request when the code will be fully working, there is still some functionalities that we want to add. In any case we will take into account your comments.

kovalp:
Dear Quiming,

Let me expand on the pull request from yesterday. The code is realizing an iterative algorithm for linear response TDDFT, while we would like to add more features such as GW/BSE framework.
Currently, there is little overlap of pySCF code with our modules. Indeed, we are specializing the LCAO framework to use the numerical atomic orbitals (NAO), rather than the GTOs which are at heard of pySCF. Initially, our code base was working with a pseudo-potential starting point, in such a way that the active electrons were described with the orbitals of similar extent. Under these conditions, a product basis we proposed (so-called dominant products) is working well delivering an accurate auxiliary basis for the density fitting. When I started to play with pySCF, I was planning to devote this work to the all-electron systems and there would be a greater overlap with the rest of the code. However, the all-electron option (I was planning to follow the FHI-Aims algorithm) requires a firm attention and is not under active development yet. Instead, we are striving to get a performance from the python code similar to the Fortran code and so far concentrating on TDDFT with LDA kernels. In the nearest future, I will concentrate on an iterative algorithm for BSE, which will require GW starting point at some development stage. Furthermore, I hope to get the spin-resolved GW and bulk version of GW. Marc will introduce the necessary additions to compute electron stopping within LDA TDDFT and continue to parallelize the code (with MPI???)? All these developments would be using DFT starting point as provided by SIESTA or GPAW (Troullier-Martins pp starting point, which is weak at GPAW currently) or Fireball or ADF (in principle). I am not sure we will be able soon to use GTO-based LCAOs from pySCF effectively, although I view this as a very desirable feature.

We are looking forward to your opinion. Certainly, it would be nice for us to get our merges to the master version of pySCF. This is to avoid complete changes of the pySCF we are using. Currently, we depend on the 3d integration grids and coupling to the LibXC. The other changes in the pull request are not crucial or even accidental. The deepest change I allowed myself to do in the python source is in the gen_grids.py. I needed to specify the smaller spatial cutoffs for each atom in the grid generation and also I needed a simple radial grid (Gauss-Legendre without any redefinition of integration variable as common for all-electron codes). The other point are unittests. I store them in the pyscf/test and run with 

    `python -m unittest discover test`

I am not sure how you organized the unittests in pySCF.

Best regards,

Peter

sunqm:
Peter, thanks a lot for the explanation.  I drafted a list of the modules to be merged. Please correct me if the listed plans are inaccurate:
* GW/BSE
* Gauss-Legendre grids
* nao module: the basic class to provide the fundamental methods including the ao evaluation, integration for various types of integrals.  (will it be possible to have the nao module provide the same abstract as gto module?  Then the methods on top of nao may be used the same way as other gto-driven methods)
* C/Fortran library to compute nao relevant functions (does it include the support of GPU?)
* numba and GPU driver for nao module.
* nao-driven TDDFT
* SIESTA interface and relevant tools
* GAPW interface and relevant tools

I think MPI is not urgent.  The purpose of pyscf main branch is to provide solid working code rather than the performance-optimal code.  For unittest, I prefer to put the small tests in each module so that we can run required tests when working on a specific module.

Best,
Qiming

kovalp:
Dear Quiming,

I am not sure to what this list should correspond in the programming, but certainly, the GW/BSE framework deserves to stay in a separate module or even a subdirectory, but so far we do not offer any implementation along this line. 

- "Gauss-Legendre grids" are implemented in `numpy.polynomial.legendre import leggauss` and I need only a linear transformation of the integration range from 0 to a maximal value as realized in nao/m_gauleg.py. This can be added to one of the python modules which already exist in the subdirectory dft/. Maybe the module radi.py is right place. Moreover, to use this grid, I need to call this subroutine with correct arguments from the subroutine dft/get_grid.py/gen_atomic_grids. Note also using a keyword argument atom2rcut which is used in this case.

-"nao module:" I think the idea of a nao module as you describe is very logical and we started to realize it in the module m_system_vars.py. The compatibility can be an issue however. Let me spell out an important difference between your approach for bookkeeping (counting) of the orbitals and our approach. Your counting unit is "shell" or radial orbital, while our unit is atom. Moreover, we are realizing a concept of chemical specie in the code (radial orbitals for carbon stored only once for a C60 molecule). This is important for NAO because it is impossible to duplicate the relatively large amount data to describe the numerical radial orbitals for molecules of appreciable size.

- "C/Fortran library..." So far we are using Fortran to realize a low-level computing and this includes evaluation or orbitals, density and some matrix elements. Yes, this is a perfect library. There is no GPU support and I don't see a need for such support.

- "numba and GPU driver for nao module." I am not sure about this. There are only a few python subroutines which were sped up with numba. I am not sure about GPU.

- "nao-driven TDDFT...". Well, the iterative TDDFT we propose can be done also with Gaussian orbitals. The only problem with Gaussians that they are usually very extended comparing to SIESTA's numerical orbitals and there is not much sparsity left in the matrix elements to exploit. Moreover, the algorithm of iterative TDDFT is about using matrix-vector operations in solving Dyson equation rather than determining the whole response functions. Practically, the offered implementation relies on the m_system_vars.py (nao module), but principally not.

- "SIESTA interface and relevant tools"
- "GAPW interface and relevant tools" these are perfect modules, but they would be useful only for NAO... It is generally difficult to import SIESTA's NAOs into GTO.

Practically, how do you prefer to do this merge? I think we can remove git conflicts from 

gto/__init__.py
gw/gw.py
setup.py
tools/__init__.py 

by keeping your original version. In CMakeLists we need to enable Fortran and in gen_grid.py we need some changes.

Best regards,

Peter

kovalp:
>For unittest, I prefer to put the small tests in each module so that we can run required tests when
>working on a specific module.

Do you have a trick to run all unittests in a batch?

kovalp:
In the current state of the branch nao_back, the merge can be done automatically. There are no conflicts.

Peter

sunqm:
For specific module, I run
nosetest -v /path

sunqm:
Dear Peter,

The master branch is in the prerelease stage.  I will not add new feature to the master branch.  So please first issue PR of the code in nao_back branch to pyscf dev branch.

gw/gw.py and tools/__init__.py can be directly merged.  For setup.py, the main differences are the strings of version info and some other parameters.  It can be also safely merged after you updating the relevant code.   For gen_grid, we can first use the changes you made.  I will update this module sooner or later, to apply a more flexible interface for the functions to generate radial grids.

In the C/Fortran code,  I will disable Fortran compiler for default compiling option because Fortran compiler is not by default installed in many systems.  A cmake argument is preferred to enable the compilation of the Fortran code of nao module.  However, this is not urgent, we can make this change after all other merges finished.

tberkel:
The current gw/gw.py is a very suboptimal version I wrote a while ago.  I
have a better version locally, but need to find time to clean it up before
committing it.  I would be happy to help coordinate a more efficient
molecular GW code in pyscf.  There are obvious overlaps with CIS, RPA,
TDHF, and coupled-perturbed HF elsewhere throughout pyscf.

-- 
Timothy C. Berkelbach
Neubauer Family Assistant Professor
Department of Chemistry and James Franck Institute
The University of Chicago

929 E. 57th Street, GCIS E133, Chicago, IL 60637
Cell: 917-446-2687
Office: 773-834-9879
E-mail: berkelbach@uchicago.edu
Web: berkelbachgroup.uchicago.edu

mbarbry:
So we are closing this pull request and open a new one with the right version of the code.

PR #120, opened 2017-09-14
Update to libxc 3 and remove duplicated code

susilehtola:
This PR updates the libxc dependency to version 3.0, as well as removes some of the duplicated code i.e. the IS_LDA, IS_GGA etc arrays that are accessible through accessing libxc's functions.

### DIFF
--- a/pyscf/dft/libxc.py
+++ b/pyscf/dft/libxc.py
@@ -463,9 +463,15 @@ def is_gga(xc_code):
         return all((is_gga(x) for x in xc_code))
 
 def max_deriv_order(xc_code):
-    hyb, fn_facs = parse_xc(xc_code)
-    deriv = [_itrf.LIBXC_max_deriv_order[xc_id[0]] for xc_id in fn_facs]
-    return min(deriv)
+    if isinstance(xc_code, str):
+        if xc_code.isdigit():
+            return  _itrf.LIBXC_max_deriv_order(ctypes.c_int(xc_code))
+        else:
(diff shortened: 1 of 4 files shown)


susilehtola:
There's only one more failing test. It, however, does not make any sense:
```

  File "/home/travis/build/sunqm/pyscf/pyscf/scf/test/test_addons.py", line 84, in test_frac_occ

    self.assertAlmostEqual(mf.scf(), -107.13465364012296, 9)

AssertionError: -104.11587378866246 != -107.13465364012296 within 9 places
```
This is a Hartree-Fock calculation
```
    def test_frac_occ(self):
        mol = gto.Mole()
        mol.verbose = 5
        mol.output = '/dev/null'
        mol.atom = '''                                                                                                                                                                        
            7      0.   0  -0.7                                                                                                                                                               
            7      0.   0   0.7'''
        mol.basis = 'cc-pvdz'
        mol.charge = 2
        mol.build()
        mf = scf.RHF(mol)
        mf = scf.addons.frac_occ(mf)
        self.assertAlmostEqual(mf.scf(), -107.13465364012296, 9)
```
and the test runs perfectly fine on my workstation with the same code tree... Could there be something wrong with Travis itself?

susilehtola:
@sunqm any thoughts why the fractional occupation calculation fails?

sunqm:
I observed the similar problem before.  The tests are all passed on my machine. This frac_occ test occasionally fails on travis-CI. Looks like a problem of symmetry broken in different ways in different runs.

PR #125, opened 2017-09-22
VV10 + range-separated hybrids

nmardirossian:
Enabled range-separated hybrid feature for RKS and UKS
Added VV10 NLC functional for RKS and UKS
Working VV10-containing functionals: wB97M_V, wB97X_V, B97M_V, VV10, LC_VV10
Working RSH functionals: wB97M_V, wB97X_V, CAM_B3LYP, LC_VV10, M11

### DIFF
--- a/pyscf/dft/libxc.py
+++ b/pyscf/dft/libxc.py
@@ -447,6 +447,9 @@ def is_gga(xc_code):
     _itrf.LIBXC_is_gga.restype = ctypes.c_int
     return all((_itrf.LIBXC_is_gga(ctypes.c_int(xid))==1) for xid, fac in fn_facs)
 
+def is_nlc(xc_code):
+    return xc_code[-6:]=='__VV10'
+
 def max_deriv_order(xc_code):
     hyb, fn_facs = parse_xc(xc_code)
     _itrf.LIBXC_max_deriv_order.restype = ctypes.c_int
@@ -481,6 +484,28 @@ def hybrid_coeff(xc_code, spin=0):
             _itrf.LIBXC_hybrid_coeff.restype = ctypes.c_double
             hyb += _itrf.LIBXC_hybrid_coeff(ctypes.c_int(xid))
     return hyb
+
+def nlc_coeff(xc_code):
+    '''Get NLC coefficients
+    '''
+    xid=parse_xc(xc_code)[1][0][0]
+    _itrf.LIBXC_nlc_coeff.argtypes=[ctypes.c_int,ctypes.POINTER(ctypes.c_double)]
+    _itrf.LIBXC_nlc_coeff.restype=None
+    nlc_pars=(ctypes.c_double*2)()
+    _itrf.LIBXC_nlc_coeff(xid,nlc_pars)
+
+    return nlc_pars[0], nlc_pars[1]
+
+def rsh_coeff(xc_code):
+    '''Get RSH coefficients
+    '''
+    xid=parse_xc(xc_code)[1][0][0]
+    _itrf.LIBXC_rsh_coeff.argtypes=[ctypes.c_int,ctypes.POINTER(ctypes.c_double)]
+    _itrf.LIBXC_rsh_coeff.restype=None
+    rsh_pars=(ctypes.c_double*3)()
+    _itrf.LIBXC_rsh_coeff(xid,rsh_pars)
+
+    return rsh_pars[0], rsh_pars[1], rsh_pars[2]
 
 def parse_xc_name(xc_name='LDA,VWN'):
     '''Convert the XC functional name to libxc library internal ID.
--- a/pyscf/dft/rks.py
+++ b/pyscf/dft/rks.py
@@ -12,6 +12,7 @@ import numpy
 from pyscf import lib
 from pyscf.lib import logger
 from pyscf.scf import hf
+from pyscf.scf import jk
 from pyscf.dft import gen_grid
 from pyscf.dft import numint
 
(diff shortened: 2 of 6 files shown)


pyscf/dft/libxc.py
```diff
@@ -447,6 +447,9 @@ def is_gga(xc_code):
     _itrf.LIBXC_is_gga.restype = ctypes.c_int
     return all((_itrf.LIBXC_is_gga(ctypes.c_int(xid))==1) for xid, fac in fn_facs)
 
+def is_nlc(xc_code):
+    return xc_code[-6:]=='__VV10'
```
susilehtola: You should just check here, if the functional has the ```XC_FLAGS_VV10``` flag.

pyscf/lib/dft/libxc_itrf.c
```diff
@@ -383,6 +383,27 @@ double LIBXC_hybrid_coeff(int xc_id)
...
 }
 
+void LIBXC_nlc_coeff(int xc_id, double *nlc_pars) {
+
+        xc_func_type func;
+        if(xc_func_init(&func, xc_id, XC_UNPOLARIZED) != 0){
+                fprintf(stderr, "XC functional %d not found\n", xc_id);
+                exit(1);
+        }
+        XC(nlc_coef)(&func, &nlc_pars[0], &nlc_pars[1]);
```
susilehtola: I think this should also check if the functional actually is NLC.

pyscf/lib/dft/libxc_itrf.c
```diff
@@ -383,6 +383,27 @@ double LIBXC_hybrid_coeff(int xc_id)
...
+}
+
+void LIBXC_rsh_coeff(int xc_id, double *rsh_pars) {
+
+        xc_func_type func;
+        if(xc_func_init(&func, xc_id, XC_UNPOLARIZED) != 0){
+                fprintf(stderr, "XC functional %d not found\n", xc_id);
+                exit(1);
+        }
+        XC(hyb_cam_coef)(&func, &rsh_pars[0], &rsh_pars[1], &rsh_pars[2]);
```
susilehtola: This should check if the functional has XC_FLAGS_HYB_CAM.

pyscf/dft/uks.py
```diff
@@ -34,16 +35,32 @@ def get_veff(ks, mol=None, dm=None, dm_last=0, vhf_last=0, hermi=1):
...
         logger.debug(ks, 'nelec by numeric integration = %s', n)
         t0 = logger.timer(ks, 'vxc', *t0)
 
     hyb = ks._numint.hybrid_coeff(ks.xc, spin=mol.spin)
-    if abs(hyb) < 1e-10:
+
+    #enabling range-separated hybrids
+    omega, alpha, beta = ks._numint.rsh_coeff(ks.xc)
+    if abs(omega) > 1e-10:
+        hyb = alpha + beta
```
susilehtola: What's the point of this? In libxc notation, alpha is full Hartree-Fock exchange, and beta is short-range exchange.

pyscf/dft/rks.py
```diff
@@ -82,14 +100,28 @@ def get_veff(ks, mol=None, dm=None, dm_last=0, vhf_last=0, hermi=1):
             getattr(vhf_last, 'vk', None) is not None):
             ddm = numpy.asarray(dm) - numpy.asarray(dm_last)
             vj, vk = ks.get_jk(mol, ddm, hermi)
+            vk *= hyb
+            if abs(omega) > 1e-10:
+                mol.set_range_coulomb(omega)
+                vklr = jk.get_jk(mol, ddm, 'ijkl,jk->il')
+                vklr *= (alpha - hyb)
+                vk += vklr
+                mol.set_range_coulomb(0)
```
susilehtola: I'm confused. Why do you call get_jk, when the only thing you need is K?

susilehtola: Also, I think it would be faster to evaluate the **short-range** exchange, instead of the **long-range** one, if you're doing any screening.

pyscf/dft/uks.py
nmardirossian: No it is not. CAM-B3LYP, for example, has 19%/65% SR/LR, and returns this:
alpha = 0.65;
beta  =-0.46;
Clearly, SR=alpha+beta, LR=alpha.

pyscf/dft/rks.py
nmardirossian: There is no get_k yet.
PySCF only has erf integrals at the moment.

pyscf/lib/dft/libxc_itrf.c
nmardirossian: It seemed like LibXC was returning zeros if those parameters weren't defined so I thought this was okay. I might change it.

pyscf/lib/dft/libxc_itrf.c
nmardirossian: It seemed like LibXC was returning zeros if those parameters weren't defined so I thought this was okay. I might change it.

pyscf/dft/uks.py
susilehtola: Well, you're not wrong. As I said, alpha is full Hartree-Fock exchange, and beta is short-range only. Thus, in the long range you get just alpha. And in the short range, you get alpha+beta. I'm just saying that ```hyb``` is misleading, since that's the (non-range-separated) hybrid coefficient which is just alpha.

pyscf/lib/dft/libxc_itrf.c
susilehtola: Yes, they should be initialized to zero. But the thing is that there's some functionals that use a different range separation, such as CAMY_BLYP and CAMY_B3LYP that use Yukawa screening.

pyscf/dft/rks.py
sunqm: get_k function does not call the range separated coulomb operator when integrals are cached in memory.

susilehtola: I'm not sure I understand what you mean: would calling  ```get_k()``` give you the wrong answer then?

Does ```mol``` store the integrals? Shouldn't you have different objects to store the integrals for different values of omega?

sunqm: integrals are cached in mf object if they can be fit in memory.  I didn't consider the possibility of multiple types of coulomb integrals when I wrote the scf class.  mf.get_k function will use the cached integrals to construct K matrix no matter which omega value is assigned.  jk.get_jk function computes the JK matrix directly.

PR #130, opened 2017-10-04
NAO is able to utilize pySCF's RHF

kovalp:
Dear Quiming,

recently I achieved to utilize the pySCF's RHF class. In order to this, I need to touch the scf/hf.py. I see there are other changes to the same file and would like to enter the features crucial for the NAO branch. As you could see, I realized the necessary matrix elements (laplacian, hartree, exchange and nuclear-electron interaction) as methods of my Mole analogon system_vars_c. I would like that these methods are called instead of those implemented in the scf/hf.py if a corresponding attribute exists in the instance of Mole class or its substitution.

I think also that it makes sense to define some functions from hf.py as methods of Mole class. Namely, the initial guess for the density matrix is an operation which needs only the data from Mole and therefore it can painlessly become a method. Moreover, the calculations of J and K matrix elements only need density matrix additionally to the data in the Mole object and therefore may be good candidates to become methods. The same applies to the nuclear-electron interaction (maybe up to core polarization data?) and kinetic energy. These do not need even anything additional.

Among the other changes, we improved the low-level library, @mbarbry also did some additions to the iterative TDDFT algorithm to use the wrappers from a new version of scipy.

We are also using the cmake.arch.inc mechanism. I updated the installation instructions and deleted CMakeLists.txt from the examples directory. The feature works as a clock! I am very satisfied now. For better handling, I would suggest to rename the directory with cmake.arch.inc examples to cmake_arch_inc_examples.

Best regards,

Peter


gkc1000:
Hi, PySCF is intended to be for more general calculations than mean-field
or mean-field + correlation methods. For example, an initial one-particle
DM has no meaning in a DMRG calculation ... so this is why the initial
guesses for the calculation methods are inside the calculation objects
themselves.

Garnet

gkc1000:
Well, how exactly do you need to generalize scf?

One possibility is to examine the PBC RHF objects. These are derived from
the molecular RHF and SCF classes, but reimplement some of the methods.

We are thinking of implementing PW SCF within the same SCF framework, it
may not be general enough so some things may also need to be tweaked there.

kovalp:
I need to override the methods to compute matrix elements basically.
Probably to squeeze PW into the same RHF one needs deeper changes...

Peter

gkc1000:
Overriding the methods to compute matrix elements (e.g. get_t, get_j etc.)
is what was done to implement PBC RHF ... so you could try the same model
for now, unless Qiming has other suggestions.

kovalp:
Thank you for pointing this out! I will have a look.

kovalp:
Well, I had a look. I am not convinced this would be a right thing to do in
my case. In pbc.hf.py you have a trouble with complex quantities and
therefore you just wrote another version of scf/hf.py

I believe in my case it is not justified to reimplement basically the same
functionality, with the same type of quantities.

At the same time, I do not like too much my current spaghetti-solution.

Best regards,

Peter

kovalp:
Well, maybe if I inherit and move the definitions of matrix elements into
my nao/hf.py then the reimplementations are minimal. I see the point!

I will try that.

Peter

sunqm:
It's too hacky to patch the GTO scf/hf.py with the mol.get_jk.  It will be hard for others to understand why there is such an exception. I think an scf/hf.py in nao module would be better.

gkc1000:
Peter: if you look at the pbc/hf.py, there is basically no duplication. The
differences are not due to complex numbers, they are because the integrals
with PBC are different since one needs to have lattice sums. So the
molecular get_t, get_jk, get_nuc, get_ovlp must be replaced. But once these
are defined, then all the other things like the scf kernel are either not
implemented in the pbc/hf.py, because they just use the original scf/hf.py
implementation, or they are simple 3 line wrappers to take into account
that the Mole object is replaced by a Cell.

If one wants to subclass the integral generating objects themselves, that
is also possible to consider. Currently, that is already done for the JK
part, because there are several DF (density fitting) objects that generate
the JK matrix, e.g. DF, mixed DF, pbc.DF, etc. that underneath work with
different fitting bases. One can imagine solving the Poisson equation in
NAO and that being its own kind of DF. The part which is not subclassable
is the one-electron integrals, which currently call either mol.intor_XXX or
pbc.intor_XXX. So, if one really wanted NAO's for both molecular and PBC
settings, that reuse the lattice sums in pyscf for PBC, then one might have
an "intor" object which is subclassable.

But for the time being, I think the simplest thing for you is to implement
nao/hf.py.

Garnet

kovalp:
I see. I agree.

kovalp:
Ok, I did this exercise. Now I touched scf/hf.py only to allow transmitting
a keyword argument.

Peter

sunqm:
I didn't see the pyscf.mpi module would be used anywhere except the nao module.  I don't have a good idea how to organize the mpi component in pyscf for now.  So I'll move this mpi code to pyscf/nao directory.

kovalp:
Muy bien.

PR #157, opened 2018-02-08
Fixing 2D FFTDF

jdmcclain47:
I realized there were a couple of things that needed to be fixed in the 2D FFT concerning proper behavior for ewald exchange treatment as well as ERI creation.  A full list of changes is given below:

- Fixed a number of warning/error messages concerning the use of the low_dim_ft_type with other density fitting techniques such as AFTDF.

- Fixed the low-dimensional FFT treatment within the fft_ao2mo module.

- When given eri's explicitly the HF energy would be different than if they weren't given explicitly.  This would only happen if the mesh grid weren't large enough, but having similar behavior for both cases is nice.

I've added a few tests as well and updated some other tests based on this change in how the ewald exchange divergence is treated.

### DIFF
--- a/examples/pbc/32-graphene.py
+++ b/examples/pbc/32-graphene.py
@@ -25,8 +25,6 @@ from pyscf.pbc import tools
 nk = 1
 kpts = [nk,nk,1]
 Lz = 25 # Smallest Lz value for ~1e-6 convergence in absolute energy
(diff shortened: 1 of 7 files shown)


jdmcclain47:
Note: that while the 2D FFT case seems to work now for CCSD in systems like graphene, there is still weird behavior in systems like MoS2.

PR #164, opened 2018-03-07
Fixed eval_gto for low-dimensional systems

jdmcclain47:
For systems with reduced periodicity, there were problems when the summation of lattice vectors in the non-periodic direction was not included.  These effects were most noticeable when the atoms were situated near the cell boundary and only affects the FFTDF methods (since the AFTDF does not seem to use this function).

This fixes one of the tests in pbc/scf/test_hf.py


sunqm:
In the low-dimension system, this function is also used in DFT numerical integration code. The repeated images should not be placed in the vacuum when computing DFT density in real space.  There are two kinds of workaround I can imagine:

1. pbc_eval_gto function takes an extra flag to force the lattice summation in vacuum.

2. Force cell.dimension=3 when low_dim_ft_type is set

PR #165, opened 2018-03-12
Dev

januseriksen:
Not sure that the mo_energy = None hack in mc1step.py is the best (far from it, really), but I'll let you decide this.


sunqm:
I think set mo_energy=None is reasonable.  Another possible choice is to set it to HF orbital energy.  But I guess this may be slightly misleading since the casscf code supports any initial orbitals and then "orbital energy" is not the HF orbital energy. Although setting it to None may lead to some sort of TypeError if people use it later, I guess it is better than a wrong value.

PR #168, opened 2018-03-16
Ccsd4metals

jdmcclain47:
This is a pull request to add functionality to the k-point CCSD module so that it can handle a non-equal number of occupied orbitals per k-point.  There are a couple of ways to do this, but I settled on just padding MO coefficients with zeros so that each k-point effectively has the same number of occupied orbitals and MO's at each k-point.

The main files affected are in the pyscf/pbc/cc folder.  However, I did add some minor "functionality" to the base pyscf/cc branch so that it can do a damping procedure for the iterative CC calculations.  This seems to be the surest way to get many k-point CCSD calculations converged; I'm working on some new ways to ensure convergence.

I also attempted to clear up a slightly unrelated issue, by throwing an appropriate exception, that occurred when trying to convert a KGHF object into a KRHF object.


jdmcclain47:
I realize this is breaking some of the previous frozen tests... I'll fix that.

PR #176, opened 2018-04-10
Update dmrgci.py

etaffet:
Added functions get_es and get_excited to class DMRGCI for the comprehensive evaluation of excited-state multireference wavefunctions.

### DIFF
--- a/pyscf/dmrgscf/dmrgci.py
+++ b/pyscf/dmrgscf/dmrgci.py
@@ -171,6 +171,98 @@ class DMRGCI(lib.StreamObject):
 
         self._keys = set(self.__dict__.keys())
 
+        
+    def get_es(self, mol, mc, mo_dmrgscf, mocas, aolst):
+        '''
+For a multireference calculation, return the static and dynamic
+correlation for each state.  For non-DMRG fcisolvers, also return
+the local spin and wavefunction coefficients for each state.
+
+Input:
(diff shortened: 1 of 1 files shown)


sunqm:
@etaffet  I think they are very good examples of how to run NEVPT calculation on top of CASSCF/DMRG-CASSCF calculation and how to analyze the relevant results, like spin, local-spin, transition dipole and so forth. Dmrg interface does not look like the right place to put the two features. I'd recommend creating a new example with some background of the calculation for the two features.

pyscf/dmrgscf/dmrgci.py
```diff
@@ -171,6 +171,98 @@ def __init__(self, mol, maxM=None, tol=None, num_thrds=1, memory=None):
...
+self:  mc.fcisolver
+mol:  the molecule inputted
+mc:  mcscf.CASCI
+e_gs:  computed ground-state energy
+e_s:  computed total energies of all relevant states
+ci:  computed ground-state wavefunction
+spin_gs:  number of unpaired ground-state electrons.
+        '''
+        orbcas=mc.mo_coeff[:,mc.ncore:mc.ncore+mc.ncas]
+        mol.set_common_orig_((0,0,0))
```
sunqm: The transition dipole has the dependence on the gauge origin. It seems better to set the origin to the nuclear charge center.

PR #179, opened 2018-04-12
'fake' argument to SCF + incore option for DIIS + new CC test 15-incore_mo_ccsd_t.py

januseriksen:
Also updated CC tests

41-ccsd_with_given_hamiltonian_noscf.py
10-ao_direct_ccsd.py

to reflect recent changes.

### DIFF
--- a/examples/cc/10-ao_direct_ccsd.py
+++ b/examples/cc/10-ao_direct_ccsd.py
@@ -21,6 +21,7 @@ mf = scf.RHF(mol).run()
 #
 mycc = cc.CCSD(mf)
 mycc.direct = True
+mycc.frozen = 1 # frozen core
 mycc.kernel()
-print('CCSD correlation energy', mycc.e_corr)
+print('CCSD correlation energy =', mycc.e_corr)
 
--- a/examples/cc/15-incore_mo_ccsd_t.py
+++ b/examples/cc/15-incore_mo_ccsd_t.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+#
+# Author: Qiming Sun <osirpt.sun@gmail.com>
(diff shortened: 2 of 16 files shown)


sunqm:
Do you need to remove IO completely?  The overhead of writing chkfile is not high.  The temporary file is only created and it is deleted automatically after calculation finished.

Attribute fake is not necessary in my opinion. I can slightly modify the scf code to mute chkfile when mf.chkfile is set to None or False.

januseriksen:
The thing is that my code may be parallelized over the individual cores of a single or multiple nodes, and many, many CC calculations (e.g., of the type in 15-incore_mo_ccsd_t.py) are then run simultaneously on every core of a node. Hence, I/O - or even file handling in generally - is to be avoided, in particular when it's redundant, as is the case when the mean-field object isn't really used at all, but only serves as input for a subsequent CC calculation. 

The present proposal is just that - a proposal - and I admit that it's perhaps not too optimal. Do you have specific suggestions, which I can implement, or would you rather make changes to my code yourself?

sunqm:
I see. I have some idea but not a clear solution right now.  Roughly, it is to initialize the file handler of chkfile lazily when needed. The extra argument fake of scf object does not look like a good solution for me in this scenario.  This change also affects post-HF method implementation. The post-HFmethods will need to also check this parameter when accessing mean-field chkfile.

Do you mind to revert your PR a little bit back, to the point where attribute "incore" of DIIS object was implemented while "fake" was not in scf object?  I think we can create an issue about the I/O and file handler of chkfile, and include other developers in the discussion.

januseriksen:
OK, I think I got it all reverted while keeping everything else intact.

@sunqm - do you mind opening an issue? I think you know what I want, and then you may init the discussion with your idea on how to address things...?

PR #187, opened 2018-05-15
Mpiccsd fix

jdmcclain47:
The mpiccsd still needs a fair bit of work in terms of clean-up, but I made changes so that it works properly with the current pyscf.  I'll be focusing on cleaning it up a bit and we can decide what to do about the various tests and whether it should be located in maybe a different branch.

Overall the changes were predominantly contained inside the mpicc directory, however I did make one change to the kernel in the cc/kccsd_rhf.py so that the mpicc/kccsd_rhf.py could reuse the same kernel.  In it, I added a method for calling a local function to obtain the norm of the T amplitudes since for some large t2 arrays in the mpiccsd code this can cause problems.


jdmcclain47:
Forgot to mention that it also allows for freezing of orbitals now across k-points.

PR #197, opened 2018-06-18
Arbitrary-order coupled-cluster implementation

pulkin:
Only general (unrestricted) theory up to CCSDT so far. EOM-CC is not completely valid (likely targets wrong states).


pulkin:
Try this one. Code generating files are not yet ready to be public.

PR #221, opened 2018-08-21
kconserv

pulkin:
Re-implemented kconserv to support arbitrary dimensions.


sunqm:
I'm not quite sure what improvement this PR brings. @pulkin, could you describe a little bit more about the purpose of this PR? Why we need a generalized version of kconserv function? Does the new function give better performance? Thanks.

pulkin:
I had to re-implement this function for my code. You may need it when using different index notation for ERI/T-amplitudes. It will also be necessary for CC-triples. I did not test it performance-wise.

PR #222, opened 2018-08-30
updated pyscf/dft/libxc.py for libxc-4.2.3

zwang123:
Added new metaGGA functional REVSCAN as well as VV10/rVV10 corrections. Other functionals in the latest libxc library could also be included later.


sunqm:
The default version of libxc library has been upgraded to 4.2.3 on the dev branch.

M05 functional was affected in this updates. M05-X and M05-C functionals will not be supported in pyscf-1.6.

PR #234, opened 2018-10-01
KRCCSD EOM code reshuffle

pulkin:
Reorganized the code, fixed a couple of bugs, added some tests.

* EOM IP and EA routines are now in separate files;
* the vector representation of amplitudes no longer includes padding entries. As such, EOM vector sets belonging to different k-points are now lists of arrays of different shapes rather than plain arrays;
* included initial guess tests and fixed bugs preventing use of initial guesses;
* all EOM functions now have shorter names. The kernel EOM functions are now called `kernel`. Both `cc.ipccsd` and `cc.eaccsd` can still be called but all other service EOM functions no longer belong to the `RCCSD` class.


pulkin:
Be specific, please. If it is only about naming conventions that can be easily fixed. If new issues are introduced with this PR please point me to them.

tberkel:
This really isn't targeted at your PR specifically, but rather to discuss how the CC PBC code should be organized; although yours does change a lot of the naming conventions and organization, in addition to implementing new features, test cases, and bugfixes.  All I am encouraging is that the PBC code should look like the molecular code, unless there are good reasons otherwise.

But because you asked me to be specific: pyscf/cc has files such as ccsd.py, rccsd.py, and eom_rccsd.py (not separated into separate IP and EA for example).  The latter implements an EOMIP and EOMEA class, which inherit from an EOM base class.  The classes are instantiated in the CCSD class in ccsd.py, and exposed to the user via ipccsd() and eaccsd() methods.

So, I'll propose (and very welcome to input from all developers), that we restructure the CC PBC code into files such as kccsd.py, krccsd.py (not our relic kccsd_rhf.py), and eom_krccsd.py.  These would have classes like KCCSD and KRCCSD along with EOM, EOMIP, EOMEA, etc.  The dev branch is a natural place for this restructuring.

As a more minor example specific to your PR: the renaming of the amplitudes_to_vector() functions to a2v() seems (to me at least) excessively opaque.  It makes it harder to compare with the molecular code, which uses the amplitudes_to_vector() naming, and the meaning of a2v() could only be guessed by someone familiar with the older amplitudes_to_vector() name.

pulkin:
If you think I can further improve this PR you are welcome with your comments. I kindly ask again, to be specific about any issues **with this PR** (not issues in molecular, pbc modules or any other part of pyscf). I will gladly help fixing those but lets not discuss it here.

sunqm:
I agree it would be better to have consistent names and methods in molecular CC module and PBC CC module. Some code layout and names in the molecular module may be bad. I'd like to hear more input from different developers as well.

Although not closely related to this PR, another thing that is worth to discuss for the CC modules is the convention of ERIs. I chose the chemist's convention without anti-symmetrization at the beginning because it requires relatively less IO. But it turns out that in many literatures the integrals and intermediates are actually antisymmetrized tensors in the physicist's convention. It seems causing some confusion. Whether should we use the same convention everywhere? Or keep it as it is and check the conventions in the code? We should discuss and make updates in the next major release (v2.0)

jdmcclain47:
Not to derail the PR, but is there a place to discuss these issues?  Maybe could happen through a github organization/team discussion or just general issues.

(As a side note I personally think the AS physicists are fine but maybe just need some further checks when used in other parts of the code.  They're not meant to be optimal at this point and mainly for checking, the latter of which is easy to do when the literature has everything in AS physicists notation.  Also, the eom codes *are* getting fairly long, we could discuss maybe breaking these up in some manner.)

jdmcclain47:
The question for whether there was some “future directions” section was referring to Qiming’s request for other dev input.

Id agree with @tberkel ‘s assessment that I think the bugfixes and tests should be included but that breaking up the files should maybe be done at a later date as this moves further from the current goal of making the periodic and molecular codes more consistent.

pulkin:
I feel like "moves it further away" is going to haunt this PR. It looks like everyone agrees here that EOM and ground state codes have to be separated but somehow ignores the fact that eom_rccsd.py is a huge chunk of code with 1800 lines (compared to 450 lines in rccsd.py) and multiple similar functions and classes with similar names which do quite different things. I am totally positive about keeping API consistent and this PR exposed several functions where one would probably expect them to see. I am not positive about enforcing bad decisions from the past to keep the file structure. I had to fix some bugs and I had to be sure that I do not accidentally use `vector_to_amplitudes_ip` in an EA routine. This is why the code was reorganized and these functions were put in different places. I am closing this PR because I do not see a way to fix it.

tberkel:
Of course the other contributions in this PR should be merged if possible,
such as the handling of padding orbitals at different k-points.

We should simultaneously continue discussion via an “issue” on github or
over email of interested developers. I think the most important part is
consistency, and we could discuss options of organization and API for the
EOM code, in both the molecular and PBC code simultaneously.

sunqm:
@pulkin Would you mind to first merge your updates of
kccsd.py
kccsd_rhf.py
kmp2.py
test_kpoint_rhf.py
kpts_helper.py
test_kpts.py 
if it's not too complicated? Thanks.

pulkin:
Merge where?

sunqm:
dev branch

pulkin:
I do not quite understand you but you should look into https://github.com/pulkin/pyscf/tree/cc-fix for these changes. The branch also contains other fixes I need (such as reverting your unfinished work on gdf/2d-kccsd). You a free to pull it (its up-to-date for me).

PR #243, opened 2018-10-31
Additional guess for initial UHF density matrix

jdwhitfield:
Symmetry breaking of the UHF and RHF solutions does not occur when using
any of the broken symmetry initial guess schemes in PySCF.  This is because
current methods break symmetry without breaking the spatial density. The
function `init_guess_mixed` mixes the HOMO and LUMO of the guess by core
 to break spatial symmetry of alpha and beta HOMOs following Szabo
chapter 3.  

This has been tested for the stretched hydrogen symmetry breaking problem 
but not beyond this two-electron case.

### DIFF
--- a/examples/scf/56-h2_symm_breaking.py
+++ b/examples/scf/56-h2_symm_breaking.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python
+# Author: James D Whitfield <jdwhitfield@gmail.com>
+'''
+Scan H2 molecule dissociation curve comparing UHF and RHF solutions
+
+The initial guess available for 
+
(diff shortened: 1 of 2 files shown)


sunqm:
This is a useful feature. Mixing the HOMO, LUMO, and other valence orbitals regardless of the spatial symmetry can be used in many circumstances, not just the initial guess from Hcore. The implementation is not easy to be extended to the general case.  I recommend to put it in the "example" folder, as a demonstration how to construct SCF initial guess.

jdwhitfield:
Sure I can put the full stretched H2 UHF/RHF symmetry breaking in the "example" folder

PR #257, opened 2018-12-10
Add fix spin for ucasci.

hshi:
A function fix_spin_ is added in ucasci so that we can use fix spin for UCAS.

### DIFF
--- a/pyscf/mcscf/ucasci.py
+++ b/pyscf/mcscf/ucasci.py
@@ -404,6 +404,21 @@ class UCASCI(casci.CASCI):
         dm1a,dm1b = self.make_rdm1s(mo_coeff, ci, ncas, nelecas, ncore)
         return dm1a+dm1b
 
+    def fix_spin_(self, shift=.2, ss=None):
+        r'''Use level shift to control FCI solver spin.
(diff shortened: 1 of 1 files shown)


pyscf/mcscf/ucasci.py
```diff
@@ -404,6 +404,21 @@ def make_rdm1(self, mo_coeff=None, ci=None, ncas=None, nelecas=None,
...
+
+            (H + shift*S^2) |\Psi\rangle = E |\Psi\rangle
+
+        Kwargs:
+            shift : float
+                Level shift for states which have different spin
+            ss : number
+                S^2 expection value == s*(s+1)
+        '''
+        fci.addons.fix_spin_(self.fcisolver, shift, ss)
```
sunqm: This function works for restricted orbitals only. The integrals of operators S_x, S_y have to be explicitly considered for unrestricted orbitals. They are not simply zero or one as the spatial degenerated orbitals. If alpha and beta orbitals are very different, calling the restricted version would cause large error.

sunqm:
This PR does not solve the spin problem in UCASCI

PR #270, opened 2019-01-22
Bugfix: carbon-atom ANO basis contraction coefficients

MatthewRHermes:
In the version of the ANO-RCC basis distributed with PySCF, the carbon atom contraction coefficients are the same as the nitrogen atom contraction coefficients.  In fact, they should be different; see bse.pnl.gov or the basis set libraries of other programs such as OpenMolcas (https://gitlab.com/Molcas/OpenMolcas/blob/master/basis_library/ANO-RCC).


sunqm:
@MatthewRHermes  Thanks for finding this issue.
 
Adding relevant info. The basis set data was downloaded from EMSL later 2014. The basis of carbon was updated 2015 in Molcas-2015 release due to the error reported in this page: http://www.molcas.org/ANO/indexC.html

PR #275, opened 2019-01-30
Add easy way to specify basis contraction scheme  in ano@6s5p3d2f form

bogdanoff:
It is the only practical way to utilize ANO-RCC basis sets. In future one can introduce aliases like `O.ANO-RCC-VDZ` for `O.ANO-RCC...3s2p` which are also there in Molcas.
I choose `@` symbol arbitrary as it is not very often used in filenames.

### DIFF
--- a/pyscf/gto/basis/__init__.py
+++ b/pyscf/gto/basis/__init__.py
@@ -326,6 +326,32 @@ def parse_ecp(string, symb=None):
     return parse_nwchem.parse_ecp(string, symb)
 parse_ecp.__doc__ = parse_nwchem.parse_ecp.__doc__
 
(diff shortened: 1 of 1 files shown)


sunqm:
Does the truncation notation work for basis other than ANO?

bogdanoff:
You are right, I assumed general contraction. 
Now I tried simply to set `optimize=True ` when ask for truncation to do "segment contraction -> general contraction".
But it looks like `parse_nwchem.optimize_contraction` does something else, as there are still two s blocks for H.cc-pvdz.

bogdanoff:
Truncation should work now for segmented basis sets too.

Issue #294, opened 2019-03-13
[Bug] IAO does not span the occupied space


zhcui:
Hi all,

The current IAO code has a bug, i.e. the trace of density matrix in IAO basis is not equal to the number of electrons in the system.  The IAO, from design, should span the occupied space. You can check the trace of `dm` in `02-pop_with_iao.py` example, which has error compared with nelectron.

The bug comes from a missing orthogonalization for the intermediate `ctild` (see equation (1) and Appendix C in IAO paper for details).  We only need to add one line to fix it, i.e.
`ctild = orth.vec_lowdin(ctild, s1)`

Best,
Zhihao

gkc1000:
Please submit a pull request.

zhcui:
Hi Garnet,

I have created a pull request, which fixes this issue and allows k point sampling for IAO.

Best,
Zhihao

sunqm:
Fixed in PR #295

PR #295, opened 2019-03-13
fix bug in IAO, add possible k point sampling for IAO

zhcui:
See also the issue #294

### DIFF
--- a/examples/local_orb/02-pop_with_iao.py
+++ b/examples/local_orb/02-pop_with_iao.py
@@ -3,7 +3,6 @@
 '''
 Mulliken population analysis with IAO orbitals
 '''
-
 import numpy
 from pyscf import gto, scf, lo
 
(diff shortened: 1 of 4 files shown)


gkc1000:
Before I can accept this pull request, please update the failing tests (see the travis-CI log).

zhcui:
There are strange randomness or numerical instability in the `test_localizer.py` , which makes the test sometimes fail. However, I do not change anything related to this test file.  Also I do not see a random part in the script.

zhcui:
Hi Qiming,

@sunqm 

Could you take a look at `test_localizer.py`, which seems not always passing even without the pull request I made?

Best,
Zhihao

sunqm:
The failure of test_localizer is caused by the Boys localization. It does randomly converge to a local solution. I should update it at some point. Your code should be fine

zhcui:
Hi Qiming,

Thanks for your reply. 

Just for curiosity: Is there any randomness in Boys and PM if the default guess is 'atomic'? My feeling is that the solution should be deterministic. 

Best,
Zhihao

gkc1000:
Typically things are not completely deterministic if there are multiple
threads due to different roundoff error when different orders of threads
complete.

zhcui:
Thanks, Garnet.

BTW: the pull request is now merged only to the master branch, and should it be merged to dev as well?

Best,
Zhihao

sunqm:
Zhihao, I'll update the dev branch in my next PR to dev

Qiming

On Fri, Mar 15, 2019 at 01:43:55AM -0700, Zhihao Cui wrote:
> Thanks, Garnet.
> 
> BTW: the pull request is now merged only to the master branch, and should it be
> merged to dev as well?
> 
> Best,
> Zhihao
> 
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly, view it on GitHub, or mute the thread.*
> 

-- 
Dr. Qiming Sun
Principal Researcher, Tencent Quantum Laboratory

Issue #329, opened 2019-05-10
PySCF fails to compile with GCC 9


susilehtola:
PySCF fails to compile with GCC 9 due to broken OpenMP code. For example,
```
/home/work/pyscf/pyscf/lib/np_helper/npdot.c:74:34: error: ‘k’ not specified in enclosing ‘parallel’
   74 |                 int nblk = MAX((k+nthread-1) / nthread, 1);
      |                                 ~^~~~~~~~
/home/work/pyscf/pyscf/lib/np_helper/npdot.c:27:27: note: in definition of macro ‘MAX’
   27 | #define MAX(X,Y)        ((X) > (Y) ? (X) : (Y))
      |                           ^
/home/work/pyscf/pyscf/lib/np_helper/npdot.c:70:9: error: enclosing ‘parallel’
   70 | #pragma omp parallel default(none) shared(a, b, c) \
      |         ^~~
/home/work/pyscf/pyscf/lib/np_helper/npdot.c:76:59: error: ‘m’ not specified in enclosing ‘parallel’
   76 |                 double *cpriv = malloc(sizeof(double) * (m*n+2));
      |                                                          ~^~
/home/work/pyscf/pyscf/lib/np_helper/npdot.c:70:9: error: enclosing ‘parallel’
   70 | #pragma omp parallel default(none) shared(a, b, c) \
      |         ^~~
/home/work/pyscf/pyscf/lib/np_helper/npdot.c:76:59: error: ‘n’ not specified in enclosing ‘parallel’
   76 |                 double *cpriv = malloc(sizeof(double) * (m*n+2));
      |                                                          ~^~
/home/work/pyscf/pyscf/lib/np_helper/npdot.c:70:9: error: enclosing ‘parallel’
   70 | #pragma omp parallel default(none) shared(a, b, c) \
      |         ^~~
```

sunqm:
This error is caused by the changes in OpenMP 4.5 standard.  In the old OpenMP standard (up to v4.0), session 2.14.1.2 states "Variables with const-qualified type having no mutable member are shared.". GCC 9 uses OpenMP 4.5 standard by default and the new standard removes this rule.

Although this error is easy to fix, I'd recommend NOT to use GCC 9 in your production code, for pyscf-1.6, pyscf-1.7. We didn't have enough tests for GCC 9. The changes of OpenMP standard is an example of the problem using GCC 9 series.

sunqm:
Problem is solved by PR #330

PR #330, opened 2019-05-10
Fix compilation with GCC 9.

susilehtola:
This addresses issue #329 by removing the unnecessary OpenMP pragmas.

Instead of manually defining the roles of all the parameters, it suffices to specify the private ones, since by default everything is shared.

The code should work exactly as before, since I've left in all the `private` statements.

### DIFF
--- a/pyscf/lib/cc/ccsd_grad.c
+++ b/pyscf/lib/cc/ccsd_grad.c
@@ -81,9 +81,7 @@ void CCvhfs2kl(double *eri, double *dm, double *vj, double *vk, int ni, int nj)
         memset(vj, 0, sizeof(double)*ni*nj);
         memset(vk, 0, sizeof(double)*ni*nj);
 
-#pragma omp parallel default(none) \
-        shared(eri, dm, vj, vk, ni, nj) \
-        private(ij, i, j, off)
+#pragma omp parallel private(ij, i, j, off)
         {
                 double *vj_priv = malloc(sizeof(double)*ni*nj);
                 double *vk_priv = malloc(sizeof(double)*ni*nj);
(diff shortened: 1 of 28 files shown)


sunqm:
This PR removes "default(none)" from the statement "#pragma omp parallel". There a few questions to sort out before merging.

1. Does OpenMP standard specify the default data sharing rules if "default(xxxx)" not specified? AFAIK, different variables, e.g. the one on stack and the one on heap, may apply different default rules.

2. Old GCC compilers probably use old OpenMP standards (2.5 or 3?). We at least need to support GCC-4.4 which is the default compiler in redhat 6. Are OpenMP standards the same for "#pragma omp parallel", in all GCC series, from GCC-4.4 to GCC9?

sunqm:
@matt-chan you may know more or less on this.

susilehtola:
1) @sunqm if the default is not specified, all variables except the loop variable are shared. Variables that are declared inside the parallel block are private to each thread. Same goes for variables that have explicitly been declared private.

2) Yes, the data sharing stuff has been the same ever since the first releases of OpenMP.

sunqm:
@susilehtola  Thanks for the clarification. OpenMP had an implicit statement about the default treatment if nothing specified in session "Data-sharing Attribute Rules". The standard says "Objects with dynamic storage duration are shared." It's likely identical to the clause default(shared). Changes in this PR do not have bugs. But I think it's better to use the explicit statement "default(none)" than the implicit rules. It avoids ambiguity.

For the compiling error in  issue #329, putting the const-qualified variables in the shared list is probably enough to solve the problem.

susilehtola:
But it would appear that breaks compatibility with older compilers, see
http://jakascorner.com/blog/2016/07/omp-default-none-and-const.html

sunqm:
According to the GCC 9 porting instructions  https://gcc.gnu.org/gcc-9/porting_to.html#ompdatasharing , this PR is probably the only workaround for issue #329

PR #331, opened 2019-05-10
change uccsd_lambda einsum

zhutianyu1991:
There are many numpy.einsum's used in uccsd_lambda.py, making OpenMP parallelization very inefficient for large systems. A test case can be seen in "https://github.com/zhutianyu1991/uccsd_lambda_test/". A simple solution might be replacing numpy.einsum with pyscf.lib.einsum, and the test shows OpenMP efficiency is much better with this replacement (~18 folds vs. 3 folds acceleration for benzene/6-31G* on 20 cores).


sunqm:
Some of np.einsum should not be replaced. These operations are memory bounded. Calling lib.einsum does not help improve memory efficiency. It actually leads to the overhead of memory footprint.

Issue #48, opened 2016-10-20
Use libxc's parsing routines instead of hardcoded lists


susilehtola:
Hi,

I just had a look at https://github.com/sunqm/pyscf/blob/master/dft/libxc.py and noticed that you're using hardcoded functional lists.

However, libxc has had the functionality to parse functional keywords and id numbers for a long time: 
int   XC(functional_get_number)(const char *name);
char *XC(functional_get_name)(int number);

which implement the necessary functionality. I would recommend removing XC_CODES, XC_KEYS, LDA_IDS, GGA_IDS, etc from dft/libxc.py, and replacing them with the two above functionals.

The functions is_lda, is_gga and so on can be reimplemented by checking the functional family with 
  int   XC(family_from_id)(int id, int *family, int *number);
or by checking the func.info->family entry of 
  xc_func_type func;
that has been initialized to the used functional. The character of the functional (x, c, or xc) can be similarly checked from func.info->kind.

It doesn't really make sense to replicate this information, since that entails more maintenance. The stuff is already maintained within libxc, where it is always up to date.

Also, I don't see how you handle range separated hybrids. Are they not supported in pyscf?

sunqm:
The functional list XC_CODES is hardcoded because it is designed to be hardcoded.   When you run in python shell, it's straightforward to check what has been supported by typing

>>> from pyscf import dft
>>> dft.lib<tab>.X<tab>_C<tab><enter>    # dft.libxc.XC_CODES
>>> dft.xcf<tab>.X<tab><enter>                 # dft.xcfun.XC_CODES

I don't see any other way more simple and consistent to do this in python shell.

What  is_lda, is_gga functions did is more than a single check for one functional ID, it is a part of the parser function parse_xc,  which allows you customizing functionals like ".218*HF + .073*LDA + .709*MPW91, .871*LYP + .129*VWN".  I don't know whether libxc has this feature in recent version.  But I'm sure it's not when I implemented the libxc.py interface.  They have to be there until I found better solution.

I agree LDA_IDS, GGA_IDS is duplication of XC(family_from_id) function.  I can reimplement it in the future when I optimize the DFT code.

Range separated hybrids are not coded in the dft module.  The required integrals should be all available.  Let me know if you need this functionality.  I can show you where to access the integrals or I can implement it in next release.

Qiming

susilehtola:
[...]

PR #332, opened 2019-05-11
Remove hardcoded functional lists, ask libxc instead.

susilehtola:
This PR removes hardcoded functional lists, and employs libxc's Python interface instead.

Libxc is also updated to version 4.3.4. No functionals have been lost from libxc, but the schemes in which they are offered has changed a bit.

I have also added a warning in case range-separated functionals with a non-error function kernel are requested, because PySCF does not appear to calculate the exchange integrals correctly.


susilehtola:
This would resolve #48.

sunqm:
Sorry for the long silence. Merging this PR potentially requires to update the prerequisite version of cmake (from cmake 2 to cmake 3). Updating the prerequisite of the lowest cmake version has been brought to discussion. Merge should be possible in 1.7 release or later version when the compiling prerequisite is updated.

susilehtola:
Actually, since I ended up not using the libxc python interface at all, one does not need to modify the cmake file, which means this could be included in the next release.

susilehtola:
> Sorry for the long silence. Merging this PR potentially requires to update the prerequisite version of cmake (from cmake 2 to cmake 3). Updating the prerequisite of the lowest cmake version has been brought to discussion. Merge should be possible in 1.7 release or later version when the compiling prerequisite is updated.

CMake version 2 should suffice now.

sunqm:
@susilehtola definitions of exchange functionals m05, m052x, m06, m062x, tpss, bloc, and correlation functionals m05, m052x, m06, m062x, tpss are very different to the old version (v3.0.1). Seems some of the differences are caused by the definition in the new functional, like whether to include HF exchange contribution. Is there a way to access the old definition?

susilehtola:
@sunqm what do you mean by "very different" and "whether to include HF exchange contribution"? The earlier versions exposed the worker functions that did not make any sense at all by themselves, because the functionals had been parametrized as hybrids but the worker functionals lacked the exact exchange contribution; see https://tddft.org/programs/libxc/changes/.

When used correctly, the results should be exactly the same as before.

sunqm:
I think I found them in https://gitlab.com/libxc/libxc/blob/master/src/xc_funcs_removed.h .
Are these functional completely not available?

sunqm:
This PR has been merged in dev branch

susilehtola:
> I think I found them in https://gitlab.com/libxc/libxc/blob/master/src/xc_funcs_removed.h .
> Are these functional completely not available?

The old "definitions" were wrong, since they only included the DFT part and not the exact exchange part. It was very confusing having e.g. the worker functionals `mgga_x_m06`, `mgga_c_m06`, and then the actual functional `hyb_mgga_xc_m06`, which differed from `mgga_x_m06,mgga_c_m06` by the exact exchange parameters. One can still "access" the old version by setting exact exchange to zero without scaling what comes out of libxc.

Issue #382, opened 2019-07-19
Array reshape bug in Cubegen.mep() when used with resolution parameter


mjw99:
With PySCF=1.6.2 and the following input:

```
from pyscf import gto
from pyscf import dft
from pyscf.tools import cubegen

h2o = gto.M(atom='      O 0.00000000,  0.000000,  0.119748;\
                        H 0.00000000,  0.761561, -0.478993;\
                        H 0.00000000, -0.761561, -0.478993',\
                        basis='sto-3g')
h2o.build()
h2o.verbose = 0

method = dft.RKS(h2o)
method.grids.prune = dft.gen_grid.treutler_prune
method.xc = 'b3lypg'
method.scf()

cubegen.mep(h2o, 'water_pot.cube', method.make_rdm1(),resolution=0.1)

```

I see the following error:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-1-fbe17ce7cc5e> in <module>
     16 method.scf()
     17 
---> 18 cubegen.mep(h2o, 'water_pot.cube', method.make_rdm1(),resolution=0.1)

~/.conda/envs/pyscf_mep/lib/python3.7/site-packages/pyscf/tools/cubegen.py in mep(mol, outfile, dm, nx, ny, nz, resolution)
    170 
    171     MEP = Vnuc - Vele     # MEP at each point
--> 172     MEP = MEP.reshape(nx,ny,nz)
    173 
    174     # Write the potential

ValueError: cannot reshape array of size 384480 into shape (80,80,80)
```

PR #386, opened 2019-07-21
Fix for #382 and minor URL correction

mjw99:


sunqm:
Thanks for the PR. Issue 382 has been fixed in a couple of changes made recently. So I close this PR

PR #394, opened 2019-07-29
Polarizable Embedding Model

maxscheurer:
This PR implements the Polarizable Embedding (PE) model through the [CPPE](https://github.com/maxscheurer/cppe) library.

**Features implemented:**
- [x] ground-state (HF, DFT) PE
- [x] PE-TDA excitation energies

**ToDo:**
- [ ] install CPPE through Travis to enable running the functionality test
- [ ] add user examples
- [ ] add test for PE-TDA

**Merge:**
- [ ] Ready for merge


sunqm:
When PE was applied, the function for second-order orbital derivatives (_gen_rhf_response etc in pyscf.soscf.newton_ah) may need to be updated. Before merging this feature, we should include this flexibility in the _gen_rhf_response functions. Then all relevant modules, like stability analysis, second-order SCF, etc can work correctly.

PR #403, opened 2019-08-06
Easier install for IBM i

ThePrez:
Install reliably fails on the IBM i platform unless the `PYSCF_INC_DIR` environment variable is set. I'd rather if a simple `pip3 install pyscf` would work without knowing about this environment variable. 

I fixed it several ways, but setting this env var in setup.py seemed the least intrusive.


ThePrez:
Also, without this change, install blatantly fails with the explicit platform check (not sure how I forgot to mention that)

PR #410, opened 2019-08-19
Add more DKH basis sets

bogdanoff:
"Lighter" basis set optimized for relativistic effects at DKH2 level.
It is useful, because orbital optimization with ANO-RCC of good quality somehow takes very long time with pyscf.

I used `uncontract_general=true` to remove the general contraction present in EMSL.
General contraction doesn't make sense for SARC, as we have here mostly one exponent per contraction.
This way one gets the same data as on [http://dpantazis.org/sarc.html](http://dpantazis.org/sarc.html) (and  in the original papers).


bogdanoff:
I also add missing cc-pv{d,t,q}z-dk basis sets from J.G. Hill and K.A. Peterson, J Chem. Phys. 147, 244106 (2017).

PR #412, opened 2019-09-11
Define long := int when using python3

ehermes:
In [pbc/gto/cell.py](https://github.com/pyscf/pyscf/blob/master/pyscf/pbc/gto/cell.py#L233), there is an `isinstance` check against the type `long`, which is not a built-in type as of Python 3. For compatibility with Python 3, we must define `long = int`. This seems to be the only place in PySCF that currently uses the Python 2.x `long` type, so it's a one-line fix.

### DIFF
--- a/pyscf/pbc/gto/cell.py
+++ b/pyscf/pbc/gto/cell.py
@@ -55,6 +55,7 @@ EXP_DELIMITER = getattr(__config__, 'pbc_gto_cell_split_basis_exp_delimiter',
 # For code compatiblity in python-2 and python-3
 if sys.version_info >= (3,):
     unicode = str
+    long = int
 
 libpbc = _pbcintor.libpbc
 


ehermes:
Also, something tangentially related. I'm currently writing a PySCF wrapper for the Python package ASE ([a link to the merge request on gitlab](https://gitlab.com/ase/ase/merge_requests/1362)). For this MR, I would like to add the [PySCF logo](https://github.com/pyscf/pyscf/blob/master/doc/logo/pyscf-logo.png) to ASE's website. Does ASE have permission to put a copy of this logo on their website?

gkc1000:
That's great that you're writing a wrapper to ASE! I played around a little
with ASE when we were developing the pbc module, it was very nice, and
there is some residual wrapper somewhere in the code used to import crystal
structures.

You can certainly go ahead with using the PySCF logo.

Issue #417, opened 2019-09-28
Install according to README does fails because of missing tddft.org website


gnibeil:
The installation procedure gets the libxc tarball from the URL tddft.org. Since this is down for several weeks the automated installation of the pyscf program does not work.

sunqm:
libxc download is updated in commit 1eb386a072be839633a17ff4b303020774c5d560

PR #418, opened 2019-09-28
added URLs to libxc releases on gitlab.com since tddft.org is longer …

gnibeil:
since tddft.org is longer unavalable and corresponding configuration.
Fixes issue #417


sunqm:
Can the step of autoconf be skipped? Autoconf is not always available in a system.

PR #422, opened 2019-10-14
electron spin-rotation coupling, effective charges including transition-metals

jiachenchem:
New code for electron spin-rotation coupling in /pyscf/prop/esr
Effective charge for two-electron spin-orbit coupling in /pyscf/prop/zfs/uhf.py


sunqm:
merged in dev branch (commit id 3eb25961783b2b08f46a8ec4c53c39b2f2200552)

PR #424, opened 2019-10-22
Periodic EE-EOM and CIS

xwang862:
In this pull request I added the following features:
- Spin-orbital (GHF-based) and spin-restricted (RHF-based) EOM-EE-CCSD with k-point sampling. The spin-restricted version only allows singlet excitation.
- RHF-based CIS with k-point sampling and for singlet excitation. A direct algorithm using density-fitting integrals (and the Davidson method) is implemented. Non-zero exciton momentum is allowed in EE-EOM and CIS.
- A few example files (in `pyscf/examples/pbc/`) and unit tests
- New all-electron basis sets (`pob-tzvp`, `pob-tzvpp`,  and `crystal-cc-pvdz`) designed for solid calculations (available in Crystal)


xwang862:
Hi @sunqm the following links seem to be inaccessible:
- http://www.sunqm.net/pyscf/files/src/libxc-4.3.4.tar.gz
- http://www.sunqm.net/pyscf/files/bin/pyscf-1.7-deps.tar.gz
Can you please confirm?

pyscf/pbc/cc/kccsd_rhf.py
```diff
@@ -728,25 +729,29 @@ def __init__(self, cc, mo_coeff=None, method='incore'):
...
-        # the virtual orbital energies may overlap which may lead to numerical
-        # issue in the CCSD iterations.
-        # FIXME: Whether to add this correction for other exxdiv treatments?
-        # Without the correction, MP2 energy may be largely off the correct value.
-        madelung = tools.madelung(cell, kpts)
-        self.mo_energy = [_adjust_occ(mo_e, nocc, -madelung)
-                          for k, mo_e in enumerate(self.mo_energy)]
+        if cc.keep_exxdiv:
+            self.fock = np.asarray([np.diag(mo_e) for k, mo_e in enumerate(cc._scf.mo_energy)], dtype=dtype)
+            self.mo_energy = [self.fock[k].diagonal().real for k in range(nkpts)]
```
sunqm: There is no guarantee that orbitals diagonalize the Fock matrix. I think it's better to construct the Fock matrix explicitly than using the orbital energies as the diagonal part.

pyscf/pbc/ci/__init__.py
```diff
@@ -28,3 +28,9 @@ def UCISD(mf, frozen=0, mo_coeff=None, mo_occ=None):
 def GCISD(mf, frozen=0, mo_coeff=None, mo_occ=None):
     mf = scf.addons.convert_to_ghf(mf)
     return cisd.GCISD(mf, frozen, mo_coeff, mo_occ)
+
+def KCIS(mf, frozen=0, mo_coeff=None, mo_occ=None, keep_exxdiv=False):
+    mf = scf.addons.convert_to_rhf(mf)
+    return kcis_rhf.KCIS(mf, frozen, mo_coeff, mo_occ, keep_exxdiv)
+
+CIS = KCIS
```
sunqm: Please add a new line here

pyscf/scf/hf.py
```diff
@@ -1516,7 +1516,18 @@ def make_rdm1(self, mo_coeff=None, mo_occ=None, **kwargs):
...
+
+        # If Cell object is fed into molecular mean field `mf`,
+        # self.mol.energy_nuc() will compute nuclear-repulsion energy in
+        # a periodic setting, while mf.energy_elec() computes electronic energy
+        # in a molecular setting.
+        if isinstance(self.mol, pbcgto.Cell):
+            e = mole.energy_nuc(self.mol)
+        else:
+            e = self.mol.energy_nuc()
+        return e
```
sunqm: Let's simply return gto.mole.energy_nuc(self.mol)

gkc1000:
Re: the Mac error, following the stackoverflow thread, I used:

open
/Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg

to properly set up the headers e.g. stdio.h. Then it is fine to use clang
to compile.

pyscf/lib/CMakeLists.txt
```diff
@@ -148,16 +148,16 @@ if(APPLE)
     #GIT_REPOSITORY https://gitlab.com/libxc/libxc/
     #GIT_TAG master
     #URL http://www.tddft.org/programs/octopus/down.php?file=libxc/4.3.4/libxc-4.3.4.tar.gz
-    #URL https://gitlab.com/libxc/libxc/-/archive/4.3.4/libxc-4.3.4.tar.gz
-    URL http://www.sunqm.net/pyscf/files/src/libxc-4.3.4.tar.gz
```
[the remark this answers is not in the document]
xwang862: Agree, fewer is better 👍

pyscf/lib/CMakeLists.txt
```diff
@@ -148,16 +148,16 @@ if(APPLE)
...
+    #URL http://www.sunqm.net/pyscf/files/src/libxc-4.3.4.tar.gz
     PREFIX ${PROJECT_BINARY_DIR}/deps
     INSTALL_DIR ${PROJECT_SOURCE_DIR}/deps
     BUILD_IN_SOURCE 1
     # When Xcode was installed on Mac, CMAKE_C_COMPILER may be set to the
     # Xcode C compiler, which leads to "stdio.h not found" error. See relevant
     # discussion in
     # https://stackoverflow.com/questions/52509602/cant-compile-c-program-on-a-mac-after-upgrade-to-mojave
-    CONFIGURE_COMMAND <SOURCE_DIR>/configure --prefix=<INSTALL_DIR> --libdir=<INSTALL_DIR>/lib
+    CONFIGURE_COMMAND autoreconf -i || autoreconf && <SOURCE_DIR>/configure --prefix=<INSTALL_DIR> --libdir=<INSTALL_DIR>/lib
```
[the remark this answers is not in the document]
xwang862: Fixed

pyscf/pbc/cc/kccsd_rhf.py
xwang862: We wanted to test whether keeping the exxdiv effect in electron correlation will improve the k-point convergence. Anyway, `keep_exxdiv = False` by default and I have hidden it from the constructors (and added a warning comment to avoid user modification). Or, we can simply remove the changes here and keep them in my local branch.

pyscf/pbc/ci/__init__.py
xwang862: Fixed

pyscf/scf/hf.py
xwang862: Fixed

pyscf/pbc/cc/kccsd_rhf.py
sunqm: This comment is about the way to construct self.fock and self.mo_energy. Computing fock matrix with the code
```
self.fock = np.asarray([np.diag(mo_e) for k, mo_e in enumerate(cc._scf.mo_energy)], dtype=dtype)
```
requires canonical HF orbitals. I think it's better to compute Fock from definition as the old code did
```
fockao = cc._scf.get_hcore() + cc._scf.get_veff(cell, dm)
self.fock = np.asarray([reduce(np.dot, (mo.T.conj(), fockao[k], mo))
                                for k, mo in enumerate(mo_coeff)])
```
without the temporary environment statement
```
with lib.temporary_env(cc._scf, exxdiv=None):
```

I agree it's useful to add the flag  keep_exxdiv

xwang862: Now in both cases, Fock is computed from density matrix and Fock AO. `keep_exxdiv` is kept. See commit 2da05c3.

xwang862:
The Travis CI failure comes from mismatched CASCI energies (from test `test_casci_4o4e` of `mcscf/test/test_n2_df.py`) with Python 2.7.15. I tested it on my mac with the same version of python and it passed with no problem. 
I noticed this test has inconsistently passed or failed in previous PRs. Any clue how to fix it?

sunqm:
> The Travis CI failure comes from mismatched CASCI energies (from test `test_casci_4o4e` of `mcscf/test/test_n2_df.py`) with Python 2.7.15. I tested it on my mac with the same version of python and it passed with no problem.
> I noticed this test has inconsistently passed or failed in previous PRs. Any clue how to fix it?

It sometimes converge to wrong state. Not a big issue

PR #453, opened 2019-11-29
Fix off-by-one error in Huckel guess.

susilehtola:
This PR fixes an off-by-one error in the Huckel guess, improves the documentation - orbitals, not density matrices - and splits the implementation into one step that builds the Huckel basis and another that assembles the core matrices.

The minimal atomic basis produced by the routine could be used to replace the `minao` guess as soon as the SAD algorithm supports ECPs.


susilehtola:
Since formation of the minimal orbital basis has already been split into two, this pull is now just a one liner and should be mergeable straight away @sunqm

PR #468, opened 2019-12-12
Update nao.rst

chrinide:
This is the right OpenMX (http://www.openmx-square.org/) for nano-scale material simulations based on density functional theories from T. Ozaki (Univ. of Tokyo)


fishjojo:
fixed in https://github.com/pyscf/pyscf-doc/commit/ab8e3fbc21cfa6251728865a1c7748883b28c99b

PR #477, opened 2020-01-02
Remove linear dependencies in basis set projections

susilehtola:
The SCF addons module's projection utilities are currently doing a linear equation solve, which generates warnings for ill-conditioned basis sets. This PR changes the code so that linear dependencies are removed within the projection.


susilehtola:
Actually, looks like the warnings are generated in the case where the reciprocal condition number is smaller than machine epsilon, meaning that this "fix" would also fail.

PR #492, opened 2020-01-14
Make code Python 3 compliant.

susilehtola:
The PR has straightforward print syntax changes to pyscf/tools/Molpro2Pyscf/MolproXml.py and pyscf/tools/Molpro2Pyscf/example_Pd_atom.py, which should not cause any problems.

The PR also fixes incorrect indentation in pyscf/dmrgscf/p_dmrg.py, for which Python throws an error.


susilehtola:
@sunqm thanks for the other merges; this is the final one for now.

PR #508, opened 2020-01-30
Get rid of SyntaxWarning

h-larsson:


h-larsson:
I see that frozen can also be a list. An ugly fix would be something like ```numpy.all(numpy.asarray(mycc.frozen) == 0)``` or explicit ```isinstance``` checking.

The SyntaxWarning for identity checks occurs for python 3.8:
https://docs.python.org/3.8/whatsnew/3.8.html#changes-in-python-behavior

sunqm:
when .frozen is an iterable object, its elements are the indices of the orbitals to be frozen ( https://github.com/pyscf/pyscf/blob/master/pyscf/cc/ccsd.py#L835) . The statement `numpy.all(frozen==0)` does not give the right check for this definition.

I'd consider the logic like
```
if not ((frozen is None)
        or (isinstance(mycc.frozen, (int, numpy.integer)) and mycc.frozen == 0)
        or (len(mycc.frozen) == 0)):
    ....
```
if isinstance(mycc.frozen, np.ndarray) and len(mycc.frozen)

sunqm:
Merged in commit 3fac20bfc9373191488458ce55bf9905451b788a

PR #561, opened 2020-03-21
Spin density analysis for UHF

jamesETsmith:
I added two new functions in `uhf.py`: `mulliken_spin_pop()` and `mulliken_meta_spin()` (aliased as `mulliken_spin_pop_meta_lowdin_ao()`). I've also added a test to `test_uhf.py` comparing the results from the new functionality to results from another package.

Let me know if there are other features/documentation/testing I should add. I've chosen to keep this function separate from `mf.analyze()` since calling it and returning the spin density would break the previous behaviour of `analyze()` in several scripts in PySCF. 

Thanks to @linusjoonho for pointing out the need for this and writing the `mulliken_spin_pop()` function.

### DIFF
--- a/pyscf/scf/test/test_uhf.py
+++ b/pyscf/scf/test/test_uhf.py
@@ -121,6 +121,13 @@ class KnownValues(unittest.TestCase):
         pop, chg = mf.mulliken_pop_meta_lowdin_ao(mol, dm, pre_orth_method='ano')
         self.assertAlmostEqual(numpy.linalg.norm(pop), 12.322626374896178, 9)
 
+    def test_mulliken_spin_pop(self):
+        Ms_true = [0.990027, 0.504987, 0.504987]
+        _, Ms = mf2.mulliken_spin_pop()
+        self.assertAlmostEqual(Ms[0], Ms_true[0],5)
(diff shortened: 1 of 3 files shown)


sunqm:
Looks good to me. The spin density analysis may be added to uhf analyze function, if you prefer the analysis dumped by default in the UHF calculations. I'll leave this PR a few days for possible updates.

jamesETsmith:
@sunqm would it be ok if `mf.analyze()` called the spin density analysis, but didn't return the values? If users wanted to work with the spin density directly they could call the spin density functions directly. Does that seem like an ok solution?

sunqm:
This solution looks good to me.

On Mon, Mar 23, 2020 at 08:02:48AM -0700, James E T Smith wrote:
> @sunqm would it be ok if `mf.analyze()` called the spin density analysis, but didn't return the values? If users wanted to work with the spin density directly they could call the spin density functions directly. Does that seem like an ok solution?
> 
> -- 
> You are receiving this because you were mentioned.
> https://github.com/pyscf/pyscf/pull/561#issuecomment-602654928

--

jamesETsmith:
I've added the solution I mentioned above. Travis seems to have failed for two builds here, but on my fork, it passed all builds.

PR #565, opened 2020-03-24
Make breaksym behave the same way in atom and huckel guess.

susilehtola:
This PR adds `breaksym` to the `huckel` guess, as well as fixes `breaksym` for the `atom` guess that used to

1. set the alpha density matrix to 1e-2 times the overlap and
2. set the block diagonal of the beta density matrix to that of the alpha matrix

i.e. the `atom` guess was not used at all with `breaksym`!

### DIFF
--- a/pyscf/scf/uhf.py
+++ b/pyscf/scf/uhf.py
@@ -54,10 +54,10 @@ def init_guess_by_atom(mol, breaksym=BREAKSYM):
     dm = hf.init_guess_by_atom(mol)
     dma = dmb = dm*.5
     if breaksym:
-        #Add off-diagonal part for alpha DM
-        dma = mol.intor('int1e_ovlp') * 1e-2
+        #remove off-diagonal part of beta DM
+        dmb = numpy.zeros_like(dma)
         for b0, b1, p0, p1 in mol.aoslice_by_atom():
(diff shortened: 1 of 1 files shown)


fishjojo:
The breaksym option for huckel guess has been added in 7123be0ca4be25e513494f79852e001f181048d7.
The atom guess is ok as `dma` is block diagonalized instead of `dmb`.

susilehtola:
If you use my code, please use my commits...

susilehtola:
I don't see how the atom guess is correct.

fishjojo:
Sorry, I have to add other things besides your code...
For atom guess,
```
def init_guess_by_atom(mol, breaksym=BREAKSYM):
    dm = hf.init_guess_by_atom(mol)
    dma = dmb = dm*.5
    if breaksym:
        #Add off-diagonal part for alpha DM
        dma = mol.intor('int1e_ovlp') * 1e-2
        for b0, b1, p0, p1 in mol.aoslice_by_atom():
            dma[p0:p1,p0:p1] = dmb[p0:p1,p0:p1]
    return numpy.array((dma,dmb))
```
You see that `dma` is set to `dmb`, not the opposite... Note that atom guess is block diagonalized, so here we add small off-diagonal terms to `dma`.

susilehtola:
OK, I guess that's fine.

PR #566, opened 2020-03-24
WIP: implement the superposition of fitted atomic potentials guess

susilehtola:
This PR implements the [fitted version](https://arxiv.org/abs/2002.02587) of the [superposition of atomic potentials](https://pubs.acs.org/doi/10.1021/acs.jctc.8b01089) guess (#564), which only requires three-center two-electron integrals.

Things that are still missing:

1. proper normalization of the contraction coefficients. The contraction coefficients should **not** be normalized by PySCF; instead, the input coefficients should be kept as is. They should sum to `-Z`, the negative of the nuclear charge. Then, before the integrals are calculated, the coefficients `c_p` just have to be multiplied with `(\alpha_p / pi)^{3/2}` so that the density primitives are normalized.
2. handling of ghost atoms; they should be skipped altogether in the auxiliary basis


susilehtola:
I'd appreciate help for the missing pieces.

PR #569, opened 2020-03-24
Sacasscf state gradient

MatthewRHermes:
Gradients and geometry optimizations for specific states in state-average CASSCF calculations. State average mix version not yet available.

### DIFF
--- a/pyscf/grad/casscf.py
+++ b/pyscf/grad/casscf.py
@@ -56,7 +56,8 @@ def grad_elec(mc_grad, mo_coeff=None, ci=None, atmlst=None, verbose=None):
     nao, nmo = mo_coeff.shape
     nao_pair = nao * (nao+1) // 2
 
-    # Necessary kludge because gfock isn't occ-virt space in sacasscf
+    # Necessary kludge because gfock isn't zero in occ-virt space in SA-CASSCf
+    # Among many other potential applications!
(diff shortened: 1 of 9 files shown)


pyscf/grad/casscf.py
```diff
@@ -51,24 +56,34 @@ def grad_elec(mc_grad, mo_coeff=None, ci=None, atmlst=None, verbose=None):
     nao, nmo = mo_coeff.shape
     nao_pair = nao * (nao+1) // 2
 
+    # Necessary kludge because gfock isn't zero in occ-virt space in SA-CASSCf
+    # Among many other potential applications!
+    if hasattr (mc, '_tag_gfock_ov_nonzero'):
+        if mc._tag_gfock_ov_nonzero:
+            nocc = nmo
```
sunqm: Why nocc is condtionally set to nmo? Based on the implementations below, I think it's safe to remove nocc and use nmo to replace it, right?

MatthewRHermes: Well, yes, except this causes some of the tests of the original single-state CASSCF gradient to fail (with error << value but > the threshold of the test scripts) because of the finite convergence threshold of CASSCF iteration. I made it conditional for the sake of strict backwards-compatibility, in other words.

ssh2:
This looks good to me.

PR #579, opened 2020-04-05
Restricted IP-/EA-ADC and efficiency improvements to unrestricted IP-/EA-ADC

Samragni-93:
This PR contributes changes to the ADC module that supports calculation of IP/EA of closed- and open-shell molecules. The new features are:

- Spin-restricted ADC (RADC) for closed-shell molecules 
- The efficiency of the UADC code has been significantly improved by taking advantage of BLAS operations for expensive tensor contractions and reduction in the overall memory usage
- Preconditioner for Davidson’s iterative algorithm has been modified to ensure better convergence
- Modifications to the ADC code interface to allow for a direct calculation of excitation energies and properties
- New example files (pyscf/examples/adc) and unit tests


gkc1000:
Have you compared lib.einsum using tblis, rather than np.einsum(optimize=True)?

ssh2:
> Have you compared lib.einsum using tblis, rather than np.einsum(optimize=True)?

We tried replacing np.einsum with lib.einsum. For most of the contractions, it did not improve efficiency and sometimes increased timings by ~ 10 %. Most of the expensive contractions were implemented using np.dot by forming intermediates, without the use of einsum. There are few contractions where @Samragni-93 ended up using lib.einsum.

gkc1000:
But was lib.einsum also compiled to use tblis? We have generally seen
significant speedups when tblis is used, particularly w.r.t.
multithreading, if I recall correctly.

ssh2:
> But was lib.einsum also compiled to use tblis? We have generally seen significant speedups when tblis is used, particularly w.r.t. multithreading, if I recall correctly.
> […](#)

Not sure. We'll check that and let you know.

ssh2:
> > But was lib.einsum also compiled to use tblis? We have generally seen significant speedups when tblis is used, particularly w.r.t. multithreading, if I recall correctly.
> > […](#)
> 
> Not sure. We'll check that and let you know.

We replaced np.einsum with lib.einsum and tested pyscf compiled with tblis. Indeed, we see up to 15-20% speedup. Thanks for the suggestion. This is addressed in a9ef3721c2d2252a599723fc206d727dc89ba414 .

ssh2:
We are ready to merge this. Any objections?

gkc1000:
Go ahead and merge

PR #592, opened 2020-04-24
allow passing breaksym in arguments for kmf calculations

sunchong137:
Hi there,

I am using the KUHF function in pbc.scf to do some k-space mean-field calculations. I found that the initial guess of KUHF breaks spin symmetry by default, which causes unphysical results in my calculations. I tried to turn it off but found that breaksym argument is defined as an environment variable, which makes this option very inflexible. Therefore, I added the argument "breaksym" to the get_init_guess() function with the minimal change to the code. I only made changes to the parts that I am using.

Thanks,
Chong

### DIFF
--- a/pyscf/pbc/scf/kuhf.py
+++ b/pyscf/pbc/scf/kuhf.py
@@ -387,7 +387,7 @@ class KUHF(khf.KSCF, pbcuhf.UHF):
                     'alpha = %d beta = %d', *self.nelec)
         return self
 
-    def get_init_guess(self, cell=None, key='minao'):
+    def get_init_guess(self, cell=None, key='minao', breaksym=False):
         if cell is None:
             cell = self.cell
         dm_kpts = None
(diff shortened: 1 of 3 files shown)


fishjojo:
You can add a `.pyscf_conf.py` file at your working directory or HOME directory in which you set 
`scf_uhf_init_guess_breaksym=False`

sunchong137:
Hi there,
Thanks for taking care of the pull request.
I figured out that if I only make changes to the get_init_guess() functions that I used, other functions will get errors when called, so the pull request should be closed. 
However, as I wrote in the pull request comments, setting the environment variable will make calculations ONLY break symmetry or not break symmetry, I have to change scf_uhf_init_guess_breaksym constantly when submitting jobs, which is very inconvenient.

Thanks,
Chong

fishjojo:
@sunchong137 
I have added an attribute `init_guess_breaksym` for UHF. So now you can specify whether or not to use symmetry broken initial guesses as follows:
```
kmf = pbc.scf.KUHF(cell, kpts)
kmf.init_guess_breaksym = False
kmf.kernel()
```

PR #604, opened 2020-05-08
Bare bones support for libxc development branch

susilehtola:
The current development branch of libxc has reorganized the handling of hybrid functionals to allow functionals like HISS to be implemented. The branch also includes support for double hybrids.

With this patch, the libxc master branch can be used with PySCF.

The patch also restores earlier functionality that allows reading in the list of functionals from libxc, which is now handled via a `__config__` parameter.


susilehtola:
Could this be merged?

PR #658, opened 2020-07-06
Newton casscf dev

MatthewRHermes:
Enable newton_casscf and SA-CASSCF gradient application to the state_average_mix MC-SCF solver (state_average_mix is an important feature that other programs like OpenMolcas don't have). This involved tightening the loop over states in newton_casscf.py, which also increases efficiency because the inactive-virtual orbital rotation diagonal block of the Hessian is identical for all states. It also involves writing new functions "states_make_rdm*" which I've added to both state_average and state_average_mix FCI solvers.


MatthewRHermes:
Not sure how to approach the failing checks for the Python 3.6 and Python 3.7 cases, since those tests don't fail locally for me and I'm using Python 3.7.3 myself.

sunqm:
It might be a bug of eigh function in scipy 1.5. wheels for python3.6 and python3.7 . I can occasionally catch the problem but not yet find a way to reproduce. I'll handle the PR when the scipy problem is solved.

MatthewRHermes:
Well I merged the latest dev myself and it works now...

PR #699, opened 2020-09-07
Actions/workflow CI

mjw99:
Initial actions/workflow CI.

This should mimic what was previously being done in the Travis CI. It should also form the basis for further actions such as linting, deployment etc.

Example of this running can be found [here](https://github.com/mjw99/pyscf/actions/runs/243418711).


fishjojo:
looks ok to me

PR #700, opened 2020-09-08
Batching of integrals for PE and PE(ECP) repulsive potentials

maxscheurer:
For large environments, `pyscf` can run into memory issues in PE calculations. This PR splits up the integral computations into individual batches, depending on how much memory is available.

Furthermore, this PR adds the feature to use repulsive potentials on the environment sites to avoid electron spill-out (https://doi.org/10.1021/acs.jctc.9b01162).


maxscheurer:
I have no idea why the tests are failing... can anyone help me with this please? 😄

maxscheurer:
The memory estimation for PE is still a bit unreliable, so I'm going to make some tweaks before this can be merged.

PR #737, opened 2020-10-14
Add warning and screen negative norms in SC-NEVPT2

bogdanoff:
It turns out that even with deterministic solvers, some of the perturber states have negative norm.
Now all the contributions with negative norms are screened and a warning is printed.

Here is an output of CAS(2,2) case for system including two Cu2+ with cc-pvdz basis
```
CASCI E = -3851.42285404094  E(CI) = -1.21547445991519  S^2 = 0.0000000
Density matrix diagonal elements [1.08742974 0.91257026]
Inactive and virtual orbitals were pseudo-canonicalized for NEVPT
Sr    (-1)',   E = -0.00003725711411
Si    (+1)',   E = -0.00002053164168
Sijrs (0)  ,   E = -2.29929845036311

WARN: 26075 perturber states have negative norm
Largest negative norm is -0.00032381144271730663
Sum of negative norms is -0.021714265535472073

Sijr  (+1) ,   E = -0.10135929403462

WARN: 66729 perturber states have negative norm
Largest negative norm is -0.00032158452641913254
Sum of negative norms is -0.04182706487968975

Srsi  (-1) ,   E = -0.16495261735336
Srs   (-2) ,   E = -0.00006119692458
Sij   (+2) ,   E = -0.00004437366205
Sir   (0)' ,   E = -0.02011240527878
Nevpt2 Energy = -2.585886126372277
:: NEVPT2 for sing -3854.008740
```
But even C2 example gives negative norms.

This PR is only the first step to the actual fix. The negative norms show up for Sijr(+1) Srsi(-1), which seem to be prone to numerical instabilities. @sunqm , do you have an idea?

### DIFF
--- a/pyscf/mrpt/nevpt2.py
+++ b/pyscf/mrpt/nevpt2.py
@@ -323,7 +323,8 @@ def Sr(mc,ci,dms, eris=None, verbose=None):
         +  numpy.einsum('ipqr,rpqa,ia->i',h2e_v,dm2,h1e_v)*2.0\
         +  numpy.einsum('ip,pa,ia->i',h1e_v,dm1,h1e_v)
 
-    return _norm_to_energy(norm, ener, mc.mo_energy[nocc:])
+    return _norm_to_energy(norm, ener, mc.mo_energy[nocc:],
+                           log=logger.Logger(mc.stdout, logger.WARN))
 
 def Si(mc, ci, dms, eris=None, verbose=None):
     #Subspace S_i^{(1)}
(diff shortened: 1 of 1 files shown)


sunqm:
Good to have this check. I don't understand why this happens. Is there a theoretical reason or a signal of bug in the program? Is this observed in other NEVPT2 implementations?

gkc1000:
The norm has to be positive, but the way it is evaluated in the code
involves lots of addition and subtraction. Someone should really check
against DALTON or Molpro ...

Issue #754, opened 2020-11-03
Molden file loading


fockeke:
Loading Molden files is a feature that I use a lot. I have now run into an problem. The Molden program reads the different inputs in any order. Because of this, programs like Dalton write them in any order. However, PySCF expects them in a certain order, at least to a certain degree. PySCF cannot read Molden files where the [Atoms] section comes after the [GTO] section.
This problem is reproducible with any Molden file by switching the sections around.

PR #757, opened 2020-11-09
Molden file loading

fockeke:
MOs and GTOs are enforced to be read after Atoms. This allows the file to have any order. See #754 
The sections are stored in a dictionary.

Additionally, Dalton does write MO coefficients below a certain threshold. Therefore, the value zero has to be added for them. Otherwise the creation of numpy arrays fails.

### DIFF
--- a/pyscf/tools/molden.py
+++ b/pyscf/tools/molden.py
@@ -210,6 +210,8 @@ def _parse_gto(lines, envs):
 # * Do not use iter() here. Python 2 and 3 are different in iter()
     def read_one_bas(lsym, nb, fac=1):
         fac = float(fac)
+        if fac == float(0):
+            fac = float(1)
         bas = [lib.param.ANGULARMAP[lsym.lower()],]
         for i in range(int(nb)):
             dat = _d2e(next(lines_iter)).split()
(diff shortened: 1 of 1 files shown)


fockeke:
The official info about the molden format says, that the factor for GTOs is deprecated. Maybe this is the reason for the NWChem developers to set the factor to zero, when NWChem writes molden files.
PySCF still reads the factor and uses it in calculations. I do not know, whether this serves a purpose right now. Because 0 never makes sense as a factor and results in a division by zero error, I check for the factor being 0 and set it to the default value of 1 if it actually is 0.

sunqm:
The original implementation was based on certain specific samples obtained from
maybe molpro or other code. The GTO factor was not considered carefully there.
This update looks reasonable to me. 

On Thu, Nov 19, 2020 at 03:14:22AM -0800, fockeke wrote:
> The official info about the molden format says, that the factor for GTOs is deprecated. Maybe this is the reason for the NWChem developers to set the factor to zero, when NWChem writes molden files.
> PySCF still reads the factor and uses it in calculations. I do not know, whether this serves a purpose right now. Because 0 never makes sense as a factor and results in a division by zero error, I check for the factor being 0 and set it to the default value of 1 if it actually is 0.
> 
> -- 
> You are receiving this because you commented.
> https://github.com/pyscf/pyscf/pull/757#issuecomment-730302811

--

Issue #773, opened 2020-12-01
DeprecationWarning in imp.py


mbarbry:
Hi,

running some tests, I got the following DeprecationWarning. Any reason to not fix this?

    /media/marc-pc/HDD1/programs/pyscfMain/pyscf/pyscf/__init__.py:71: in <module>
        from pyscf import lib
    /media/marc-pc/HDD1/programs/pyscfMain/pyscf/pyscf/lib/__init__.py:24: in <module>
        from pyscf.lib import numpy_helper
    /media/marc-pc/HDD1/programs/pyscfMain/pyscf/pyscf/lib/numpy_helper.py:27: in <module>
        from pyscf.lib import misc
    /media/marc-pc/HDD1/programs/pyscfMain/pyscf/pyscf/lib/misc.py:25: in <module>
        import imp
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

        """This module provides the components needed to build your own __import__
        function.  Undocumented functions are obsolete.
        
        In most cases it is preferred you consider using the importlib module's
        functionality over this module.
        
        """
        # (Probably) need to stay in _imp
        from _imp import (lock_held, acquire_lock, release_lock,
                          get_frozen_object, is_frozen_package,
                          init_frozen, is_builtin, is_frozen,
                          _fix_co_filename)
        try:
            from _imp import create_dynamic
        except ImportError:
            # Platform doesn't support dynamic loading.
            create_dynamic = None
        
        from importlib._bootstrap import _ERR_MSG, _exec, _load, _builtin_from_name
        from importlib._bootstrap_external import SourcelessFileLoader
        
        from importlib import machinery
        from importlib import util
        import importlib
        import os
        import sys
        import tokenize
        import types
        import warnings
        
    >   warnings.warn("the imp module is deprecated in favour of importlib; "
                      "see the module's documentation for alternative uses",
                      DeprecationWarning, stacklevel=2)
    E   DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses

    /home/marc-pc/anaconda3/lib/python3.8/imp.py:31: DeprecationWarning

fishjojo:
`imp.lock_held()` is used. I guess this is only relevant for python2.

[...]

PR #777, opened 2020-12-07
import imp now compatible with python 3

mbarbry:
Solves issue https://github.com/pyscf/pyscf/issues/773.


mbarbry:
Dear @sunqm,

these changes are ready. They are just some minor corrections for Python3 supports.

PR #810, opened 2021-01-27
Update Libxc to version 5.1.2

susilehtola:
Libxc 5.1.2 is a huge upgrade, bringing 

- loads of new functionals like r2SCAN
- 4th order derivatives for all functionals,  and 
- improved numerical stability for especially range-separated functionals


susilehtola:
Thanks Qiming!

PR #815, opened 2021-02-02
Locally alias time functions in logger

### LINKED ISSUE #814: Importing pyscf.lib.logger breaks time reporting
Since https://github.com/pyscf/pyscf/commit/aec4b6f5125345afd26f263e7f8224505f451bfb whenever you import `pyscf.lib.logger` on python >=3.8 the reported time by anything that relies on system time no longer returns the correct time. You can see this most clearly with something like:

```python3
from datetime import date
print(date.today())
from pyscf.lib import logger
print(date.today())
```

The cause of this is that setting `time.time` to be `time.process_time` is incorrect and breaks any function that relies on getting system time (which includes most of the datetime module). Overriding members in a stdlib module have global impact and will break others code in unexpected ways.

mtreinish:
This commit changes how the compatibility aliasing is done in the
pyscf logger. To accommodate differing time APIs the module was
previously overriding methods in stdlib time to accommodate changes in
behavior between python 2.x and newer version of python 3.x. However,
changing functions and overriding methods in an external module
(especially a stdlib module) is problematic and can break other
people's code using pyscf depending on documented behavior of the
external libraries. To fix this and not cause unexpected side effects
for users this changes how this compatibility aliasing is done, it uses
a local alias in the module and sets things to call the correct method
depending on the python version.

Fixes #814

### DIFF
--- a/pyscf/adc/radc.py
+++ b/pyscf/adc/radc.py
@@ -19,7 +19,7 @@
 '''
 Restricted algebraic diagrammatic construction
 '''
-import time
+
 import numpy as np
 import pyscf.ao2mo as ao2mo
 from pyscf import lib
@@ -35,7 +35,7 @@ def kernel(adc, nroots=1, guess=None, eris=None, verbose=None):
     if adc.method not in ("adc(2)", "adc(2)-x", "adc(3)"):
        raise NotImplementedError(adc.method)
 
-    cput0 = (time.clock(), time.time())
+    cput0 = (logger.process_clock(), logger.perf_counter())
     log = logger.Logger(adc.stdout, adc.verbose)
     if adc.verbose >= logger.WARN:
         adc.check_sanity()
--- a/pyscf/adc/radc_ao2mo.py
+++ b/pyscf/adc/radc_ao2mo.py
@@ -20,7 +20,7 @@ import numpy as np
 import pyscf.ao2mo as ao2mo
 from pyscf import lib
 from pyscf.lib import logger
-import time
+
 import h5py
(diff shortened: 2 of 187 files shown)


sunqm:
Could you resubmit the patch to branch 2.0? thanks!

On Tue, Feb 02, 2021 at 06:29:31AM -0800, Matthew Treinish wrote:
> This commit changes how the compatibility aliasing is done in the
> pyscf logger. To accommodate differing time APIs the module was
> previously overriding methods in stdlib time to accommodate changes in
> behavior between python 2.x and newer version of python 3.x. However,
> changing functions and overriding methods in an external module
> (especially a stdlib module) is problematic and can break other
> people&#39;s code using pyscf depending on documented behavior of the
> external libraries. To fix this and not cause unexpected side effects
> for users this changes how this compatibility aliasing is done, it uses
> a local alias in the module and sets things to call the correct method
> depending on the python version.
> 
> Fixes #814
> 
>   https://github.com/pyscf/pyscf/pull/815
> 
> -- Commit Summary --
> 
>   * Locally alias time functions in logger
> 
> -- File Changes --
> 
>     M pyscf/lib/logger.py (22)
> 
> -- Patch Links --
> 
> https://github.com/pyscf/pyscf/pull/815.patch
> https://github.com/pyscf/pyscf/pull/815.diff
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> https://github.com/pyscf/pyscf/pull/815

--

mtreinish:
Superseded by #817

PR #829, opened 2021-03-02
Molden: transpose missing

fockeke:
The checks for the functionality provided by molden.py no longer fail.
Thanks for merging and the improvements!


fockeke:
Fixed by 6f8aab8 in a better way.

PR #833, opened 2021-03-05
Example on how to restructure the examples directory

susilehtola:
As was just discussed on Slack, it'd be good to refactor the examples in a way that they can be included verbatim in the user documentation. This means removing any redundant, lower-level entries in the example code so that the example can focus on the task at hand.

The way I see this is that the first examples discuss the `gto` module and show how to build molecules in PySCF. Next, in the SCF examples, the code can just `import` the molecules from the `gto` examples, and discuss how to set up SCF calculations. Similarly, in MP2 and CC examples, the HF calculation can be `import`ed from the SCF example, so the only code that is necessary is the bit to do the MP2/CC calculation.

The benefits of this approach are that

-  the example codes will be more focused on the task at hand,
-  they'll become shorter so they'll easily fit in a PDF manual,
-  there'll be no code duplication between the PySCF examples and the user manual,
-  the examples in the manual will be included in the continuous integration tests, and
-  the users can easily run the examples in the manual just by running the PySCF example file.

### DIFF
--- a/examples/scf/00-simple_hf.py
+++ b/examples/scf/00-simple_hf.py
@@ -11,29 +11,13 @@ A simple example to run HF calculation.
 '''
 
 import pyscf
+from systems import mol_HF_ccpvdz
 
-mol = pyscf.M(
-    atom = 'H 0 0 0; F 0 0 1.1',  # in Angstrom
-    basis = 'ccpvdz',
-    symmetry = True,
-)
-
-myhf = mol.HF()
+myhf = mol_HF_ccpvdz.HF()
 myhf.kernel()
-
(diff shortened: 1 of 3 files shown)


januseriksen:
This looks like a great solution, @susilehtola!

susilehtola:
Well yes, but we sort of agreed on Slack it's best to not modularize the examples and instead keep them the way they are; even if this means that the user docs have code snippets that aren't tested..

januseriksen:
I guess that's a valid point as well.
(how does one get on the PySCF Slack channel btw?)

PR #838, opened 2021-03-11
allow smearing for molecules

fishjojo:


pyscf/scf/addons.py
```diff
@@ -26,12 +26,16 @@
 from pyscf.gto import mole
 from pyscf.lib import logger
 from pyscf.scf import hf
+from pyscf.pbc.scf import addons as pbcaddons
```
sunqm, line 29: define smearing function and make a lazy import inside the function. PBC module has complicated module dependencies. import at scf module level may break "import scf"

PR #848, opened 2021-03-19
Various minor DFT changes

januseriksen:
Various minor DFT changes related to the use of PySCF in a project of mine (https://github.com/januseriksen/decodense):

 - Commit #2901915 - This replaces pyscf#610 (introduces identical functionality).
 - Commit #bd1dbe8 - This is related to calling the code in ```numint.py:eval_rho()``` from within multiprocessing regions. Pickling of open files (e.g., in a ```mol``` object) is an issue and we may then allow for passing ```nbas``` and ```ao_loc``` explicitly instead (with ```mol``` set to ```None```). Also, calling into ```lib.dot()``` doesn't appear to work - at least locally on my Mac(?) - and we can then enforce the use of ```np.dot()``` instead by setting the ```ENFORCE_NUMPY``` config variable.

I trust these changes are not disruptive.

### DIFF
--- a/pyscf/dft/numint.py
+++ b/pyscf/dft/numint.py
@@ -622,8 +622,8 @@ def _dot_ao_dm(mol, ao, dm, non0tab, shls_slice, ao_loc, nbas, out=None):
        ao.ctypes.data_as(ctypes.c_void_p),
        dm.ctypes.data_as(ctypes.c_void_p),
        ctypes.c_int(nao), ctypes.c_int(dm.shape[1]),
-       ctypes.c_int(ngrids), pnbas,
-       pnon0tab, pshls_slice, ctypes.c_int(nbas))
+       ctypes.c_int(ngrids), ctypes.c_int(nbas),
+       pnon0tab, pshls_slice, pao_loc)
     return vm
 
 def _scale_ao(ao, wv, out=None):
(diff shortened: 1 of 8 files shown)


januseriksen:
Alright, I think (hope) I've cleaned everything up by now.

sunqm:
The original API seems enough to implement the EDA formula of the paper (IJQC, 109, 2464). It's not obvious to me that the extra keywords idx, nbas, ao_loc are inevitable for the EAD implementation.

* It's recommended to modify the grids' coordinates before calling `eval_ao`. For Grid-EDA, I can imagine a short implementation that defines a set of custom classes inherited from `pyscf.dft.Grids` for each EDA species, then calling `numint.Numint(mol).nr_rks(mol, custom_grids, xc_code)` . The return value exc is then the sum of E^A(DFTX) and E^A(DFTC). If individual X and C contributions are required, they can be evaluated  by calling nr_rks function twice with different xc code, for example
```
# xc = 'b88,p86'
ex_A = nr_rks(mol, custom_grids, xc='b88,')[1]
ec_A = nr_rks(mol, custom_grids, xc=',p86')[1]
```
The parameter idx seems not needed.

* Mole object (and Cell) object can be serialized and deserialized with .pack and .unpack function. For example,
```
pyscf.gto.loads(pickle.loads(pickle.dumps(mol.dumps())))
```
For multiprocessing executors, using mol.dumps() as the argument would be a better choice.

* What's the error of lib.dot ? Do you install pyscf from pip or build from source code?

januseriksen:
I took the time to restructure things on my side over the weekend and so these changes are not needed anymore. Thanks for your comments. Closing this & pyscf#610 as a result.

PR #860, opened 2021-03-25
Check for empty set bug fix in einsum

verena-neufeld:
Bug fix when checking for empty set when using lib.einsum. I am still working on understanding the einsum functionality better in lib.einsum but this fix seems to speed up some of my calculations. Thanks to @tberkel for helpful explanations.

### DIFF
--- a/pyscf/lib/numpy_helper.py
+++ b/pyscf/lib/numpy_helper.py
@@ -131,7 +131,7 @@ def _contract(subscripts, *tensors, **kwargs):
     # Find the shared indices being summed over
     shared_idxAB = uniq_idxA.intersection(uniq_idxB)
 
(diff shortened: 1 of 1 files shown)


verena-neufeld:
This seems to have caused test failures.

verena-neufeld:
I was able to reproduce a segmentation fault on my local machine in "pyscf/lib/gto/test/test_ecp.py", not sure whether this is it, still investigating.

sunqm:
> I was able to reproduce a segmentation fault on my local machine in "pyscf/lib/gto/test/test_ecp.py", not sure whether this is it, still investigating.

May be caused by the size of pre-allocated buffer

verena-neufeld:
Thank you!

PR #861, opened 2021-03-25
Adds update of Ewald parameters to pbc.tools._build_supcell_

Majokux:
Since ce16dc85f `pbc.tools.super_cell` no longer calls the build function of the super cell - as a result, the Ewald parameters `_ew_eta` and `_ew_cut` do not get updated to adjust to the change of `cell.mesh`. This leads to inconsistent nuclear energies (per primitive cell), as you increase the super cell size. This PR adds the update of the Ewald parameters to `_build_supcell_`.

A potential issue is that the user may have specified `_ew_eta` and `_ew_cut` themselves in the original cell - `_ew_from_build` could be checked for this. In this case neither leaving the parameters as they are, nor updating them behind the users back seem like great options to me. Is it possible to work out the equivalent supercell value from the equations in `get_ewald_params` from the values the user had chosen originally?

### DIFF
--- a/pyscf/pbc/tools/pbc.py
+++ b/pyscf/pbc/tools/pbc.py
@@ -594,6 +594,9 @@ def _build_supcell_(supcell, cell, Ls):
     supcell._atm = np.asarray(_atm, dtype=np.int32)
     supcell._bas = np.asarray(_bas.reshape(-1, BAS_SLOTS), dtype=np.int32)
     supcell._env = _env
+
+    # Update Ewald parameters
+    supcell._ew_eta, supcell._ew_cut = supcell.get_ewald_params(supcell.precision, supcell.mesh)
     return supcell
 
 
--- a/pyscf/pbc/tools/pbc.py
+++ b/pyscf/pbc/tools/pbc.py
@@ -596,7 +596,7 @@ def _build_supcell_(supcell, cell, Ls):
     supcell._env = _env
 
     # Update Ewald parameters
-    if hasattr(supcell, "_ew_eta"):
+    if hasattr(supcell, "_ew_eta") and hasattr(supcell, "get_ewald_params"):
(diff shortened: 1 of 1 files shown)


Majokux:
Is there a more canonical way to distinguish between `gto.mole` and `pbc.gto.cell`? `isinstance` is not pythonic and `hasattr(cell, "a")` is asking for trouble.

verena-neufeld:
I wonder whether this test failure is related to my PR (#860)? It is also a segmentation fault. Could be unrelated of course.

sunqm:
> Is there a more canonical way to distinguish between `gto.mole` and `pbc.gto.cell`? `isinstance` is not pythonic and `hasattr(cell, "a")` is asking for trouble.

It's okey to test isinstance(cell, pbc.gto.Cell) I think

sunqm:
> Since [ce16dc8](https://github.com/pyscf/pyscf/commit/ce16dc85ffdfe8d6ec6365883ce707cb6294fafd) `pbc.tools.super_cell` no longer calls the build function of the super cell - as a result, the Ewald parameters `_ew_eta` and `_ew_cut` do not get updated to adjust to the change of `cell.mesh`. This leads to inconsistent nuclear energies (per primitive cell), as you increase the super cell size. This PR adds the update of the Ewald parameters to `_build_supcell_`.
> 
> A potential issue is that the user may have specified `_ew_eta` and `_ew_cut` themselves in the original cell - `_ew_from_build` could be checked for this. In this case neither leaving the parameters as they are, nor updating them behind the users back seem like great options to me. Is it possible to work out the equivalent supercell value from the equations in `get_ewald_params` from the values the user had chosen originally?

A better solution might be to generate the ewald parameters whenever they are needed unless users specified them in the input. The cost to generate these parameters is zero

Majokux:
> A better solution might be to generate the ewald parameters whenever they are needed unless users specified them in the input. The cost to generate these parameters is zero

This is much better indeed. I gave it a shot in #869 and will close this

PR #863, opened 2021-03-25
Corrections to the Sijr and Srsi classes in SC-NEVPT2

pinskip:
I compared the SC-NEVPT2 energies between PySCF and Dalton for a few molecules, including O3 with a (12, 9) space and [Fe(CO)_3NO]^- with a (10, 10) space.

Upon converging CASSCF as tightly as possible, there were deviations between PySCF and Dalton in the ballpark range of 10-30 Microhartree for the Sijr and Srsi classes. With this fix, all energies agree now to well within a Microhartree.

I can make PySCF and Dalton input files available if desired.

### DIFF
--- a/pyscf/mrpt/nevpt2.py
+++ b/pyscf/mrpt/nevpt2.py
@@ -439,7 +439,7 @@ def Sijr(mc, dms, eris, verbose=None):
     h = 2.0*numpy.einsum('rpji,raji,pa->rji',h2e_v,h2e_v,a3)\
          - 1.0*numpy.einsum('rpji,raij,pa->rji',h2e_v,h2e_v,a3)
     h += h.transpose(0, 2, 1)
-    h[:, ci_diag[1], ci_diag[1]] *= 0.5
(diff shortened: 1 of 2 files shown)


gkc1000 (commented):
Thanks for hunting this down - this has been a concern for a while.

Re: the comment on memory usage, memory usage is the same as before the commit?

pinskip:
Memory usage should remain essentially the same as before, provided that numpy returns views rather than new arrays in the additional lines with transpositions and slicing.
What I had in mind: in these two classes, only part of `norm`, `h` and `diff` actually get passed on to `_norm_to_energy`, for example only entries with `r<=s` for the `Srsi` class. If memory consumption was a concern at that point, which it may very well not be due to other bottlenecks, I think it could be halved relatively easily; albeit at the expense of making the algorithm somewhat more complicated.

gkc1000 (approved):

gkc1000:
Thanks for this tracking down this problem!

PR #868, opened 2021-03-26
function to localize orbitals via Cholesky factorization

pinskip:
This commit contains a small function to calculate localized orbitals via Cholesky factorization (including unit tests and an example). It would probably make a nice addition to the already existing localization methods.


pinskip:
Withdrawing the pull request for now due to failing unit tests (the CI uses an old scipy version...).

pinskip:
Re-opening pull request after modifications

gkc1000:
Peter, this is great!

I think for pyscf2, we can require scipy > 1.14.

Peter, if no-one else comments we will accept after issuing PR also for
userdocs for the lo module -- probably just add one or two lines to it.
(Currently, lo userdocs under review so maybe next week?)

sunqm:
What about `pyscf/lib/scipy_helper.py`. Otherwise `pyscf/lib/linalg_helper` is the module for linear algebra algorithms

pinskip:
Actually, I wasn't aware of the call to dpstrf in scf.addons. I've moved both the Lapack wrapper and the hand-written routine to a new file pyscf/lib/scipy_helper.py, as the numpy and linalg helper were already somewhat crowded (which is probably what Qiming suggested, anyway).

I've also modified the Boys class, so that it now accepts ` init_guess = 'cholesky'`. The atomic guess is still the default. Orca, for example, has been using Cholesky orbitals as the default guess for its iterative localization methods for several years now.

I've added a bit of documentation in the pull request https://github.com/pyscf/pyscf-doc/pull/46 to pyscf-doc. One part I wasn't sure about is the column is compatibility with pbc, which I am not familiar with at all. For now I've written "no", as there are no special provisions made, but maybe someone more knowledgeable can double-check.

PR #875, opened 2021-03-29
Add check for Laplacian for libxc functionals

### LINKED ISSUE #832: PySCF gives wrong results for Laplacian dependent functionals
PySCF is not currently checking whether a functional depends on the electron density laplacian, and thereby gives incorrect results for e.g. the SCAN-L family of functionals.

susilehtola:
Closes #832


susilehtola:
I've also removed some code that shouldn't be necessary anymore in the Libxc 5.1 series, since we've eliminated a lot of NaN related errors by rewriting the functionals to be numerically stable.

PR #879, opened 2021-04-01
Adding utils to make NOs from restricted and unrestricted methods.

jamesETsmith:
# Summary
This PR adds a new function to `mcscf.addons` called `make_natural_orbitals` to facilitate running CASCI/CASSCF calculations. 

## Motivation
Based on the [discussion](https://github.com/pyscf/pyscf-doc/pull/35#discussion_r605115424) going on in `pyscf-doc`, it would be nice to show users how to setup CASCI calculations based on MP2/CISD natural orbitals. It's only a few lines of math to make natural orbitals, but I thought it would be a good idea to add some simple tools to make this as easy as possible for users.

## Testing
 I've added two new tests which I checked against ORCA to ensure that the natural orbital occupations my tools produce is accurate.


pyscf/mcscf/addons.py
```diff
@@ -307,6 +307,46 @@ def sort_mo_by_irrep(casscf, mo_coeff, cas_irrep_nocc,
...
+        Dm = rdm1
+    else:
+        raise ValueError("\n\tThe rdm1 generated by method_obj.make_rdm1() was a {}.".format(type(rdm1))
+            + 
+            "\n\tThis type is not supported, please select a different method and/or open an issue at https://github.com/pyscf/pyscf/issues"
+        )
+
+    # Diagonalize the DM in AO (using Eqn. (1) referenced above)
+    A = reduce(numpy.dot, (S, Dm, S))
+    w, v = scipy.linalg.eigh(A, b = S)
```
susilehtola, line 341: You shouldn't use scipy here; you should use the method's `eig` which may have been overwritten

susilehtola: This implementation is not safe in linearly dependent basis sets.

A better alternative would be to diagonalize in the basis of the molecular orbitals, which form an orthonormal set.

pyscf/mcscf/addons.py
```diff
@@ -307,6 +307,46 @@ def sort_mo_by_irrep(casscf, mo_coeff, cas_irrep_nocc,
...
+
+    # Flip NOONs (and NOs) since they're in increasing order
+    noons = numpy.flip(w)
+    natorbs = numpy.flip(v, axis=1)
```
susilehtola, line 345: An easier way would be to diagonalize the negative of the density matrix

PR #888, opened 2021-04-07
Cleanup FNOCCSD

tberkel:


examples/cc/22-fnoccsd.py
```diff
@@ -15,16 +15,29 @@
 mol.basis = 'cc-pvqz'
 mol.build()
 
+# test code correctness
 mf = scf.RHF(mol).run()
 mycc = cc.CCSD(mf).run()
 print(mycc.e_corr - -0.3170511898840137)
+nvir = mycc.nmo-mycc.nocc
```
verena-neufeld, line 22: I would make this more explicit, e.g. `nvir_mo`?

pyscf/mp/mp2.py
```diff
@@ -199,6 +199,14 @@ def make_fno(mp, thresh=1e-6, pct_occ=None, nvir_act=None, t2=None):
...
     Frozen natural orbitals
 
+    Attributes:
+        thresh : float
+            Threshold on NO occupation numbers. Default is 1e-6 (very conservative).
+        pct_occ : float
+            Percentage of total occupation number. Default is None. If present, overrides `thresh`.
+        nvir_act : int
+            Number of virtual NOs to keep. Default is None. If present, overrides `thresh` and `pct_occ`.
+
```
verena-neufeld, line 209: Great to see this added! For completeness, `t2` should probably very briefly be described, too.

verena-neufeld (commented):
Thanks for all this clean up! 
- I am getting this warning: `<class 'pyscf.cc.ccsd.CCSD'> does not have attributes  delta_emp2` (but this is not a big issue)
- Is there a danger that people will have been using the files you are removing, breaking backwards compatibility or do you consider them very niche or faulty so they should be removed?
Thanks!

verena-neufeld:
I think the warning can stay unless you think of a quick way to not make it appear (I felt I should point it out in the off chance you missed it). Renaming `nvir` is a personal taste, you can choose to ignore it :D.

sunqm (approved):
Just to ensure, rccsd_hybrid code was not accidentally deleted?

tberkel:
Correct, not an accident. That was development code that accidentally got
included in an earlier PR. I’m sure no one is using it.

PR #903, opened 2021-04-13
WIP: Print out libxc version info in calculations

susilehtola:
DFT calculations should print out information on the functional library used in the calculation. The information should include the name and version of the library, as well as the literature reference. The info should be printed out in combination with the XC functional; libxc can also give out the reference(s) to the XC functional that should likewise be printed out.

This PR implements the necessary changes to the code.

### DIFF
--- a/pyscf/dft/libxc.py
+++ b/pyscf/dft/libxc.py
@@ -50,6 +50,10 @@ _itrf.LIBXC_hybrid_coeff.restype = ctypes.c_double
 _itrf.LIBXC_nlc_coeff.argtypes = [ctypes.c_int,ctypes.POINTER(ctypes.c_double)]
 _itrf.LIBXC_rsh_coeff.argtypes = [ctypes.c_int,ctypes.POINTER(ctypes.c_double)]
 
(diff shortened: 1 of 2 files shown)


sunqm:
Do you think they should be printed every time the DFT method was called or eval_xc for a particular XC functional was called?

susilehtola:
That's a good question.

It should be sufficient to print out the version of libxc/xcfun at the beginning of every calculation. Information on the density functional also needs only be printed out once, in the corresponding initialization routine. This naturally would fit as part of the quadrature grid info printout that already displays the functional.

However, since PySCF can change the backend on the fly, it might be necessary to print out that information every time :| (?)

susilehtola:
Oh and this PR is missing the functions to query information about a functional; that information is available in the libxc `xc_func_type` structure in `func.info->refs[i]->ref` and `func.info->refs[i]->doi` where `0 <= i < XC_MAX_REFERENCES` is the index of the functional reference. The former is the bibliography entry in text, and the latter is the DOI.

sunqm:
I made a few modifications and put your commits in another PR (#922). Please put your changes and comments in that PR

PR #907, opened 2021-04-15
KMP2 with 3-index DF ints and timers

xwang862:


verena-neufeld:
I haven't had a chance to look at this yet again but saw that you added basis sets. I think it would be nice to give some citation and explanation where those come from or is that already explained somewhere else in the code? :)

xwang862:
@verena-neufeld  Do you mean commit 05bc4d9 "Add modified gth-tzvp"? It was for my own purpose and has been reverted in a later commit cf2faf1

sunqm (approved):

sunqm:
@xwang862 @verena-neufeld will you make changes to docs or examples? Otherwise I will merge this feature

verena-neufeld:
@xwang862 Have any of the APIs changed for a user?

pyscf/pbc/mp/kmp2.py
```diff
@@ -92,13 +107,91 @@ def kernel(mp, mo_energy, mo_coeff, verbose=logger.NOTE, with_t2=WITH_T2):
...
 
+def _init_mp_df_eris(mp):
+    """Add 3-center electron repulsion integrals, i.e. (L|pq), in `eris`,
+    where `L` denotes DF auxiliary basis functions and `p` and `q` canonical
+    crystalline orbitals. Note that `p` and `q` contain kpt indices `kp` and `kq`,
+    and the third kpt index `kL` is determined by the conservation of momentum.
+
+    Arguments:
+        mp {Kmp} -- A Kmp instance
+        eris {_mp_ERIS} -- A _mp_ERIS instance to which we want to add 3c ints
```
verena-neufeld, line 127: This is no longer an argument

pyscf/pbc/mp/kmp2.py
```diff
@@ -92,13 +107,91 @@ def kernel(mp, mo_energy, mo_coeff, verbose=logger.NOTE, with_t2=WITH_T2):
...
+
+    Returns:
+        _mp_ERIS -- A _mp_ERIS instance with 3c ints
```
verena-neufeld, line 130: is this equal to `Lov`?

pyscf/pbc/mp/kmp2.py
```diff
@@ -28,18 +28,23 @@
...
 from pyscf.mp import mp2
+from pyscf.pbc import df
 from pyscf.pbc.lib import kpts_helper
 from pyscf.lib.parameters import LARGE_DENOM
 from pyscf import __config__
 
 WITH_T2 = getattr(__config__, 'mp_mp2_with_t2', True)
 
-
 def kernel(mp, mo_energy, mo_coeff, verbose=logger.NOTE, with_t2=WITH_T2):
```
verena-neufeld, line 43: Might be nice to have a docstring here

xwang862:
> @xwang862 @verena-neufeld will you make changes to docs or examples? Otherwise I will merge this feature

Yes @sunqm , there might be some changes in docs. Thanks for @verena-neufeld 's comments in advance

xwang862:
> @xwang862 Have any of the APIs changed for a user?

Old APIs are not affected. But after this PR, `Lov` (the 3-index DF ints) will be computed and stored for GDF by default, which needs (Nk^2 * Naux * Nocc * Nvir * 16) disk space. Users can choose to turn it off by setting `mymp.with_df_ints = False`.

verena-neufeld:
Is that something that only applies to k MP2? So would you recommend adding a section in the MP2 user guide about this special feature in k MP2?

xwang862:
> Is that something that only applies to k MP2? So would you recommend adding a section in the MP2 user guide about this special feature in k MP2? I'd be happy to add a small section about `mymp.with_df_ints = False` if the memory space needed gets too large.

This change only applies to KMP2, but the feature is similar to molecular `DFMP2`. Are we covering both molecular and PBC MP2 in the MP2 user guide? If so, yes, adding a few lines about setting `mymp.with_df_ints` sounds good.

obackhouse:
@verena-neufeld I believe that this is also implemented in `pbc.ci.kcis_rhf` and also partially in `pbc.cc.kccsd` - they use `direct` as the keyword instead of `with_df_ints`.

xwang862:
@obackhouse good catch! Indeed they are the same feature but with different keywords. The use of `direct` makes more sense in KCIS and KCCSD - because we build 4c ERIs on the fly instead of storing them. In KMP2 we don't store full ERIs anyway, so the concept of being "direct" or not becomes vague here. So I decided to use a more straightforward keyword `with_df_ints`.

xwang862:
If there are no more comments, I think it's ready to be merged.

PR #942, opened 2021-05-13
Changes variable names in ADC kernel to prevent redeclaration of `str`

obackhouse:
The `str` function is reassigned to a string in the ADC kernel, leading to an exception when the `str` function is called if `nfalse >= 1` is met. This PR changes the variable to `header` and the same in the `analyze` function for good measure.


ssh2 (approved):
Thanks for spotting this! This looks good to me.

PR #943, opened 2021-05-15
Fix atomic initial guess if multiple elements have ecp

sunqm:


zhcui (approved):
I checked this patch and it resolves my previous issue.

fishjojo (approved):

PR #945, opened 2021-05-18
Accelerating sgx module

kylebystrom:
Using the new grids integrals added with #930, this PR significantly accelerates the sgx module with faster integral evaluation as well as the option of P-junction screening.

In particular, my implementation has two screening steps, S-junction and P-junction, whereas the CVHFOpt struct only has one screening function member. Currently, I just hard coded in the use of SGXnr_pj_screen, but it might be better to let that choice of screening function be changed. I would appreciate any thoughts on how to best implement this.

Update: I added a couple fixes for issues with the original revisions to SGX, namely:
- Added the direct_scf_sgx key to SGXHF to avoid "does not have attribute" warning.
- Take absolute value of grid weights in P-junction screening step in case they are negative.
- Added an example for P-junction screening with a larger molecule (dodecane).
- Moved the warning and correction issued when dfj=False and pjs=True from pre_kernel to build, in case the object is used without the SCF loop
- Added `rebuild_nsteps` parameter to reset the direct JK matrix construction every `rebuild_nsteps` SCF steps, which helps avoid numerical error.

I have attached a couple of figures summarizing speedups compared to the previous SGX version and analytical exchange.
image
image

One notable occurrence is that the speedup decreases going from octane (26 atoms) with q-zeta basis to the "cubane" (80 atoms) with t-zeta basis. I still need to investigate whether this is due to COSX being less efficient for the structure/basis, other routines taking up more runtime, or SGX P-junction screening being less efficient than for the analytical routine. If the latter, I would like to improve it if possible.


leebo-2020:
Could you show your input? thank you

leebo-2020 (approved):

kylebystrom:
This PR is ready for a review now. I have a couple questions that I would appreciate feedback on:
- Do you have any thoughts on the P-junction screening and general code structure? The P-junction screening follows the Neese et al. COSX implementation, which has the flaw of being dependent on the grid density (all integrals would get screened for an infinitely dense grid). I can implement a different one if that would be preferable. The current P-junction screening function is also currently hard-coded rather than being passed through the optimizer object. P-junction screening is off by default.
- When P-junction screening is turned on, dfj is set to True automatically in the pre_kernel call, since P-junction screening and COS Coulomb are incompatible. Is this good practice or should the code issue a user warning instead?

susilehtola:
The Ochsenfeld group has just pointed out that one doesn't even need double precision for seminumerical exchange!

http://doi.org/10.1063/5.0045084

leebo-2020 (approved):

pyscf/lib/vhf/nr_sgx_direct.c
```diff
@@ -35,138 +36,223 @@ typedef struct {
...
+        }
+        IntorEnvs envs = {natm, nbas, atm, bas, tmp_env, shls_slice, ao_loc, NULL,
+                          cintopt, ncomp};
         SGXJKArray *v_priv[n_dm];
         for (i = 0; i < n_dm; i++) {
-                v_priv[i] = jkop[i]->allocate(shls_slice, ao_loc, ncomp);
+                v_priv[i] = jkop[i]->allocate(shls_slice, ao_loc, ncomp, env[NGRIDS]);
         }
-        double *buf = malloc(sizeof(double) * di*di*ncomp);
+        double *buf = malloc(sizeof(double) * env[NGRIDS]*di*di*ncomp);
```
sunqm, line 239: do not use env[NGRIDS] here, it's wrong type

pyscf/lib/vhf/nr_sgx_direct.c
```diff
@@ -35,138 +36,223 @@ typedef struct {
...
+        double* tmp_env = (double*) malloc(env_size * sizeof(double));
+        for (i = 0; i < env_size; i++) {
+            tmp_env[i] = env[i];
+        }
+        IntorEnvs envs = {natm, nbas, atm, bas, tmp_env, shls_slice, ao_loc, NULL,
+                          cintopt, ncomp};
         SGXJKArray *v_priv[n_dm];
         for (i = 0; i < n_dm; i++) {
-                v_priv[i] = jkop[i]->allocate(shls_slice, ao_loc, ncomp);
+                v_priv[i] = jkop[i]->allocate(shls_slice, ao_loc, ncomp, env[NGRIDS]);
```
sunqm, line 237: do not use env[NGRIDS] here, it's wrong type

pyscf/lib/vhf/nr_sgx_direct.c
```diff
@@ -35,138 +36,223 @@ typedef struct {
...
+                        grids[3*tmp_ngrids+2] = all_grids[3*k+2];
+                        inds[tmp_ngrids] = k;
+                        tmp_ngrids++;
+                }
+                env[NGRIDS] = tmp_ngrids;
+        }
+
+        int grid0, grid1;
+        const int dims[] = {ao_loc[ish+1]-ao_loc[ish], ao_loc[jsh+1]-ao_loc[jsh], tmp_ngrids};
+        for (grid0 = 0; grid0 < env[NGRIDS]; grid0 += BLKSIZE) {
```
sunqm, line 170: env[NGRIDS] has wrong type. use tmp_ngrids instead of env[NGRIDS]

pyscf/scf/_vhf.py
```diff
@@ -147,8 +147,66 @@ def get_dm_cond(self, shape=None):
...
+                fsetdm = getattr(libcvhf, self._dmcondname)
+            if self._dmcondname == 'SGXsetnr_direct_scf_dm':
+                fsetdm(self._this,
+                       dm.ctypes.data_as(ctypes.c_void_p), ctypes.c_int(n_dm),
+                       ao_loc.ctypes.data_as(ctypes.c_void_p),
+                       c_atm.ctypes.data_as(ctypes.c_void_p), natm,
+                       c_bas.ctypes.data_as(ctypes.c_void_p), nbas,
+                       c_env.ctypes.data_as(ctypes.c_void_p),
+                       ctypes.c_int(ngrids),
+                       ctypes.c_int(0 if self.is_allocd else 1))
```
sunqm, line 184: is_allocd is not needed. we can test if dm_cond is allocated with (dm_cond == NULL)

pyscf/sgx/sgx_jk.py
```diff
@@ -246,31 +250,37 @@ def batch_nuc(mol, grid_coords, out=None):
...
-        atm, bas, env = gto.mole.conc_env(mol._atm, mol._bas, mol._env,
-                                          fakemol._atm, fakemol._bas, fakemol._env)
+    def jk_part(mol, grid_coords, dms, fg, weights):
+        if pjs:
+            sgxopt.set_dm(fg/numpy.sqrt(weights[None,:]), mol._atm, mol._bas, mol._env)
+            #sgxopt.set_dm(fg, mol._atm, mol._bas, mol._env)
+        atm, bas, env = mol._atm, mol._bas, mol._env
+        ngrids = grid_coords.shape[0]
+        env = numpy.append(env, grid_coords.ravel())
+        env[11] = ngrids
```
sunqm, line 277: env[gto.NGRIDS] = ngrids
env[gto.PTR_GRIDS] = mol._env.size

pyscf/sgx/sgx.py
```diff
@@ -109,26 +119,48 @@ def reset(self, mol=None):
...
+                self._last_vk = vk.copy()
+
+            else:
+                self._last_dm = numpy.asarray(dm)
+                vj, vk = with_df.get_jk(dm, hermi, self.opt, with_j, with_k,
+                                        self.direct_scf_tol, omega)
+
+            return vj, vk
 
         def post_kernel(self, envs):
```
sunqm, line 165: Also reset last_vj and last_vk in post_kernel

sunqm:
> The Ochsenfeld group has just pointed out that one doesn't even need double precision for all parts of seminumerical exchange. I wonder whether this would imply any performance benefits in PySCF
> 
> http://doi.org/10.1063/5.0045084

I believe it's a good point to test after merging this PR.

sunqm:
> This PR is ready for a review now. I have a couple questions that I would appreciate feedback on:
> 
> * Do you have any thoughts on the P-junction screening and general code structure? The P-junction screening follows the Neese et al. COSX implementation, which has the flaw of being dependent on the grid density (all integrals would get screened for an infinitely dense grid). I can implement a different one if that would be preferable. The current P-junction screening function is also currently hard-coded rather than being passed through the optimizer object. P-junction screening is off by default.
> * When P-junction screening is turned on, dfj is set to True automatically in the pre_kernel call, since P-junction screening and COS Coulomb are incompatible. Is this good practice or should the code issue a user warning instead?

I think the current structure is enough to support the basic non-relativistic J/K matrices. When considering more features, function signature and code structure may need to be restructured or optimized. IMO, we don't need to worry about it too early.

I think it's not bad to print a warning message about the P-junction issue in this version. We can remove the warning messages after we find better designs or architectures to smartly handle all these different screening conditions.

pyscf/lib/vhf/nr_sgx_direct.c
kylebystrom: Fixed by 4ef2ea7

pyscf/lib/vhf/nr_sgx_direct.c
kylebystrom: Fixed by 4ef2ea7

pyscf/lib/vhf/nr_sgx_direct.c
```diff
@@ -241,6 +328,41 @@ void SGXsetnr_direct_scf(CVHFOpt *opt, int (*intor)(), CINTOpt *cintopt,
         free(cache);
 }
 }
+
+void SGXsetnr_direct_scf_dm(CVHFOpt *opt, double *dm, int nset, int *ao_loc,
+                            int *atm, int natm, int *bas, int nbas, double *env,
+                            int ngrids, int alloc)
+{
+        nbas = opt->nbas;
+        if (1) {
```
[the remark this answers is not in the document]
kylebystrom: Fixed by 4ef2ea7

pyscf/lib/vhf/optimizer.c
```diff
@@ -25,13 +25,17 @@
...
+
 int int2e_sph();
 
 void CVHFinit_optimizer(CVHFOpt **opt, int *atm, int natm,
                         int *bas, int nbas, double *env)
 {
         CVHFOpt *opt0 = (CVHFOpt *)malloc(sizeof(CVHFOpt));
         opt0->nbas = nbas;
+        opt0->ngrids = 0;
+        opt0->pscreen = 0;
```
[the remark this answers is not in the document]
kylebystrom: Fixed by 4ef2ea7

pyscf/lib/vhf/nr_sgx_direct.c
kylebystrom: Note, I used ngrids here, not tmp_ngrids, which is only defined in a different scope.

pyscf/scf/_vhf.py
kylebystrom: Fixed by 4ef2ea7

pyscf/sgx/sgx_jk.py
kylebystrom: Fixed by 4ef2ea7

pyscf/sgx/sgx.py
kylebystrom: Fixed by 4ef2ea7

kylebystrom:
Thank you for the feedback @sunqm! I made the requested changes, and I am currently running a few linear octane calculations to double-check the P-junction screening one more time. Should I add a unit test with a really long molecule with a very small basis set to check P-junction screening in the test suite? That way it will check that the long-range screening works without excessive computational cost.

kylebystrom:
In addition to fixing the issues above, I did a couple more cleanup things:
- Added the direct_scf_sgx key to SGXHF to avoid "does not have attribute" warning.
- Take absolute value of grid weights in P-junction screening step in case they are negative.
- Added an example for P-junction screening with a larger molecule (dodecane).
- Moved the warning and correction issued when dfj=False and pjs=True from pre_kernel to build, in case the object is used without the SCF loop
I also signed myself up for the docs for this module. Should I complete those now or make a separate PR after this is merged?

kylebystrom:
It turns out that gcc 4 and 9 have incompatible approaches to shared const variables for openmp. gcc 4 does not allow specifying const variables in shared even when default(none) is used, while in gcc 9 it is necessary to specify const variables as shared when default(none) is used. I fixed this by changing ish0 and jsh0 to firstprivate; since it is only two int variables, both of which are const, this should not cause any overhead or make the code susceptible to bugs. See https://gcc.gnu.org/gcc-9/porting_to.html for details.

sunqm:
> Thank you for the feedback @sunqm! I made the requested changes, and I am currently running a few linear octane calculations to double-check the P-junction screening one more time. Should I add a unit test with a really long molecule with a very small basis set to check P-junction screening in the test suite? That way it will check that the long-range screening works without excessive computational cost.

Yes, please add unit test. The test pipeline will skip these tests if the test function has the decorator `@unittest.skip('.....')` or suffix "_skip" in the name of the test function. Should be sufficient to distinguish to the rest. In the future, they may be configured to run at different stages of development

sunqm:
> In addition to fixing the issues above, I did a couple more cleanup things:
> 
> * Added the direct_scf_sgx key to SGXHF to avoid "does not have attribute" warning.
> * Take absolute value of grid weights in P-junction screening step in case they are negative.
> * Added an example for P-junction screening with a larger molecule (dodecane).
> * Moved the warning and correction issued when dfj=False and pjs=True from pre_kernel to build, in case the object is used without the SCF loop
> 
> I also signed myself up for the docs for this module. Should I complete those now or make a separate PR after this is merged?

Please update your changes in this PR, especially the negative weights problem.

The documents for sgx was quite outdated.  Please update the document file user/sgx.rst of https://github.com/pyscf/pyscf.github.io. Next document-hackathon will focus on the program reference, and the docstring of the code. If that's not available yet, we can merge this PR and leave the program reference to next project.

sunqm (approved):

kylebystrom:
I updated the PR description with the updates listed above, added the P-junction screening unit test with a skip decorator and note about the computational cost, and opened PR 82 at https://github.com/pyscf/pyscf.github.io updating the user docs.

kylebystrom:
Based on the discussion on docs PR 82 (https://github.com/pyscf/pyscf.github.io/pull/82), I added a `rebuild_nsteps` feature for use with direct SGX construction to avoid creeping numerical error. I will add this to the PR description as well.

sunqm:
Can you merge the master branch to your local branch? The fail in github action should be fixed in latest updates.

kylebystrom:
I merged with master and pushed, but the CI workflow failed in the same place again, seemingly on routines unrelated to the SGX changes (https://github.com/kylebystrom/pyscf/runs/3454940355?check_suite_focus=true). The two failures were small numerical errors:
```
======================================================================
FAIL: test_with_df (test_dfccsd.KnownValues)
----------------------------------------------------------------------
  File "/home/runner/work/pyscf/pyscf/pyscf/cc/test/test_dfccsd.py", line 81, in test_with_df
    self.assertAlmostEqual(lib.finger(numpy.array(eris.oooo)), 4.962033460861587 , 12)
AssertionError: 4.962033460862377 != 4.962033460861587 within 12 places (7.895906151134113e-13 difference)

======================================================================
FAIL: test_update_amps1 (test_uccsd.KnownValues)
----------------------------------------------------------------------
  File "/home/runner/work/pyscf/pyscf/pyscf/cc/test/test_uccsd.py", line 415, in test_update_amps1
    self.assertAlmostEqual(lib.finger(t2[1]), -9675.767769478574, 7)
AssertionError: -9675.767769601576 != -9675.767769478574 within 7 places (1.2300188245717436e-07 difference)
```
Actually, it is a bit strange that these assertions fail in the first place; I thought assertAlmostEqual passed if the error was <1.5*10^-d, with d being the decimal place.

sunqm:
What about the current head of master branch? The relevant thresholds were loosened a little bit.

kylebystrom:
That worked! https://github.com/kylebystrom/pyscf/actions/runs/1180560641

susilehtola:
FWIW here's a new paper by Neese et al
https://aip.scitation.org/doi/full/10.1063/5.0058766

pyscf/scf/_vhf.py
```diff
@@ -147,8 +147,59 @@ def get_dm_cond(self, shape=None):
...
+        data = ctypes.cast(self._this.contents.dm_cond,
+                           ctypes.POINTER(ctypes.c_double))
+        return numpy.ctypeslib.as_array(data, shape=shape)
+    dm_cond = property(get_dm_cond)
+
+
 class _CVHFOpt(ctypes.Structure):
     _fields_ = [('nbas', ctypes.c_int),
+                ('ngrids', ctypes.c_int),
                 ('_padding', ctypes.c_int),
```
sunqm, line 203: The placeholder `_padding` can be removed now.

pyscf/lib/vhf/optimizer.h
```diff
@@ -20,6 +20,7 @@
 #define HAVE_DEFINED_CVHFOPT_H
 typedef struct CVHFOpt_struct {
     int nbas;
+    int ngrids;
     int _padding;
```
sunqm, line 24: `_padding` the placeholder can be removed

sunqm (approved):

PR #963, opened 2021-05-31
"Native" DF-MP2 for RHF and UHF with densities

pinskip:
Dear PySCF developers,

we would like to contribute a new DF-MP2 code. Unlike the existing implementation of DF-MP2 (which only seems to support RHF references), the new code fully makes use of the density fitting approximation, and therefore does not require as much memory for calculations on larger molecules.

Its current features are:
- energies
- orbital-unrelaxed and orbital-relaxed 1-RDM
- full support of RHF and UHF references
- frozen core
- spin-component scaling

It wouldn't be a too big step to add the additional contributions for the analytical gradient; unfortunately I cannot do it myself at the moment, as I need to prioritize other projects. I won't promise, but we might follow up on the gradient in the future.

An open question from my side would be how to make it accessible easily. Currently the classes need to be imported from `pyscf.mp.dfmp2_native` or `pyscf.mp.dfump2_native`. I didn't add any imports to `__init__.py` in the `mp` module so far. It's not derived from the conventional `MP2` class, as its workings are rather different. For example, the full sets of amplitudes or four-index integrals (which the `MP2` class has got methods for) are never calculated and stored together. Of course, there is also a number of classes that derive from the already existing `DFMP2` class, and `MP2.density_fit()` returns an instance of the already existing `DFMP2`, not the new one.

There is no documentation yet. I intend to add it once I have received some feedback whether this code will be added to PySCF, and how I could best make it accessible without breaking existing features.

Kind regards,
Peter Pinski, HQS Quantum Simulations


sunqm (commented):
Looks good to me

sunqm:
I think mp2 module needs this feature. After merging this feature, I think the existing MP2 class can be refactored to enable the SCS treatment and make_rdm1 as this DFMP2 implementation

pinskip:
I wasn't sure if we are waiting for @fishjojo to review the code, or if everyone else is waiting for me to add documentation for completeness.
Anyway, I have added some documentation in https://github.com/pyscf/pyscf.github.io/pull/79, which describes the code as it is in this pull request right now.

fishjojo (approved):

PR #989, opened 2021-06-20
Some linting changes and add some docstrings to cisd.py

verena-neufeld:
I don't remember what linter I used, possibly pylint. Some changes are personal taste, so I can take them back if desired. 
It would be great if someone checks my description of the functions - I might have misunderstood some aspects.

Note that some parts of the docstring were copied from other docstrings in the pyscf code.

### DIFF
--- a/pyscf/ci/cisd.py
+++ b/pyscf/ci/cisd.py
@@ -99,7 +99,7 @@ def kernel(myci, eris, ci0=None, max_cycle=50, tol=1e-8, verbose=logger.INFO):
 
 def make_diagonal(myci, eris):
     '''
-    Return diagonal of hamiltonian in Slater determinant basis.
+    Return diagonal of CISD hamiltonian in Slater determinant basis.
 
     Note that a constant has been substracted of all elements.
     The first element is the HF energy (minus the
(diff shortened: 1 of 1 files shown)


pyscf/ci/cisd.py
```diff
@@ -730,7 +795,7 @@ def __call__(self, mol_or_geom, ci0=None, **kwargs):
                 # FIXME: Whether to use the initial guess from last step?
                 # If root flips, large errors may be found in the solutions
                 ci0 = self.ci
-            self.kernel(ci0, **kwargs)[0]
+            _ = self.kernel(ci0, **kwargs)[0]
```
verena-neufeld, line 798: I think, I'll reverse this change for example to be in line with other parts of the pyscf code.

pyscf/ci/cisd.py
```diff
@@ -913,9 +979,29 @@ def _finalize(self):
...
+
+        Returns:
+            Tuple of float and numpy array or
+            tuple of float and list of numpy arrays (if nroots > 1)
+            MP2 energy and initial guess(es) for CISD coefficients.
+
+        '''
         if eris is None: eris = self.ao2mo(self.mo_coeff)
         nocc = self.nocc
+        # [TODO] in `make_diagonal` function it says:
```
sunqm, line 1002: In ccsd code, eris.mo_energy was used to compute initial MP2 amplitude. IMO, it should be okey to generate MP2 amplitudes for initial guess. This is consistent with the treatment in MP2 code.

In cisd.make_diagonal function, I think the accurate way to compute Slater determinant would be the Fock matrix constructed from the reference wavefunction. eris.mo_energy may not be appropriate to compute the value of Slater determinant if it was the mo_energy of underlying scf object. Seems in many modules it was set to fock.diagonal already.

sunqm (approved):

PR #990, opened 2021-06-21
Add X2C for GHF which includes the SOC contributions

sunqm:
* Change the X2C-GHF interface. ghf.x2c() used to generate a spin-free-X2C-GHF method. With this patch, calling ghf.x2c() will generate the standard X2C in regular spin-orbital basis representation (in the GHF framework) while ghf.sfx2c1e() will generate the spin-free-X2C-GHF as before. The new X2C-GHF object theoretically can give the same result as the (j-adapted) spinor X2C method. In practice it often breaks the spin symmetry and converges to a lower energy solution.

* Add a helper method add_keys in StreamObject class. Using it to simplify the initialization of meta-class (e.g. DF, X2C)

### DIFF
--- a/pyscf/lib/numpy_helper.py
+++ b/pyscf/lib/numpy_helper.py
@@ -1028,10 +1028,14 @@ def condense(opname, a, loc_x, loc_y=None):
             for j,j0 in enumerate(loc_y):
                 j1 = loc_y[j+1]
                 out[i,j] = op(a[i0:i1,j0:j1])
+
+    opname can be  sum, max, min, abssum, absmax, absmin, norm
     '''
-    assert(a.dtype == numpy.double)
+    assert a.dtype == numpy.double
     if not opname.startswith('NP_'):
         opname = 'NP_' + opname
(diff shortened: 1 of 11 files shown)


pyscf/lib/numpy_helper.py
```diff
@@ -1041,6 +1045,8 @@ def condense(opname, a, loc_x, loc_y=None):
     nloc_y = loc_y.size - 1
     if a.flags.f_contiguous:
         out = numpy.zeros((nloc_x, nloc_y), order='F')
+        loc_x, loc_y = loc_y, loc_x
+        nloc_x, nloc_y = nloc_x, nloc_y
```
fishjojo, line 1049: is this correct?

fishjojo (approved):

pyscf/lib/numpy_helper.py
sunqm: It's correct but seems misleading. I changed implementations in new commits

PR #991, opened 2021-06-21
Sanity checks for mcscf methods

sunqm:
* More checks for mcscf methods
* More mcscf examples

### DIFF
--- a/examples/mcscf/02-cas_space_spin.py
+++ b/examples/mcscf/02-cas_space_spin.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+
+'''
+spin settings in active space
+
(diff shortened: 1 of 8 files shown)


bogdanoff:
We widely use the following snippet to define the tuple `nelecas` using `mol.spin` and integer `nelecas`:
```
(nel, norb) = (6, 6)
spin = mol.spin
# spin = mol.spin + 2
tuple_nel = (nel%2 + nel//2 + spin//2, nel//2 - spin//2)
mc = mf.CASSCF(norb, tuple_nel)
```
It may be useful to define the default  behavior when only integer `nelecas` is set, or to define a `mc.spin` setter.

fishjojo (approved):

Issue #1006, opened 2021-07-06
Mole.search_shell_id freeze


m-i-w:
Currently the `search_shell_id` method on Mole looks like this:

> def search_shell_id(self, atm_id, l):
>        return search_shell_id(self, atm_id, l)

Where as it looks like it should point to the function with the same name in the same file.

>     search_shell_id = search_shell_id

This is how all the other features are implemented for Mole. 

I'll open a PR with this fix and attach it .

m-i-w:
rereading this now it doesn't look like it's recursive... 

Whatever problem my machine was having seems to have stopped now so this is a non-issue.

PR #1007, opened 2021-07-06
fix search_shell_id bug

m-i-w:
`search_shell_id` used to be recursive, I've changed it to call on the function of the same name outside of the class.


m-i-w:
Fixes #1006.

PR #1008, opened 2021-07-07
Use a linearly independent basis in DIIS

### LINKED ISSUE #1000: DIIS does not use linearly independent basis
The DIIS code in PySCF appears to use AO matrices. This is not recommended, see Pulay's work from the early 1980s, as the algorithm does not work in the presence of linear dependencies. 

A better alternative is to project the matrices into the orbital basis, as this eliminates the linear dependencies.

The following calculation (needs the changes from pull #999 to run)
```
from pyscf import gto, scf

mol = gto.Mole()
mol.verbose = 4
mol.atom = '''                                                                                                                                                                                
Ne 0 0 0                                                                                                                                                                                      
Ne 0 0 1e-3'''
mol.basis = 'unc-aug-cc-pvtz'
mol.symmetry = 0
mol.build()

mf = scf.RHF(mol)
mf = scf.addons.remove_linear_dep_(mf, threshold=1e-5, cholesky_threshold=1e-6, force_pivoted_cholesky=True)
mf.init_guess = '1e'
# This fails                                                                                                                                                                                  
mf.kernel()

# DIIS in PySCF does not work properly; turn it off                                                                                                                                           
mf.diis=None
# This converges in 13 iterations                                                                                                                                                             
mf.kernel()
```
is an extreme demonstration of the issue: the basis set orthogonalization eliminates 47 basis functions out of a total of 116, leaving only 69 basis functions; this is a calculation from my recent paper [Phys. Rev. A 101, 032504 (2020)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.101.032504).

When DIIS is enabled, the calculation diverges. Turning off DIIS the calculation converges in 13 iterations when started from the core guess; when started from the default guess, the calculation fails to converge.

In contrast, the same calculation in ERKALE started from the SAD guess and DIIS yields convergence in just 5 iterations.

susilehtola:
Closes #1000 which may often cause convergence problems for molecules with linearly dependent basis sets.

### DIFF
--- a/pyscf/cc/test/test_uccsd.py
+++ b/pyscf/cc/test/test_uccsd.py
@@ -573,13 +573,13 @@ class KnownValues(unittest.TestCase):
     def test_mbpt2(self):
         myucc = uccsd.UCCSD(mf)
         e = myucc.kernel(mbpt2=True)[0]
-        self.assertAlmostEqual(e, -0.12886859466216125, 10)
+        self.assertAlmostEqual(e, -0.12886859466216125, 8)
         emp2 = mp.MP2(mf).kernel()[0]
         self.assertAlmostEqual(e, emp2, 10)
 
         myucc = uccsd.UCCSD(mf_s2)
         e = myucc.kernel(mbpt2=True)[0]
-        self.assertAlmostEqual(e, -0.096257842171487293, 10)
+        self.assertAlmostEqual(e, -0.096257842171487293, 8)
         emp2 = mp.MP2(mf_s2).kernel()[0]
         self.assertAlmostEqual(e, emp2, 10)
 
--- a/pyscf/dft/test/test_gks.py
+++ b/pyscf/dft/test/test_gks.py
@@ -71,7 +71,7 @@ class KnownValues(unittest.TestCase):
         mf.xc = 'lda,vwn'
         mf.collinear = 'ncol'
         eks4 = mf.kernel()
-        self.assertAlmostEqual(lib.fp(mf.mo_energy), -26.54785447210512, 8)
+        self.assertAlmostEqual(lib.fp(mf.mo_energy), -26.54785447210512, 5)
         self.assertAlmostEqual(eks4, -74.73210527989738, 8)
 
         mf = mol1.GKS()
@@ -104,7 +104,7 @@ class KnownValues(unittest.TestCase):
         mf.collinear = 'mcol'
         mf._numint.spin_samples = 6
         eks4 = mf.kernel()
-        self.assertAlmostEqual(lib.fp(mf.mo_energy), -26.54785447210512, 8)
+        self.assertAlmostEqual(lib.fp(mf.mo_energy), -26.54785447210512, 5)
         self.assertAlmostEqual(eks4, -74.73210527989738, 8)
 
         mf = mol1.GKS()
--- a/pyscf/dft/test/test_numint.py
+++ b/pyscf/dft/test/test_numint.py
@@ -549,14 +549,14 @@ class KnownValues(unittest.TestCase):
         with lib.temporary_env(numint, _sparse_enough=_not_sparse):
             rho, vxc, fxc = mf._numint.cache_xc_kernel(mf.mol, mf.grids, mf.xc, mf.mo_coeff, mf.mo_occ)
         self.assertAlmostEqual(rho[0].dot(mf.grids.weights), 10, 4)
-        self.assertAlmostEqual(numpy.einsum('g,g,ig->', mf.grids.weights, rho[0], rho), 81.04275692925363, 8)
-        self.assertAlmostEqual(numpy.einsum('g,xg,xyg->', mf.grids.weights, rho, fxc), -6.194969637088992, 8)
+        self.assertAlmostEqual(numpy.einsum('g,g,ig->', mf.grids.weights, rho[0], rho), 81.04275692925363, 5)
+        self.assertAlmostEqual(numpy.einsum('g,xg,xyg->', mf.grids.weights, rho, fxc), -6.194969637088992, 5)
 
         with lib.temporary_env(numint, _sparse_enough=_sparse):
             rho, vxc, fxc = mf._numint.cache_xc_kernel(mf.mol, mf.grids, mf.xc, mf.mo_coeff, mf.mo_occ)
         self.assertAlmostEqual(rho[0].dot(mf.grids.weights), 10, 4)
-        self.assertAlmostEqual(numpy.einsum('g,g,ig->', mf.grids.weights, rho[0], rho), 81.04275692925363, 8)
-        self.assertAlmostEqual(numpy.einsum('g,xg,xyg->', mf.grids.weights, rho, fxc), -6.194969637088992, 8)
+        self.assertAlmostEqual(numpy.einsum('g,g,ig->', mf.grids.weights, rho[0], rho), 81.04275692925363, 5)
+        self.assertAlmostEqual(numpy.einsum('g,xg,xyg->', mf.grids.weights, rho, fxc), -6.194969637088992, 5)
 
         if hasattr(dft, 'xcfun'):
             mf.xc = 'camb3lyp'
--- a/pyscf/fci/test/test_spin0.py
+++ b/pyscf/fci/test/test_spin0.py
@@ -143,7 +143,7 @@ class KnownValues(unittest.TestCase):
             symmetry = True,
             basis = '6-311g')
         mf = scf.RHF(mol)
-        mf.scf()
+        mf.run(conv_tol=1e-10)
         mf._scf = mf
         h1e = mcscf.casci.h1e_for_cas(mf, mf.mo_coeff, ncas=2, ncore=2)[0]
         eri = ao2mo.incore.full(mf._eri, mf.mo_coeff[:,2:4])
@@ -152,7 +152,7 @@ class KnownValues(unittest.TestCase):
         ci0 = numpy.zeros((2,2))
         ci0[0,0] = 1
         e, c = cis.kernel(h1e, eri, 2, 2, ci0)
-        self.assertAlmostEqual(e, -0.80755526695538049, 10)
+        self.assertAlmostEqual(e, -0.80755526695538049, 7)
 
         cis = fci.direct_spin0_symm.FCISolver(mol)
         # Test the default initial guess. It should give "0" in the results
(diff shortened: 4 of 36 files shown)


susilehtola:
It appears some further changes are necessary to make the fix work for unrestricted calculations; maybe symmetry too.

susilehtola:
With the current changes, calculations at RHF, UHF, and ROHF level of theory work without problem.

```
from pyscf import gto, scf

mol_singlet = gto.Mole()
mol_singlet.verbose = 4
mol_singlet.atom = '''
Ne 0 0 5e-4
ghost 0 0 0
Ne 0 0 -5e-4'''
mol_singlet.basis = {'Ne' : 'unc-cc-pvtz', 'ghost' : gto.uncontract(gto.basis.load('cc-pvtz','Ca')) }
mol_singlet.spin = 0
mol_singlet.symmetry = 0
mol_singlet.build()

rhf = scf.RHF(mol_singlet)
rhf = scf.addons.remove_linear_dep_(rhf, threshold=1e-5, cholesky_threshold=1e-6)
rhf.init_guess = '1e'
rhf.kernel()

mol_triplet = gto.Mole()
mol_triplet.verbose = 4
mol_triplet.atom = '''
Ne 0 0 5e-4
ghost 0 0 0
Ne 0 0 -5e-4'''
mol_triplet.basis = {'Ne' : 'unc-cc-pvtz', 'ghost' : gto.uncontract(gto.basis.load('cc-pvtz','Ca')) }
mol_triplet.spin = 2
mol_triplet.symmetry = 0
mol_triplet.build()

uhf = scf.UHF(mol_triplet)
uhf = scf.addons.remove_linear_dep_(uhf, threshold=1e-5, cholesky_threshold=1e-6)
uhf.init_guess = '1e'
uhf.kernel()

rohf = scf.ROHF(mol_triplet)
rohf = scf.addons.remove_linear_dep_(rohf, threshold=1e-5, cholesky_threshold=1e-6)
rohf.init_guess = '1e'
rohf.kernel()
```

sunqm:
Except in certain extreme systems, I didn't see too many adavantage to use custom basis than the default AO basis in DIIS.  Numerically, extra transformations would potentially bring more round off errors.

Does it make sense to create a child class of CDIIS that allows to run DIIS on custom basis? It can be assigned to scf object when calling the remove_linear_dep_ function around https://github.com/pyscf/pyscf/blob/master/pyscf/scf/addons.py#L484 .

susilehtola:
> Except in certain extreme systems, I didn't see too many adavantage to use custom basis than the default AO basis in DIIS. Numerically, extra transformations would potentially bring more round off errors.

Is that really a concern? The change only affects the computation of DIIS error, not the extrapolation itself, and the benefits of using the orthogonal basis were noted already 40 years ago by Pulay. There are plenty of systems that don't converge without the elimination of the linearly dependent functions; this is why the removal is the default in all programs (and the rest just fail to converge!).

> Does it make sense to create a child class of CDIIS that allows to run DIIS on custom basis? It can be assigned to scf object when calling the remove_linear_dep_ function around https://github.com/pyscf/pyscf/blob/master/pyscf/scf/addons.py#L484 .

Or one could complicate the code so that the transform is only executed if `Corth is not None` but per the point above I don't see why. Making the customization in `remove_linear_dep_` would also prevent the code from working when `mf.eig` is customized in another manner, as e.g. in the pyscf examples that many users are still probably using to eliminate linear dependencies in their calculations... the current implementation is fool proof in that sense.

sunqm:
There can be a DIIS class with sophisticated treatment. Other choices of DIIS or convergence technique, like EDIIS, ADIIS, the linear dependence function are needed but they should work as backup plan. The simplest CDIIS should be there and working as the default choice IMO.

sunqm:
Another two things to be updated in the orthogonal basis DIIS:
1. the complex conjugations of Corth in the general sense
2. in the case of "f.ndim == 3", f contains several fock matrices. They are different fock matrices in different subspace, e.g. different k-points, different symmetry blocks (although not utilized in current version). Corth basis may need to be different for each fock matrices

verena-neufeld:
Hi Susi,

Would these changes help:

> diff --git a/pyscf/scf/diis.py b/pyscf/scf/diis.py
> index 9c2527c3f..2a555fbe6 100644
> --- a/pyscf/scf/diis.py
> +++ b/pyscf/scf/diis.py
> @@ -64,19 +64,19 @@ def get_err_vec(s, d, f, Corth):
>      '''error vector = SDF - FDS'''
>      if isinstance(f, numpy.ndarray) and f.ndim == 2:
>          sdf = reduce(numpy.dot, (s,d,f))
> -        errvec = numpy.dot(Corth.T, numpy.dot(sdf.T.conj() - sdf, Corth))
> +        errvec = numpy.dot(Corth.T.conj(), numpy.dot(sdf.T.conj() - sdf, Corth))
> 
>      elif isinstance(f, numpy.ndarray) and f.ndim == 3 and s.ndim == 3:
>          errvec = []
>          for i in range(f.shape[0]):
>              sdf = reduce(numpy.dot, (s[i], d[i], f[i]))
> -            errvec.append(numpy.dot(Corth.T, numpy.dot((sdf.T.conj() - sdf), Corth)))
> +            errvec.append(numpy.dot(Corth[i].T.conj(), numpy.dot((sdf.T.conj() - sdf), Corth[i])))
>          errvec = numpy.vstack(errvec)
> 
>      elif f.ndim == s.ndim+1 and f.shape[0] == 2:  # for UHF
>          nao = s.shape[-1]
>          s = lib.asarray((s,s)).reshape(-1,nao,nao)
> -        return get_err_vec(s, d.reshape(s.shape), f.reshape(s.shape), Corth)
> +        return get_err_vec(s, d.reshape(s.shape), f.reshape(s.shape), Corth.reshape(s.shape))
>      else:
>          raise RuntimeError('Unknown SCF DIIS type')
>      return errvec
> diff --git a/pyscf/scf/hf.py b/pyscf/scf/hf.py
> index d4191043e..2493693d8 100644
> --- a/pyscf/scf/hf.py
> +++ b/pyscf/scf/hf.py
> @@ -152,16 +152,12 @@ Keyword argument "init_dm" is replaced by "dm0"''')
>          mf_diis = mf.DIIS(mf, mf.diis_file)
>          mf_diis.space = mf.diis_space
>          mf_diis.rollback = mf.diis_space_rollback
>          # We get the used orthonormalized AO basis from any old eigendecomposition.
>          # Since the ingredients for the Fock matrix has already been built, we can
>          # just go ahead and use it to determine the orthonormal basis vectors.
>          fock = mf.get_fock(h1e, s1e, vhf, dm)
> -        mo_energy, mo_coeff = mf.eig(fock, s1e)
> -        if isinstance(mo_coeff, numpy.ndarray) and mo_coeff.ndim == 2:
> -            mf_diis.Corth = mo_coeff
> -        else:
> -            mf_diis.Corth = mo_coeff[0]
> +        _, mf_diis.Corth = mf.eig(fock, s1e)
>      else:
>          mf_diis = None

susilehtola:
@verena-neufeld please file a PR.

verena-neufeld:
Ok, sure. I have noticed that in kUHF, mo_coeff is a tuple, so Corth needs to be converted to np.array there before reshaping. Also, Corth is only evaluated once. Is this to save time or is there another reason? I think it should also be stored in the DIIS file for restarting to ensure consistency.

susilehtola:
> Ok, sure. I have noticed that in kUHF, mo_coeff is a tuple, so Corth needs to be converted to np.array there before reshaping. Also, Corth is only evaluated once. Is this to save time or is there another reason? 

The B matrix in DIIS is built from dot products of error vectors. `Corth` is used to calculate the error vector in the linearly independent basis; typically one uses the `X` matrix i.e. `S^-1/2` but, as this is not used in PySCF, taking in the first set of MO coefficients works just as well.

For the crystalline case, one needs to have a **k**-point dependent `Corth` as well, since the number of linearly independent vectors depends on **k**.

>I think it should also be stored in the DIIS file for restarting to ensure consistency (restoring test is failing).

Good point.

verena-neufeld:
Thanks! Is using the first set of mo coeffs more suitable to represent the X or is it a convenience thing that Corth is evaluated once and then kept?

susilehtola:
> Thanks! Is using the first set of mo coeffs more suitable to represent the X or is it a convenience thing that Corth is evaluated once and then kept?

Like I wrote above, since `Corth` is used to build the error vector in the linearly independent basis, it needs to be the same from one iteration to the next. This is why I set it to the first set of MO coeffs.

btglau:
Would this DIIS improve the convergence of periodic calculations (SCF and CCSD)? The condition number is typically over 1e10, and I'm doing calculations on pretty large systems (256 atoms) where each cycle of SCF takes 2-3 hours.

susilehtola:
> Would this DIIS improve the convergence of periodic calculations (SCF and CCSD)? The condition number is typically over 1e10, and I'm doing calculations on pretty large systems (256 atoms) where each cycle of SCF takes 2-3 hours.

It's possible. If you throw away linearly dependent eigenvectors and you don't do the same in DIIS, you may need unnecessarily many iterations. Many quantum chemistry codes have had issues in implementing DIIS properly for this reason...

susilehtola:
So, in the pyscf/pbc/cc/test/test_eom_kuccsd_diag.py test, for instance, the F, D, and S matrices are all (6, 4, 4), but Corth (as the second return argument of `eig`) is a [2 x 3] x (4, 4) double list... Does this make sense?

susilehtola:
Something has happened in PySCF though since the example above doesn't work anymore...

verena-neufeld:
Hi Susi, thanks for making updates here. One reason why I hadn't opened a PR to your repo was that I realised that when which orbital is occ and which orbital is virt varies from k point to k point, and sometimes this changes from iteration to iteration of the SCF cycle, this can cause issues if I remember correctly. Not sure whether this DIIS works with padding in pbc calcs where orbitals get frozen. Maybe you solved this issue?

verena-neufeld:
Could it be that when occupancies varies from one k point to the next and this changes from one iteration to the next, some parts of Corth belonging to k point "a" will be used for k point "b" (if "b" has more occ orbitals than initially thought for example)? I haven't tested this for a while.

susilehtola:
The same thing also happens in molecular calculations; the occupied orbitals can change symmetry. This should not be a problem.

sunqm (approved):

PR #1029, opened 2021-08-03
Add most recent GTH-HF potential

hongzhouye:
The current GTH-HF pseudopotential (PP) file `pbc/gto/pseudo/gth-hf.dat` is outdated and contains PP data for only 4 elements. Recently, Jürg Hutter refitted the GTH-HF PP for all elements up to Rn. These new PPs can be found [here](https://github.com/juerghutter/GTH/blob/master/HF/POTENTIAL). This commit adds these new GTH-HF PPs to PySCF.

@tberkel suggested that we could either
1. rename the old one to be `gth-hf-old` and make the new one as `gth-hf` or
2. keep the old one and name the new one as `gth2-hf`.

I took option 2 in this commit so that one can use it via
```
cell.pseudo = "gth2-hf"
```
but feel free to take whichever is more appropriate/causes less problems.


tberkel:
Just to complicate matters with a third option, we could also call it `rev-gth-hf`?

sunqm (approved):

sunqm:
Do you still consider to rename the potential file?

hongzhouye:
> Do you still consider to rename the potential file?

I don't have a preference. If no one else has a strong preference either we may just keep the original `GTH-HF` unchanged and go with the `GTH2-HF` name for the new one.

fishjojo:
I feel rev-gth-hf (or gth-hf-rev) is better. gth2 sounds like a new family of potentials.

tberkel:
For whatever it’s worth, I thought of gth2 because of the def2 family of
basis sets. But I am also indifferent as to how it’s named. I think
gth-hf-rev is a good idea.

sunqm:
Seems no more comments for the name of "gth-hf-rev". I'll change gth2-hf to this name

PR #1031, opened 2021-08-05
Add a memory check to KMP2

xwang862:
KMP2 calculations may be killed by an OOM error without traceback if the max memory is insufficient (especially when users don't realize that `with_t2` is true by default). In this PR a MemoryError will be raised if estimated memory usage exceeds currently available memory.

Thanks @hongzhouye for bringing it up!


verena-neufeld:
I think this is a great idea! Would it make sense to suggest that the user might set `with_t2` to `False` if not needed? When not using DF, is there another cost instead ? - Sorry for asking with DF if that is obvious, I'll read this more carefully later.

sunqm (approved):

Issue #873, opened 2021-03-29
Could stability check return a bool variable for stable/unstable?


hebrewsnabla:
For example, `uhf_stability` or `uhf_internal` only return a set of new MO (unchanged when wavefunction is stable). It would be much more convenient, when developing, if these functions also return a bool variable indicating the result of the stability check.

Only a few lines of codes need to be added at the end of each stability-check function. Here is an example
```
def uhf_internal(mf, with_symmetry=True, verbose=None):
    ...
    if e < -1e-5:
        log.note('UHF/UKS wavefunction has an internal instability.')
        nocca = numpy.count_nonzero(mf.mo_occ[0]> 0)
        nvira = numpy.count_nonzero(mf.mo_occ[0]==0)
        mo = (_rotate_mo(mf.mo_coeff[0], mf.mo_occ[0], v[:nocca*nvira]),
              _rotate_mo(mf.mo_coeff[1], mf.mo_occ[1], v[nocca*nvira:]))
        stable = False
    else:
        log.note('UHF/UKS wavefunction is stable in the internal stability analysis')
        mo = mf.mo_coeff
        stable = True
    return mo, stable
```

jjgoings:
I'd second this request -- right now there is no way to automate stability testing and re-optimization of wave functions, which is extremely useful functionality. Otherwise one is stuck spamming scripts with repeated 
```
mo1 = mf.stability()[0]
dm1 = mf.make_rdm1(mo1, mf.mo_occ)
mf = mf.run(dm1)
mf.stability()
```
calls, which is inefficient and/or impractical for large systems and throughputs. A boolean flag would at least let us wrap the above calls in a loop with a proper termination.

hebrewsnabla:
I'm going to open a PR to add the boolean flag.

PR #1054, opened 2021-09-10
Add boolean flag for stability functions and an example for stability-check loop

hebrewsnabla:
* Add boolean flag for each stability function in `scf.stability`. Now `uhf_internal()` will return `mo, stable`, and `uhf_stability()` will return `mo_i, mo_e, stable_i, stable_e`.
* Add an example for stability-check loop in `examples/scf/17-stability.py`
 related issue #873


Majokux:
A reasonable change, although it does change the function signature of `mf.stability()`.
And in principle the additional `stable` return value only save one line in a script, i.e. `stable = (mo_i is mf.mo_coeff)`

If we do decide to change the function signature anyway, how about this change:
If stable, `mf.stability()` returns `True` - otherwise it returns the perturbed MOs.

hebrewsnabla:
> A reasonable change, although it does change the function signature of `mf.stability()`.
> And in principle the additional `stable` return value only save one line in a script, i.e. `stable = (mo_i is mf.mo_coeff)`
> 
> If we do decide to change the function signature anyway, how about this change:
> If stable, `mf.stability()` returns `True` - otherwise it returns the perturbed MOs.

I think, in PySCf examples, PySCF recommends calling `mf.stability()` directly and thus it seems that the developers don't care much about the returned value. So I change the number of returned values. Maybe it will bother some users who developing their programs based on this code. But, my changes won't affect something like `mo1 = mf.stability()[0]`, which is also an existent recommended way to call the function.

I think 
> If stable, mf.stability() returns True - otherwise it returns the perturbed MOs.

is a little bit more confusing. And it breaks `mo1 = mf.stability()[0]`.

jjgoings:
> If stable, mf.stability() returns True - otherwise it returns the perturbed MOs.

In addition, it changes output type depending on result which will likely cause problems unless you are careful with type handling. I think adding the additional boolean output is the least disruptive change.

sunqm:
I think changing the returns of this function is acceptable. In case there are users running scripts depending on the old API of this function, what about adding a keyword argument `return_status=False` to control whether to return the status? Then you can obtain the status by passing return_status=True in your code. Please also update the docstring for this optional keyword.

hebrewsnabla:
Everything updated now.

sunqm (approved):

sunqm (approved):

PR #1055, opened 2021-09-12
Add serval ordinary/auxiliary basis sets to suppress df.make_auxbasis errors

ajz34:
Hi devs!

It seems that 6-31(1)+G mp2fit auxiliary basis `heavyaug-cc-pvnz-ri` does not exists.
When calling `df.make_auxbasis` using 6-31(1)+G basis set could cause error.

I try to remove those two default auxiliary basis, thus allowing automatic etb auxiliary basis generation to 6-31(1)+G. Or maybe replacing `heavyaug-cc-pvnz-ri` by `aug-cc-pvnz-ri` is another option, but I'm not sure whether that's a correct approach.


ajz34:
After discussion with [hebrewsnabla](https://github.com/hebrewsnabla), and looking at the code comment, probably some more default auxiliary basis sets could be changed. I will change contents of this pull request later.

ajz34:
Current changes:
- `def2-svp-rifit` -> `def2-svp-ri`
- add ordinary basis `aug-cc-pvdpdz` and `cc-pvdpdz`
- add auxiliary basis `heavy-aug-cc-pvdz-ri`, `heavy-aug-cc-pvtz-ri`, `aug-cc-pv5z-ri`

sunqm (approved):

Issue #1041, opened 2021-08-25
is PYSCF_MAX_MEMORY the memory per core or for the entire system?


HehnLukas:
I've noticed for some parallel computations that when I set the maximum memory, which I thought was for the entire system, close to the total available memory, that my calculations are no longer consistent. By accident I found that reducing the max memory causes them to give consistent results. 
This has made me assume that it sets the memory per core, and that somehow giving it too much memory causes some issues. Can someone confirm if this is the case? I have not found the relevant section of the code.

sunqm:
max_memory is a global setting. Python code uses this parameter for most big array allocations. Certain python calls have significant overhead in memory footprint (like lib.einsum) . Some were considered but there left some parts of the code with bad estimation of memory usage.

C code has extra memory footprint but nothing to do with `max_memory`. Memory allocated in C is not restricted to max_memory. Most C code uses very small amount of local memory per thread. It sometimes leads to more memory usage than `max_memory`.

What's the problem size of your system? NEVPT2 C code roughly needs `n_cas^4` words per thread plus `100 * n_cas^4` words global buffer.

HehnLukas:
Yeah, so I am very confused. I decided to run more tests on a water molecule, as that way the calculations are quicker. It seems if the max memory is set too high, the calculations will no longer be consistent, and for smaller molecules, too high is still at a quite low memory.

My initial tests were with a system consisting of one Fe ion and 6 water molecules, using the def2-TZVP basis set. here, I ran the calculations on 40 cores, and with 150GB of memory allocated to the cluster. When I set PYSCF_MAX_MEMORY to 112000, the resulting casscf orbitals are somewhat different each time, leading to different Nevpt2 energies. However, when I set PYSCF_MAX_MEMORY to 11200, the calculations are identical each time, and the Nev energies stay the same. Even dropping the PYSCF_MAX_MEMORY to 0 results in the same results as 11200. 

With water (here i use 5 cores), I get the same problem, but the threshold is much lower. For example, setting the max_memory to just 500 results in:

Cas1.out:Nevpt2 Energy = -0.248842607769583
Cas2.out:Nevpt2 Energy = -0.247865009848661
Cas3.out:Nevpt2 Energy = -0.248110570899806
Cas4.out:Nevpt2 Energy = -0.246335547520711
Cas5.out:Nevpt2 Energy = -0.248938029269932
Cas6.out:Nevpt2 Energy = -0.248799774140834
[...]

PR #1070, opened 2021-10-02
Fix NEVPT2 (issue #1041)

sunqm:


pyscf/mcscf/casci.py
```diff
@@ -283,7 +285,39 @@ def cas_natorb(mc, mo_coeff=None, ci=None, eris=None, sort=False,
...
+        fock_ao = mc.get_fock(mo_coeff, ci, eris, casdm1, verbose)
+
+        def _diag_subfock_(idx):
+            c = mo_coeff1[:,idx]
+            fock = reduce(numpy.dot, (c.conj().T, fock_ao, c))
+            w, c = mc._eig(fock, None, None, orbsym[idx])
+            mo_coeff1[:,idx] = mo_coeff1[:,idx].dot(c)
+
+        if occ2_idx.size > 0:
+            log.debug('Active orbitals %s (occs = %s) are canonicalized with core orbitals',
```
jamesETsmith, line 310: This is more of a stylistic comment, but it might be safer to use `log.warn` instead of `log.debug`, just so users are more likely to see these comments

jamesETsmith:
Thanks for tracking this down @sunqm! The fixes look good to me, my only suggestion is just to change the level of the logging so users are more likely to see it.

fishjojo (approved):

jamesETsmith (approved):

PR #1075, opened 2021-10-10
fix_blas_libcint: remove duplicated definition of BLAS library

acastanedam:
New versions of cmake (>3.19) lead to compilation problems due the use
of both definitions `BLA_VENDOR` and `BLAS_LIBRARIES`. Removing the
former fix this issue and it is compatible with older cmake versions.


jamesETsmith:
@acastanedam thanks for bringing this up, would you mind including a partial or complete copy of the error here? I think it'd help other people who might run into this issue in the near future.

sunqm:
The next version of libcint will drop the dependence to BLAS library. So likely this patch will not be needed in the future.

jamesETsmith:
@acastanedam, thanks for looking into this. CMake converting semicolons into spaces was the same problem I ran into when I tried out this PR. They should be fixed by #1077.

acastanedam:
Hi @jamesETsmith, @sunqm, then everything is working, I guess we can close the MR. Thanks for checking

jamesETsmith:
@acastanedam, thanks for pointing this out, we appreciate the help!

PR #1107, opened 2021-11-14
Fix a bug in the inconsistency of betaSigma for the spin magnetizatio…

hoshishin:
Fix a bug in the inconsistency of betaSigma for the spin magnetization vector and potential.

### DIFF
--- a/pyscf/dft/r_numint.py
+++ b/pyscf/dft/r_numint.py
@@ -143,6 +143,32 @@ def _vxc2x2_to_mat(mol, ao, weight, rho, vrho, non0tab, shls_slice, ao_loc):
     mat+= _dot_ao_ao(mol, aob, aow, non0tab, shls_slice, ao_loc)
     return mat
 
+def _vxc2x2_to_mat_S(mol, ao, weight, rho, vrho, non0tab, shls_slice, ao_loc):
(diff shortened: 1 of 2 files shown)


pyscf/dft/r_numint.py
```diff
@@ -191,7 +217,7 @@ def r_vxc(ni, mol, grids, xc_code, dms, spin=0, relativity=0, hermi=1,
                 matLL[idm] += _vxc2x2_to_mat(mol, ao[:2], weight, rho, vrho,
                                              mask, shls_slice, ao_loc)
                 if with_s:
-                    matSS[idm] += _vxc2x2_to_mat(mol, ao[2:], weight, rho, vrho,
+                    matSS[idm] += _vxc2x2_to_mat_S(mol, ao[2:], weight, rho, vrho,
```
sunqm, line 220: Does it equal to `matSS[idm] - _vxc2x2_to_mat(mol, ao[2:], ...`?

hoshishin: I think it does not equal because the density-dependent part should not change. The only difference is there should be a minus of the small component for the spin density-dependent part.

sunqm (approved):

Issue #1119, opened 2021-12-03
Please rename aux.py


cvjjm:
Unfortunately Microsoft OneDrive has [completely absurd restrictions](https://www.minitool.com/news/onedrive-name-or-type-isnt-allowed.html) on file/folder names. In particular it does not allow files containing the character sequence "aux." in their filename.

PySCF contains one file that conflicts with this restriction, namely the file `pyscf/agf2/aux.py`.

This leads to the situation that when someone clones/copies the pyscf source to some folder on their PC that is synced with OneDrive. The sync completely stop.

Now, I agree that this should be considered a bug in OneDrive and be fixed there, but this is unfortunately incredibly unlikely. 

So I please as the PySCF developers to slightly change the name of `pyscf/agf2/aux.py`, to maybe to `pyscf/agf2/aux_space.py`.

obackhouse:
Pretty awesome bug, submitted a PR to address this (#1120).

TNTworks:
> Now, I agree that this should be considered a bug in OneDrive and be fixed there

No, this is not a bug, but a feature. This is a well-known limitation of NTFS. AUX is forbidden filename with any extension.

bruceburge:
Here is the list of documented restriction on file and folder names on windows, specifically related to the reserved names. 
_**Do not use the following reserved names for the name of a file:

CON, PRN, AUX, NUL, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, and LPT9. Also avoid these names followed immediately by an extension; for example, NUL.txt is not recommended.**_

[Source](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN)

harryvederci:
Maybe send a patch to the OneDrive repo? Oh waaait... That's not how Microsoft operates, is it?

cvjjm:
#1120  looks like a good solution to me.

TaiAurori:
Windows + NTFS always leads to a few fun ways to break things!

waffletower:
> > Now, I agree that this should be considered a bug in OneDrive and be fixed there
> 
> No, this is not a bug, but a feature. This is a well-known limitation of NTFS. AUX as a filename with any extension is forbidden by design.

[...]

PR #1120, opened 2021-12-03
Addresses `aux.py` filename bug (issue #1119)

obackhouse:
This PR changes the name of `pyscf/agf2/aux.py` to `pyscf/agf2/aux_space.py`, as `aux.*` is a reserve filename pattern on OneDrive (see #1119). This should provide reasonable backwards compatibility.


cvjjm:
Awesome! Thanks so much!

fishjojo (approved):

PR #1124, opened 2021-12-08
Use normalized error basis in DIIS

susilehtola:
Currently the DIIS code is diagonalizing an error matrix that is inbalanced: the first row and first column are filled with ones, whereas the other entries in the matrix are dot products of the error vectors that can easily range to `1e20` even in small examples.

The better choice is to normalize the error vectors before diagonalization, and then roll the normalization back into the eigenvectors after diagonalization.

### DIFF
--- a/pyscf/lib/diis.py
+++ b/pyscf/lib/diis.py
@@ -251,7 +251,13 @@ class DIIS(object):
         g = numpy.zeros(nd+1, h.dtype)
         g[0] = 1
 
-        w, v = scipy.linalg.eigh(h)
+        # eig works better when the errors have similar norm
+        print('diag h {}'.format(numpy.diag(h)))
+        normlz = numpy.power(numpy.asarray([hd if hd!=0.0 else 1.0 for hd in numpy.diag(h)]), -0.5)
+        hnorm = numpy.dot(numpy.diag(normlz), numpy.dot(h, numpy.diag(normlz)))
+        w, vnorm = scipy.linalg.eigh(hnorm)
+        v = numpy.dot(numpy.diag(normlz), vnorm)
+
         if numpy.any(abs(w)<1e-14):
             logger.debug(self, 'Linear dependence found in DIIS error vectors.')
             idx = abs(w)>1e-14


sunqm:
Diagonalization in diis extrapolation is to check whether there are linearly dependent vectors before solving the linear equation in this basis space. I don't see why the treatment of imbalanced elements is necessary here. The modification changes the eigenvalues thus the checks of basis linear dependency. I think it will change the output of extrapolation in an unpredictable way. Do you have an example to show the necessity of the modification?

susilehtola:
@verena-neufeld yeah, the issue is exactly that one should not calculate orbital gradients including linearly dependent functions that have not been included in the orbital basis. Does the PBC code remove linear dependencies automatically unlike the molecular code?

verena-neufeld:
Thanks! I don't think it automatically removes linear dependencies.

verena-neufeld:
If checking for linear dependencies is crucial here, would one option be to first evaluate the eigenvalues as before, i.e. `w, v = scipy.linalg.eigh(h)`,  check for linear dependencies and then, if everything is ok, redo with the changes here?

susilehtola:
> If checking for linear dependencies is crucial here, would one option be to first evaluate the eigenvalues as before, i.e. `w, v = scipy.linalg.eigh(h)`, check for linear dependencies and then, if everything is ok, redo with the changes here?

It's not crucial to *check* for linear dependencies; what *is* crucial is that the DIIS error is computed using the same orbital basis in which the SCF is performed.

The approach of this PR is how DIIS is invariably implemented across quantum chemistry codes, and is what Pulay originally described 40 years ago.

sunqm:
> This modification can sometimes help convergence issues in pbc problems it seems. I'm posting an example below. As small comments, I wonder whether the print function should be changed to a logging function and complex `h` have to be accounted for (I've only briefly gone over the code, so do correct me if I'm wrong).
> 
> An example where this change has helped is a modified version of example `pbc/22-k_points_ccsd.py` where I compressed the cell, see
> 
> ```
> #PySCF copyright. Modified from example pbc/22-k_points_ccsd.py
> from functools import reduce
> import numpy
> from pyscf.pbc import gto, scf, cc
> 
> cell = gto.Cell()
> cell.atom='''
> C 0.000000000000   0.000000000000   0.000000000000
> C 1.551   1.551   1.551
> '''
> cell.basis = 'gth-dzvp'
> cell.pseudo = 'gth-pade'
> cell.a = '''
> 0.000000000, 3.102, 3.102
> 3.102, 0.000000000, 3.102
> 3.102, 3.102, 0.000000000'''
> cell.unit = 'B'
> cell.verbose = 5
> cell.build()
> 
> kpts = cell.make_kpts([2,2,1])
> kmf = scf.KRHF(cell).rs_density_fit()
> kmf.kpts = kpts
> ehf = kmf.kernel()
> 
> mycc = cc.KCCSD(kmf)
> mycc.kernel()
> ```
> 
> Without this PR's modification, I don't get a converged CCSD energy. With this modification, I get `E(RCCSD) = -11.06303211658363 E_corr = -0.2410500162324277`.
> 
> When I replace 1.551 by 1.553 and 3.102 by 3.106 (i.e. compress cell a little less), both the modified code and unmodified code converge, to a CCSD corr energy of `-0.241124`, and at 1.555 and 3.11, I get `-0.241203` hartree. So the converged energy in the original snippet posted here is consistent with these other energies.
> 
> @sunqm @susilehtola

The convergence issue here may be related to the accuracy of density fitting integrals. If you set a high integral accuracy, CCSD converges quickly. Using the updated density fitting integrals implementation #1622, I can converge CCSD in 10 iterations

sunqm:
> > If checking for linear dependencies is crucial here, would one option be to first evaluate the eigenvalues as before, i.e. `w, v = scipy.linalg.eigh(h)`, check for linear dependencies and then, if everything is ok, redo with the changes here?
> 
> It's not crucial to _check_ for linear dependencies; what _is_ crucial is that the DIIS error is computed using the same orbital basis in which the SCF is performed.
> 
> The approach of this PR is how DIIS is invariably implemented across quantum chemistry codes, and is what Pulay originally described 40 years ago.

Normalizing h matrix here will not remove the linear dependency in error vector. It does not help solve the linear equation here if h is ill conditioned.

I agree @verena-neufeld  that it is probably better to first try linalg.solve and use the eigenvalue solver as a backup option.

PR #1134, opened 2021-12-14
Raise error if specified occupations cannot be satisfied

### LINKED ISSUE #1133: No error check for irrep_nelec
Running an excited-state calculation for hydrogen yields unexpected results
```
from pyscf import gto, scf

mol = gto.M(
    verbose = 3,
    atom = '''                                                                                                                                                                                
H''',
    basis = 'unc-aug-pc-0',
    symmetry = True,
    spin = 1
)
mf = scf.UHF(mol)
mf.irrep_nelec = {'p+1' : 1}
mf.kernel()

mol = gto.M(
    verbose = 3,
    atom = '''                                                                                                                                                                                
H''',
    basis = 'unc-aug-pc-1',
    symmetry = True,
    spin = 1
)
mf = scf.UHF(mol)
mf.irrep_nelec = {'p+1' : 1}
mf.kernel()
```
The output is
```
converged SCF energy = -0.496632143462233  <S^2> = 0.75  2S+1 = 2
converged SCF energy = -0.105156478066618  <S^2> = 0.75  2S+1 = 2
```
The first run is actually the `1s` state, not the targeted `2p` state. The issue is that the aug-pc-0 basis set does not have polarization functions.

Instead of yielding an incorrect result, the program should crash if there are not sufficiently many orbitals to satisfy the desired occupations.

susilehtola:
Closes #1133

### DIFF
--- a/pyscf/scf/uhf_symm.py
+++ b/pyscf/scf/uhf_symm.py
@@ -314,10 +314,6 @@ class SymAdaptedUHF(uhf.UHF):
     def build(self, mol=None):
         if mol is None: mol = self.mol
         if mol.symmetry:
-            for irname in self.irrep_nelec:
-                if irname not in self.mol.irrep_name:
-                    logger.warn(self, 'No irrep %s', irname)
-
             hf_symm.check_irrep_nelec(mol, self.irrep_nelec, self.nelec)
         return uhf.UHF.build(self, mol)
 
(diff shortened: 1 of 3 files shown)


sunqm (approved):

pyscf/scf/uhf_symm.py
```diff
@@ -312,12 +312,9 @@ def dump_flags(self, verbose=None):
         return self
 
     def build(self, mol=None):
+        print('in uhf symm build')
```
sunqm, line 315: Remove this line

sunqm (approved):

pyscf/scf/uhf_symm.py
susilehtola: Removed, thanks for spotting it

fishjojo (approved):

Issue #1140, opened 2021-12-17
PBC GDF gives segfault for certain calculations


fishjojo:
pyscf version: 2.0.1

input:
```
from pyscf.pbc import gto, scf, df

a = 2.9
nat = 10

cell = gto.Cell()
for iat in range(0, nat):
    cell.atom.append(['Au', (iat*a, 0, 0)])

cell.exp_to_discard = 0.1
cell.ke_cutoff = 200
cell.max_memory = 20000

cell.build(
        unit = 'angstrom',
        a = [[a*nat,0,0],[0,1,0],[0,0,1]],
        dimension = 1,
        basis = 'def2-svp',
        ecp = 'def2-svp',
        verbose=7,
        )

nks = 1
kpts = cell.make_kpts([nks,1,1], scaled_center=[0,0,0])
mf = scf.KUHF(cell, kpts=kpts).density_fit()
mf = mf.newton()
mf.max_cycle = 500
ig = mf.get_init_guess()
ig[1,:,:,:] = 0

e = mf.kernel(dm0=ig)
print('uhf energy per atom = ', e/nat)
```

output:
```
******** <class 'pyscf.pbc.df.df.GDF'> ********
mesh = [83, 49, 49] (199283 PWs)
auxbasis = None
eta = 0.2
exp_to_discard = 0.1
_cderi_to_save = /scratch/global/xzhang8/tmpt9gs7l9r
len(kpts) = 1
    kpts = [[0. 0. 0.]]
Default auxbasis def2-svp-jkfit is used for Au def2-svp
num shells = 510, num cGTOs = 2470
Drop 50 primitive fitting functions
make aux basis, num shells = 460, num cGTOs = 2280
auxcell.rcut 17.70181012931563
make compensating basis, num shells = 70, num cGTOs = 490
chgcell.rcut 11.482117703874595
    CPU time for 3c2e     27.59 sec, wall time      0.99 sec
Num uniq kpts 1
uniq_kpts [[0. 0. 0.]]
max_memory 17838.278656 (MB)  blocksize 201244
Cholesky decomposition for j2c at kpt 0
Symmetry pattern (k - [0. 0. 0.])*a= 2n pi
    make_kpt for uniq_kptji_ids [0]
kpt = [0. 0. 0.]
adapted_ji_idx = [0]
```

stack backtrace:
```
[...]

PR #1144, opened 2021-12-24
Fix segfault in GDF which is caused by the failure of malloc

sunqm:
The ft_ao code allocates a buffer too large to be fit into virtual memory. Improve the estimation of buffer size. Issue #1140 may be related.


fishjojo (approved):

fishjojo:
Issue #1140 is fixed with this PR.

Issue #1111, opened 2021-11-16
2.0.1 fails to build with libcint-5.0.0: error: typedef redefinition with different types ('struct CINTEnvVars' vs 'struct CINTEnvVars')


yurivict:
```
In file included from /wrkdirs/usr/ports/science/py-pyscf/work-py38/pyscf-2.0.1/pyscf/lib/gto/ft_ao.c:48:
/wrkdirs/usr/ports/science/py-pyscf/work-py38/pyscf-2.0.1/pyscf/lib/gto/ft_ao.h:86:3: error: typedef redefinition with different types ('struct CINTEnvVars' vs 'struct CINTEnvVars')
} CINTEnvVars;
  ^
/usr/local/include/cint.h:225:3: note: previous definition is here
} CINTEnvVars;
  ^
1 error generated.
```

susilehtola:
I also ran into this issue trying to build PySCF 2.0.1 against libcint 5.0.0 in the development version of Fedora (rawhide).

ccalaza:
https://github.com/pyscf/pyscf/issues/1126#issuecomment-989944491

fishjojo:
I guess the structure defined in ft_ao.h can be commented out. @sunqm are there any other things need to be changed in pyscf?

sunqm:
Please keep libcint at v4.* . cint v5.* is working but pyscf as of 2.0.1 is not compatible with cint v5.*

susilehtola:
Which PySCF will be compatible with libcint v5?

sunqm:
Next minor release maybe. The upgrade of libcint interface is coupled with some new features and PRs. Currently it still has problems to merge the new features. It makes the upgrade a little bit complicated

sunqm:
Fixed in PR #1384

PR #1145, opened 2021-12-25
Update to libcint v5.0

sunqm:
Make libcgto module compatible with both v4 and v5 of libcint


susilehtola:
This solved #1111, thanks!

fishjojo (approved):

PR #1158, opened 2022-01-11
Adds _make_df_eris_outcore to UCCSD

Majokux:
I added an outcore AO->MO routine for UCCSD + DF, which I thought would be useful to add to master.

The implementation was tested in two limits (examples below):
1) spin-symmetric solution of H2O@cc-pVDZ, comparing RCCSD+DF vs UCCSD+DF, which leads to a difference of ~1e-10 in the correlation energy
2) The radical NO2, comparing non-DF UCCSD/cc-pVDZ  with UCCSD/cc-pVDZ+DF(aug-cc-pV5Z-ri), which leads to a difference of ~2e-6, which I guess is reasonable for the remaining DF error?

[test_1.py.txt](https://github.com/pyscf/pyscf/files/7848058/test_1.py.txt)
[test_2.py.txt](https://github.com/pyscf/pyscf/files/7848067/test_2.py.txt)

### DIFF
--- a/pyscf/cc/__init__.py
+++ b/pyscf/cc/__init__.py
@@ -124,7 +124,9 @@ def UCCSD(mf, frozen=None, mo_coeff=None, mo_occ=None):
         mf = scf.addons.convert_to_uhf(mf)
 
     if getattr(mf, 'with_df', None):
-        raise NotImplementedError('DF-UCCSD')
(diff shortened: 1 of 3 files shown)


sunqm:
Please add a small test for DF-UCCSD

sunqm (approved):

Issue #898, opened 2021-04-11
Interface to call basis exchange python API


sunqm:
It would be useful to download or update basis set db using basis exchange python api

https://github.com/MolSSI-BSE/basis_set_exchange

sunqm:
A relevant example https://github.com/pyscf/pyscf/blob/master/examples/gto/31-basis_set_exchange.py

susilehtola:
See also https://github.com/MolSSI-BSE/basis_set_exchange/issues/210

susilehtola:
There's a preliminary implementation in https://github.com/MolSSI-BSE/basis_set_exchange/pull/235 that can be used e.g. as
```
from pyscf import gto, scf
import basis_set_exchange

mol = gto.Mole()
mol.verbose = 4
mol.atom = '''
Xe 0.0 0.0 0.0'''
mol.basis = basis_set_exchange.api.get_basis(name='def2-TZVP', fmt='pyscf', elements='Xe')
mol.ecp = basis_set_exchange.api.get_basis(name='def2-TZVP', fmt='pyscfecp', elements='Xe')
mol.symmetry = 1
mol.build()

mf = scf.RHF(mol).apply(scf.addons.remove_linear_dep_)
mf.kernel()
```

susilehtola:
The question to the PySCF community is: would you like to deprecate the built-in basis set library in favor of using the Basis Set Exchange? It doesn't make much sense having to maintain basis set libraries in every program.

susilehtola:
Note that one can use the BSE's Python interface to get e.g. multiply augmented basis sets, as well as automatically formed auxiliary basis sets for RI. (I guess some of these functionalities might already be in PySCF though...)

For instance, to get a triply-augmented t-aug-cc-pVQZ basis set, one would use `get_basis(name='aug-cc-pVQZ', fmt='pyscf', augment_diffuse=2)`, and the corresponding AutoAux basis for `with_df` can be gotten with `get_basis(name='aug-cc-pVQZ', fmt='pyscf', augment_diffuse=2, get_aux=1)`.

tberkel:
The recent mention in #1151 reminded me of this discussion, and I think I support deprecating the built-in basis set library in favor of the Basis Set Exchange. If we make the BSE a dependency, then we can use the API internally and hide it from most users.

jamesETsmith:
@tberkel, I like the idea of handling the BSE API internally and it should be easy to do.

susilehtola:
[...]

PR #1166, opened 2022-01-14
Add low-level interface to Basis Set Exchange

susilehtola:
Partial solution to #898: these routines parse the JSON type data that `basis_set_exchange.api.get_basis` returns.


sunqm (approved):

susilehtola:
Thank you @sunqm !!

PR #1179, opened 2022-02-02
Sacasscf grad cleanup

MatthewRHermes:
In SA-CASSCF nuclear gradients, attach the convergence flag and the final Lagrange multipliers to the class instance so they can be inspected and manipulated after the gradient calculation is complete.

Also, cleanup: wrap grad/lagrange.py and grad/sacasscf.py to 99 columns max, and simplify or remove some excessive comments.


MatthewRHermes:
Forgot something.

Issue #1173, opened 2022-01-27
Error comes out when building from source.


beimojiguang:
I am trying to compell pyscf 2.0.1 from newest source code by using cmake 3.22.1 .
The error shows as follow:

```
[ 31%] Built target np_helper
[ 32%] Building C object gto/CMakeFiles/cgto.dir/fill_int2c.c.o
In file included from /home/huangzigeng/pyscf/pyscf/lib/gto/fill_int2c.c:22:0:
/home/huangzigeng/pyscf/pyscf/lib/deps/include/cint.h:162:40: error: ‘SIMDD’ undeclared here (not in a function)
 #define ALIGNMM __attribute__((aligned(SIMDD*8)))
                                        ^
/home/huangzigeng/pyscf/pyscf/lib/deps/include/cint.h:253:9: note: in expansion of macro ‘ALIGNMM’
         ALIGNMM double ai[SIMDD];
         ^~~~~~~
gto/CMakeFiles/cgto.dir/build.make:75: recipe for target 'gto/CMakeFiles/cgto.dir/fill_int2c.c.o' failed
make[2]: *** [gto/CMakeFiles/cgto.dir/fill_int2c.c.o] Error 1
CMakeFiles/Makefile2:376: recipe for target 'gto/CMakeFiles/cgto.dir/all' failed
make[1]: *** [gto/CMakeFiles/cgto.dir/all] Error 2
Makefile:90: recipe for target 'all' failed
make: *** [all] Error 2
```

The pyscf has been build successfully when I use cmake 3.21.3 or 3.18. But some warnings still come out as following:

```
/opt/tiger/pyscf/pyscf/lib/vhf/nr_direct.c: In function 'CVHFdot_nrs2ij':
/opt/tiger/pyscf/pyscf/lib/vhf/nr_direct.c:48:19: warning: unused variable 'jsh1' [-Wunused-variable]
   48 |         const int jsh1 = jshls[1]; \
      |                   ^~~~
/opt/tiger/pyscf/pyscf/lib/vhf/nr_direct.c:116:17: note: in expansion of macro 'DECLARE_ALL'
  116 |                 DECLARE_ALL;
      |                 ^~~~~~~~~~~
/opt/tiger/pyscf/pyscf/lib/vhf/nr_direct.c: In function 'CVHFdot_nrs2kl':
[...]

PR #1186, opened 2022-02-12
Adding CMake Option to Build qcint 

jamesETsmith:
## Summary
This PR adds the cmake option to build `qcint` instead of vanilla `libcint`. Users can specify from the command line like this:

```
cmake -B build -DUSE_QCINT=ON
```

When users specify `USE_QCINT=ON` and _do not_ define `BUILD_MARCH_NATIVE`, they are warned that `qcint` may fail to compile, which happens when `CMAKE_C_FLAGS` doesn't contain any flags anything to turn vectorization on. This problem was behind #1173.

## Next Steps
This is a somewhat loose check. If users specify `CMAKE_C_FLAGS=-O3`, but not `BUILD_MARCH_NATIVE`, then the warning is unnecessary. At the same time, I think it would complicate `CMakeLists.txt` to try to check all ways that users could turn on vectorization. Do other people have thoughts on this?

### DIFF
--- a/pyscf/lib/CMakeLists.txt
+++ b/pyscf/lib/CMakeLists.txt
@@ -140,11 +140,20 @@ set(CXX_LINK_TEMPLATE "<CMAKE_CXX_COMPILER> <CMAKE_SHARED_LIBRARY_CXX_FLAGS> <LA
 include(ExternalProject)
 option(BUILD_LIBCINT "Using libcint for analytical gaussian integral" ON)
 option(WITH_F12 "Compling F12 integrals" ON)
+option(USE_QCINT "Using the qcint library (optimized for x86-64) for gaussian integral evaluation" OFF)
+
 if(BUILD_LIBCINT)
 if(NOT EXISTS "${PROJECT_SOURCE_DIR}/deps/include/cint.h")
+  set(LIBCINT_GIT https://github.com/sunqm/libcint.git) # libcint is a portable, cross-platform implementation
+  if (USE_QCINT)
+    set(LIBCINT_GIT https://github.com/sunqm/qcint.git) # qcint is an optimized implementation for x86-64 architecture
+    if(NOT BUILD_MARCH_NATIVE)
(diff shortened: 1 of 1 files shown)


susilehtola:
I think I fixed the bug building qcint when BUILD_MARCH is off by just switching on SSE3...

jamesETsmith:
@susilehtola, thanks for taking a look. Indeed switching on SSE3 (or AVX or AVX512f) fixes the problem, but I wasn't sure of the best way to check (from within CMake) if any of those have been turned on, do you know? If you have a good way, then we can raise an error during the CMake configuration if they aren't turned on. Since I didn't know of any good way to do it, I left it as a warning, like I mentioned in the original post, it's a "loose check".

susilehtola:
Well, the fix from last May should still be in place in CMakeLists.txt
```
if(BUILD_MARCH_NATIVE)
  message("Build with -march=native")
  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -march=native -O2")
else()
  # This appears to be the lowest version that compiles
  message("Build with -march=sse3")
  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -msse3 -O2")
endif()
```
Is it not included in the version the bug was filed on?

jamesETsmith:
@susilehtola, thanks for pointing that out, I somehow missed that part of `qcint`'s CMakeLists.txt file. But to answer your question: yes that is part of the cmakelists.txt file for the version of `qcint` I'm using (5.1.1). Oddly enough, when I specify that only I want to build `qcint` (i.e. not specifying any compiler flags), the build of just `qcint` succeeds, but when I try to build everything at once it fails. Just to be clear:

1. Fails for PySCF because `qcint` build step fails
```
cmake -B build -DUSE_QCINT=ON
cmake --build build --parallel --verbose
```

2. Succeeds for `qcint`, fails for PySCF
```
cmake -B build -DUSE_QCINT=ON
cmake --build build --parallel --target libcint --verbose    # Succeeds
cmake --build build --parallel --verbose                     # Fails
```

sunqm (approved):

Issue #1161, opened 2022-01-11
k2gamma assertion error: complex MO coefficients


Simonsays095:
There are several points in the k2gamma code (specifically in the `mo_k2gamma` function) where there are assertions with no description of what causes the issue or how to resolve it. Mostly these are related to matrices that should be real, having significant imaginary components. When I come across one of these assertion errors, what should I do to reduce the errors so that my system/calculation can pass the assertion?

Here is an example that triggers an assertion error:
```
from pyscf.pbc import scf
from pyscf.pbc.gto import Cell
import numpy as np


# ----- Construct a 1-dimensional O2 chain unit cell ----- #
cell = Cell()
cell.atom = [
    ["O", (0,0,0)],
    ["O", (2,0,0)]
]

cell.dimension = 1
cell.basis = '6-31g'
cell.a = np.eye(3)*[4,10,10]
cell.verbose = 4
cell.unit = 'A'
cell.build()

kpts = cell.make_kpts([3,1,1])

# -------- KHF calculation ------- #
mf = scf.KRHF(cell, kpts).density_fit()
e_hf = mf.kernel()


from pyscf.pbc.tools.k2gamma import mo_k2gamma
mo_k2gamma(cell, mf.mo_energy, mf.mo_coeff, mf.kpts, kmesh=None)
```

Majokux:
I and @obackhouse  have been struggling with this issue for a while, and would like to know the answer as well.
It also happens for 3D materials, especially if the band gap is small and/or the point group symmetry of the unit cell high.
In principle one should always be able to find a real set of orbitals at the Gamma-point, but the routine in `k2gamma` does not seem to be robust in this respect. Maybe @zhcui can help?

zhcui:
Hi all,

This issue comes from the degeneracy of the system. The mo occupation patterns are different for different k points:

[[2. 2. 2. 2. 2. 2. 2. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
 [2. 2. 2. 2. 2. 2. 2. 2. 0. 0. 0. 0. 0. 0. 0. 0. 0. 0.]
 [2. 2. 2. 2. 2. 2. 2. 2. 2. 0. 0. 0. 0. 0. 0. 0. 0. 0.]]

The patterns break the symmetry of the phases among different k points. So the MO's cannot be realized.

Best,
Zhihao

Majokux:
Hi Zhihao, many thanks for your response.
[...]

PR #1197, opened 2022-02-25
Addition of (k,-k)-symmetry addon for KSCF

maxnus:
When performing k-point sampled SCF calculations with PySCF, the symmetry between the MO \phi(k) and the conjugated MO \phi(-k)* is frequently being broken. Often, one would like to retain this symmetry, which for example should be an exact symmetry in crystals with centrosymmetry. As an example, this is the default behaviour in VASP, see the entry for `ISYM=0` [here](https://www.vasp.at/wiki/index.php/ISYM). Retaining this symmetry also allows the solution to remain real-orbital supercell representable (see also issue #1161).

Ideally, this symmetry should be implemented in a larger effort to support point-group symmetry in crystals and sampling of the irreducible BZ only. However, this is a somewhat larger project and I am not sure if work is already being done in this direction.
Since I require (k,-k)-symmetric solution for some of my work, I instead implemented a more lightweight solution, allowing the decoration of `KRHF` and `KUHF` according to the addon syntax: `kmf = pyscf.pbc.scf.addons.kconj_symmetry_(kmf)`.
I think this could be useful for more people, until there is proper support for spacegroup symmetry.

An example is given in `pyscf/examples/pbc/43-kconj-symmetry.py`. Most of the code here is orthogonal to other parts of PySCF; the notable exception to this are the routines `get_occ` in `pbc/scf/khf.py` and  `pbc/scf/kuhf.py`.
These routines are currently buggy, in that the lead to a wrong number of occupied MOs, if an **exact** (i.e. to machine precision) degeneracy is present at the Fermi-level. I modified these functions, to crash out in this case instead (which is still better than stealthily changing the number of electrons).

As an alternative, `get_occ` could fill as many of the degenerate orbitals at the Fermi level, as there are remaining electrons, in the hope that the degeneracy only exists temporarily during the SCF loop and will disappear in the following iterations. However, an exact degeneracy is unlikely to occur when not explicitly enforcing symmetry and I thus didn't modify `get_occ` in this way for the time being.


zhcui:
Hi Max,

I think this is already implemented in the [kpoints-symmetry branch.](https://github.com/pyscf/pyscf/tree/kpoint_symm)

januseriksen:
Hi @maxnus, what's the status of this PR? Or has a similar option been implemented elsewhere, @zhcui? If so, how to enforce this type of _(k,-k)_-symmetry in an KSCF simulation? Cheers to both of you.

zhcui:
You may try something like:
```
kmesh = [3, 3, 3]
kpts_symm = cell.make_kpts(kmesh, with_gamma_point=True, space_group_symmetry=False, time_reversal_symmetry=True)                  
```

sunqm (approved):

verena-neufeld:
Since kpts symm has been merged and there is a workaround posted by @zhcui , I wonder whether this PR should be altered? I'm saying that without having had a close look at this PR.

fishjojo:
I'm going to close this PR as time-reversal and space-group symmetries are already available in pyscf. Please feel free to contribute your code to [pyscf-forge](https://github.com/pyscf/pyscf-forge) if it adds additional functionalities. And we can merge it with to the main source code once appropriate.

PR #1242, opened 2022-04-22
Ensure deterministic result from pyscf.lib.einsum

basilib:
Due to the usage of sets in `pyscf.lib.contract`, `pyscf.lib.einsum` is non-deterministic even when running on a single thread, unlike `np.einsum`. Minimal example below, fixed by this PR. 

```
from pyscf import lib
import numpy as np
np.random.seed(1)

a = np.random.random((10, 10, 10, 10))
b = np.random.random((10, 10, 10, 10))

for einsum in [np.einsum, lib.einsum]:
    res = einsum("xijk,yijk->xy", a, b)
    print("%.32f" % lib.fp(res))
```

Thanks to @obackhouse.

### DIFF
--- a/pyscf/lib/numpy_helper.py
+++ b/pyscf/lib/numpy_helper.py
@@ -157,6 +157,7 @@ def _contract(subscripts, *tensors, **kwargs):
     idxBt = list(idxB)
     inner_shape = 1
     insert_B_loc = 0
+    shared_idxAB = sorted(list(shared_idxAB))
     for n in shared_idxAB:
         if rangeA[n] != rangeB[n]:
             err = ('ERROR: In index string %s, the range of index %s is '


fishjojo:
I didn't see difference before and after the change.

basilib:
> I didn't see difference before and after the change.

Did you turn off threading?

obackhouse:
> I didn't see difference before and after the change.

To expand: without the proposed change, `shared_idxAB` is not an ordered container, so for cases with a sufficient number of indices overlapping between the two tensors in the contraction, the indices are contracted over in a non-determined order. For the example that @basilib shared, there are a number of different possibilities i.e.:

```
{'k', 'i', 'j'}
-99.79633934060159106138598872348666
```

```
{'k', 'j', 'i'}
-99.79633934060144895283883670344949
```

If for whatever reason (chance or otherwise) your platform gets the same ordering in `shared_idxAB` between different runs, it will already perform deterministically - but it certainly doesn't on my platform. Of course, this difference is < machine precision and is unavoidable with threading - but shouldn't it really be deterministic in serial? Many of the core modules of PySCF i.e. `cc` depend upon `lib.einsum` and are therefore not performing deterministically even without threading.

zhcui:
I think Python's built-in dictionary and set are already ordered since Python3. What is your platform?

obackhouse:
> I think Python's built-in dictionary and set are already ordered since Python3. What is your platform?

I thought only for `dict`s in the sense that they keep the order of item insertion? I have tried on three platforms:

```
Python 3.7.0 | packaged by conda-forge | (default, Nov 12 2018, 20:15:55) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
```

```
Python 3.8.5 (default, Sep  4 2020, 07:30:14) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
```

```
Python 3.6.8 (default, Nov 16 2020, 16:55:22) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)] on linux
```

using the script:

```
s = "ijk"
for i in set(s):
    print(s)
```

which mimics the initialisation we're referring to. I have reproduced the indeterminism in `pyscf` on the first two of those platforms.

fishjojo:
> > I didn't see difference before and after the change.
> 
> To expand: without the proposed change, `shared_idxAB` is not an ordered container, so for cases with a sufficient number of indices overlapping between the two tensors in the contraction, the indices are contracted over in a non-determined order. For the example that @basilib shared, there are a number of different possibilities i.e.:
> 
> ```
> {'k', 'i', 'j'}
> -99.79633934060159106138598872348666
> ```
> 
> ```
> {'k', 'j', 'i'}
> -99.79633934060144895283883670344949
> ```
> 
> If for whatever reason (chance or otherwise) your platform gets the same ordering in `shared_idxAB` between different runs, it will already perform deterministically - but it certainly doesn't on my platform. Of course, this difference is < machine precision and is unavoidable with threading - but shouldn't it really be deterministic in serial? Many of the core modules of PySCF i.e. `cc` depend upon `lib.einsum` and are therefore not performing deterministically even without threading.

You are right. I think this PR is fine.

sunqm (approved):

PR #1253, opened 2022-05-12
Make spin_square work for a custom fcisolver in UCASCI/UCASSCF

hczhai:
The current implementation of ``fci.spin_op.spin_square`` restricts the ``fcisolver`` in ``mcscf.ucasci`` to be ``fci.direct_spin1`` (or its subclass). The following change will allow supporting a custom fcisolver (such as the non-spin-adapted DMRG solver) with UCASCI and UCASSCF.

### DIFF
--- a/pyscf/mcscf/ucasci.py
+++ b/pyscf/mcscf/ucasci.py
@@ -296,13 +296,13 @@ class UCASCI(casci.CASCI):
             ss_core = self._scf.spin_square(mocore, ovlp_ao)
             if isinstance(self.e_cas, (float, numpy.number)):
                 ss = fci.spin_op.spin_square(self.ci, self.ncas, self.nelecas,
-                                             mocas, ovlp_ao)
+                                             mocas, ovlp_ao, self.fcisolver)
                 log.note('UCASCI E = %.15g  E(CI) = %.15g  S^2 = %.7f',
                          self.e_tot, self.e_cas, ss[0]+ss_core[0])
             else:
                 for i, e in enumerate(self.e_cas):
                     ss = fci.spin_op.spin_square(self.ci[i], self.ncas, self.nelecas,
-                                                 mocas, ovlp_ao)
+                                                 mocas, ovlp_ao, self.fcisolver)
                     log.note('UCASCI root %d  E = %.15g  E(CI) = %.15g  S^2 = %.7f',
                              i, self.e_tot[i], e, ss[0]+ss_core[0])
         else:
--- a/pyscf/fci/spin_op.py
+++ b/pyscf/fci/spin_op.py
@@ -109,13 +109,17 @@ def spin_square_general(dm1a, dm1b, dm2aa, dm2ab, dm2bb, mo_coeff, ovlp=1):
     return ss, multip
 
 @lib.with_doc(spin_square_general.__doc__)
-def spin_square(fcivec, norb, nelec, mo_coeff=None, ovlp=1):
-    from pyscf.fci import direct_spin1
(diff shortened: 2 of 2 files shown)


pyscf/mcscf/ucasci.py
```diff
@@ -296,13 +296,13 @@ def _finalize(self):
             ss_core = self._scf.spin_square(mocore, ovlp_ao)
             if isinstance(self.e_cas, (float, numpy.number)):
                 ss = fci.spin_op.spin_square(self.ci, self.ncas, self.nelecas,
-                                             mocas, ovlp_ao)
+                                             mocas, ovlp_ao, self.fcisolver)
```
sunqm, line 299: How about calling `self.fcisolver.spin_square()` here?

pyscf/mcscf/ucasci.py
```diff
@@ -296,13 +296,13 @@ def _finalize(self):
...
                 log.note('UCASCI E = %.15g  E(CI) = %.15g  S^2 = %.7f',
                          self.e_tot, self.e_cas, ss[0]+ss_core[0])
             else:
                 for i, e in enumerate(self.e_cas):
                     ss = fci.spin_op.spin_square(self.ci[i], self.ncas, self.nelecas,
-                                                 mocas, ovlp_ao)
+                                                 mocas, ovlp_ao, self.fcisolver)
```
sunqm, line 305: `self.fcisolver.states_spin_square` ?

pyscf/mcscf/ucasci.py
hczhai: I think calling ``self.fcisolver.spin_square()`` will break the original code even when the ``self.fcisolver`` is a normal ``direct_spin1`` FCI solver. The reasons are the following:

(1) Near the calling position of ``fci.spin_op.spin_square`` there is an ``if`` statement: https://github.com/pyscf/pyscf/blob/master/pyscf/mcscf/ucasci.py#L288 and the ``if`` checks that there is indeed a ``spin_square`` method. But then the original code **deliberately** does not call that ``spin_square`` method.

(2) We can also check the implementation of ``self.fcisolver.spin_square()`` in ``direct_spin1`` FCI solver: https://github.com/pyscf/pyscf/blob/master/pyscf/fci/direct_spin1.py#L829-L832 and it calls ``fci.spin_op.spin_square0`` and the documentation for it says that ``Spin square for RHF-FCI CI wfn only``. So it may indicate that calling ``self.fcisolver.spin_square()`` in UCASCI will not give the correct behavior.

pyscf/mcscf/ucasci.py
hczhai: @sunqm As explained in the May 16 reply, calling ``self.fcisolver.spin_square()`` or ``self.fcisolver. states_spin_square()`` will break the existing UCASCI code, this is why in my pull request I choose to keep ``fci.spin_op.spin_square`` but this method needs to get the 1/2rdms from the correct fcisolver. In addition, the computation of $\langle S^2 \rangle$ from 1/2rdms is general and independent from ``fcisolver`` implementation, which is what currently implemented in ``fci.spin_op.spin_square`` (See https://github.com/pyscf/pyscf/blob/master/pyscf/fci/spin_op.py#L112-L120). For DMRG fcisolver we can indeed specialize the computation of $\langle S^2 \rangle$ without computing and storing 1 and 2 rdms, which may be slightly faster, but the default FCI solver in pyscf does not have this specialization.

The error in the github action checks (below) is not relevant to this pull request, which should have been fixed at some time after I created this pull request. Someone should trigger the rerun of these checks.

sunqm (approved):

PR #1255, opened 2022-05-14
bug fix for LRC hybrid DFT with exxdiv = 'ewald'

hongzhouye:
For a STO-3G H2 molecule in a cubic box of length `a0`, the PBC DFT energy with `exxdiv='ewald'` for the wb97x LRC hybrid functional does not converge to that from a molecular calculation (see column `old Epbc` vs `Emol`).
| a0 (Ang)  |  old Epbc  |   new Epbc  |   Emol   |
| ----------|-----------|-------------|--------|
| 10             | -0.87429950  |  -1.15942760 |  |
| 20             | -0.87299382 | -1.15812192   |  |
| 1/a0^3 extrap | -0.87280729 | -1.15793540  | -1.15793625 |

The problem is found to be the `madelung` function in `pbc/tool/pbc.py` giving an incorrect correction to the long-range exchange matrix. The change in this PR fixes this (see column `new Epbc` in the table above).

A test function `test_rsh_0d_ewald` is added to `pbc/dft/test_rsk.py`. The reference values for other tests in the same file that use a LRC hybrid functional and `exxdiv='ewald'` are updated as well (the new reference values are much closer to the molecular result in all cases).

Thanks to @tberkel and @verena-neufeld for helping identify the problem and the fix.

### DIFF
--- a/pyscf/pbc/df/test/test_mdf_jk.py
+++ b/pyscf/pbc/df/test/test_mdf_jk.py
@@ -309,7 +309,7 @@ class KnownValues(unittest.TestCase):
         vj2, vk2 = pbcdf.MDF(cell, kpts).get_jk(dm, hermi=0, kpts=kpts, omega=0.3, exxdiv='ewald')
         vj3, vk3 = pbcdf.AFTDF(cell, kpts).get_jk(dm, hermi=0, kpts=kpts, omega=0.3, exxdiv='ewald')
         self.assertAlmostEqual(lib.fp(vj0), 0.007500219791944259, 9)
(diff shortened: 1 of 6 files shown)


sunqm (approved):

verena-neufeld:
@sunqm , do you think the test might pass once PR #1249 is merged? Thanks so much for your help!

sunqm:
> (just out of interest:) @sunqm , do you think the test might pass once PR #1249 is merged? Thanks so much for your help!

Yes. It seems the default macos-latest virtual machine upgraded in github actions. PR 1249 was built on top of the new macos vm. There is another crash I'm investigating. But it's not the same failure in this test pipeline

PR #1263, opened 2022-05-27
take Ls and rcut as kwargs to pbc.gto.eval_gto

willwheelera:
For certain systems, generating Ls and rcut take a significant amount of time. PyQMC uses eval_gto for orbital evaluation; we can spare this extra time by computing Ls and rcut once in advance and passing them as kwargs into eval_gto. 

I don't think this change should affect anything else.


sunqm (approved):

tberkel:
Just an idea: in order to not burden the user and to facilitate the re-use
of this feature elsewhere, would it make sense to store Ls and rcut as
(hidden?) attributes of cell?

PR #1276, opened 2022-06-03
New Module for Born-Oppenheimer Molecular Dynamics (BOMD)

matthew-hennefarth:
This PR adds a new module (md) to perform molecular dynamics. It utilizes the built in scanner methods to compute the forces on the atoms and using the Velocity Verlet algorithms propagates the nuclear coordinates in time. Currently, the PR only introduces the NVE ensemble.

The general layout of the code is as follows. I have introduced a base Integrator class which handles all aspects that a generic integrator needs (kernel method, loops, writing trajectory/energy etc). Any specific type of integrator (velocity-verlet NVE or velocity-verlet + Nose-Hoover etc) are derived from this base class and must override the _next() function. The _next() function returns the next Frame of the simulation. I take advantage of the built in __iter__ and __next__ functions to progress the simulation similar to the geometry optimizers.

As of now, the integrator class looks at the incore_anyway variable to determine if it will save all of the Frames of the simulation in memory. Additionally, setting trajectory_output and energy_output (similar to the generic output in the gto.Mole object) will have the simulation write the energy and structure (as xyz) to disk after each iteration. 

I have included 2 examples as well as unit tests for the code. Additionally, I have submitted a PR for the docs page to include the necessary documentation ([here](https://github.com/pyscf/pyscf.github.io/pull/95#issue-1260255464)) Lastly, this PR should not affect any other modules in the PySCF repo (it does introduce the COMMON_ISOTOPE_MASSES list in the data/elements.py file to be consistent with OpenMolcas and SHARC, but again should not affect anything). 

Lastly, I have extensively tested the simulations against BOMD Dynamixs package in OpenMolcas and have complete agreement up to numerical convergence of the underlying electronic structure method.


jamesETsmith:
Thanks for this @matthew-hennefarth, this is a great addition to PySCF! It might be better to do this in a follow-up PR, but is there an easy way to add a `callback` function that executes at each MD iteration? For example, users could write the dipole (or any other property) to a new file at each step.

matthew-hennefarth:
It should be relatively simple to add! I was thinking of that already, but wanted to get the general structure of the code under review first. 

Essentially, adding a new attribute, callback, to Integrator and then calling the callback() in the __next__, _next, or the kernel function (not sure where the best location would be but I would make it more consistent with the geometry optimizers).

jamesETsmith:
It's in the `calc_new()` function, so I think it'd make sense to put it in the `next()` function for you. I think it's definitely a good idea to get the initial PR out there and we can expand more fancy features later.

pyscf/md/test/test_velocverlet.py
```diff
@@ -0,0 +1,128 @@
...
+casscf_scanner = hf.CASSCF(ncas, nelecas)
+casscf_scanner.conv_tol = 1e-12
+casscf_scanner.conv_tol_grad = 1e-6
+
+
+def tearDownModule():
+    global h2o, hf_scanner, ethylene, hf, casscf_scanner
+    h2o.stdout.close()
+    ethylene.stdout.close()
+    hf_scanner.stdout.close()
```
sunqm, line 58: It may close the same file twice. You probably don't need to close `hf_scanner.stdout` and `casscf_scanner.stddout`.

pyscf/md/test/test_velocverlet.py
```diff
@@ -0,0 +1,128 @@
...
+# limitations under the License.
+
+import unittest
+import numpy as np
+from pyscf import gto, scf
+import pyscf.md as md
+
+CHECK_STABILITY = False
+
+h2o = gto.M(verbose=3,
```
sunqm, line 23: Initialization can be put in the `setUpModule` function. The default test framework is pytest now. With pytest the module level initialization may hit resources limits in the github CI pipelines.

pyscf/md/test/test_velocverlet.py
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env python
+# Copyright 2014-2018 The PySCF Developers. All Rights Reserved.
```
sunqm, line 2: Copyright 2022 The PySCF Developers. ...

pyscf/md/integrators.py
```diff
@@ -0,0 +1,459 @@
+#!/usr/bin/env python
+# Copyright 2014-2018 The PySCF Developers. All Rights Reserved.
```
sunqm, line 2: Copyright 2022 The PySCF Developers. ...

pyscf/md/__init__.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
+# Copyright 2014-2018 The PySCF Developers. All Rights Reserved.
```
sunqm, line 2: Copyright 2022 The PySCF Developers. ...

sunqm (approved):

pyscf/md/test/test_velocverlet.py
matthew-hennefarth: Pushed the setup into the `setUpModule` function.

matthew-hennefarth:
New commits address comments from @sunqm.

PR #1283, opened 2022-06-09
Adds copy argument to cisdvec_to_amplitudes

Majokux:
`cisdvec_to_amplitudes` in the CISD, UCISD, and GCISD return the wave function in terms of C0, C1, and C2 amplitudes.

Modifying the returned arrays has inconsistent effects:
C0 is always a float and as such immutable
C1 is a view into underlying `civec` - modifying the returned C1 amplitudes will also modify the underlying `civec`
C2 can be a view or an unpacked new array, so modifying these amplitudes may or may not have side effects

I think `cisdvec_to_amplitudes` should always return copies by default, which can be modified without side effects.
I thus added a keyword argument `copy` with default value `True`. Existing code, which does not modify the amplitudes, can be called with `copy=False`, to avoid unnecessary copying.

### DIFF
--- a/pyscf/cc/eom_gccsd.py
+++ b/pyscf/cc/eom_gccsd.py
@@ -240,7 +240,7 @@ class EOMIP(eom_rccsd.EOMIP):
     def vector_size(self):
         nocc = self.nocc
         nvir = self.nmo - nocc
-        return nocc + nocc*(nocc-1)/2*nvir
+        return nocc + nocc*(nocc-1)//2*nvir
 
     def make_imds(self, eris=None):
         imds = _IMDS(self._cc, eris)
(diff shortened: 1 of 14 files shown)


tberkel:
Seems wise, thanks Max. CC has similar functions to convert between
amplitudes and vectors, do they need modifications too?

Majokux:
Hi Tim,
I think these are safe as they have `copy()`s or preallocate the memory for the vector.
A quick test:
```
t1orig = t1.copy()
t2orig = t2.copy()
vec = mycc.amplitudes_to_vector(t1, t2)
vec_orig = vec.copy()

# Test side effect in vector_to_amplitudes
t1, t2 = mycc.vector_to_amplitudes(vec)
t1[:] = 0
t2[:] = 0
print(np.linalg.norm(vec - vec_orig))   # 0.0

# Test side effect in amplitudes_to_vector
t1, t2 = t1orig.copy(), t2orig.copy()
vec = mycc.amplitudes_to_vector(t1, t2)
vec[:] = 0
print(np.linalg.norm(t1 - t1orig))   # 0.0
print(np.linalg.norm(t2 - t2orig))   # 0.0
```
returned 0 everywhere

sunqm (approved):

sunqm:
@Majokux I added a few tests and create an PR to your branch. If that looks no problem please include my tests in this PR

Majokux:
@sunqm  That's great, thanks

PR #1286, opened 2022-06-09
Exposes with_frozen, with_dm1, with_mf kws in cc.make_rdm

Majokux:
The make_rdm routines in the CCSD module have keyword arguments `with_frozen` (1DM and 2DM), `with_mf` (1DM) and `with_dm1` (2DM), which are however not exposed all the way to the method level.
I use non-default options for these rather frequently and think it would be convenient to allow passing them when calling
the make_rdm routines as methods. Especially `with_frozen=False` can be very useful for 2-DMs, as the full 2-DM with frozen parts becomes quickly very large.

The defaults are all set such that the current default behaviour is preserved.


sunqm (approved):

Majokux:
Any chance this could get merged soon? We would like to use these keyword arguments in some code downstream.

PR #1288, opened 2022-06-09
Linear dependency treatment for IAOs

Majokux:
I'm frequently using IAOs and encountered two problems with the current code:

1) For large basis sets, where the AO overlap is (numerically) singular, the Cholesky decomposition of this overlap matrix will fail; this is especially common in solids. In cases where the CD fails, I added a backup inversion, a la canonical orthogonalization, i.e. removing (almost) linearly dependent eigenvectors according to some threshold `lindep_threshold`
[I actually think options for linear-dependency treatment, such as method and threshold, should be attached to the `mole`/`cell` object, as it crops up in multiple locations, but this is another discussion].
 
 2) If at some k-point (or in one spin channel, for a 1e-system), there are not occupied orbitals, the call to `vec_lowdin` will fail
 (there are no orbitals to orthogonalize). I added an early return to deal with this case.
 
Lastly, the IAO construction for molecules and for solids at a given k-point is the same; previously the code was just copied,
 with a k-point loop in the case for solids, but to reduce code duplication I put it into a function, which is being looped over for k-points.

### DIFF
--- a/pyscf/lo/iao.py
+++ b/pyscf/lo/iao.py
@@ -37,14 +37,23 @@ from pyscf.data.elements import is_ghost_atom
 #     vec_lowdin(iao_coeff, mol.intor('int1e_ovlp'))
 MINAO = getattr(__config__, 'lo_iao_minao', 'minao')
 
-def iao(mol, orbocc, minao=MINAO, kpts=None):
+def iao(mol, orbocc, minao=MINAO, kpts=None, lindep_threshold=1e-8):
     '''Intrinsic Atomic Orbitals. [Ref. JCTC, 9, 4834]
 
-    Args:
-        mol : the molecule or cell object
+    For large basis sets which are close to being linearly dependent,
+    the Cholesky decomposition can fail. In this case a canonical orthogonalization
+    with threshold `lindep_threshold` is used.
 
+    Args:
+        mol : molecule or cell object
         orbocc : 2D array
             occupied orbitals
+        minao : str, optional
+            reference basis set for IAOs
+        kpts : 2D ndarray, optional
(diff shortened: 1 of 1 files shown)


sunqm (approved):

susilehtola:
Does it make sense to duplicate the code in scf.addons?

Majokux:
>Does it make sense to duplicate the code in scf.addons?
Sorry, which duplicated code to you mean?

susilehtola:
> > Does it make sense to duplicate the code in scf.addons?
> 
> Sorry, which duplicated code to you mean?

Handling of linear dependencies. Canonical orthogonalization is implemented in `scf.addons.canonical_orth_`, but the addons also have more robust methods i.e. partial Cholesky orthonormalization.

My biggest problem with PySCF is that linear dependencies are an issue in many routines, especially RI based ones as the `(I|J)` matrix is often less well conditioned than the AO overlap matrix!

Majokux:
@susilehtola Makes sense, I changed the PR to use `scf.addons.canonical_orth_` instead.

zhcui (approved):

zhcui:
should `x.T` in line `p12 = numpy.linalg.multi_dot((x, x.T, s12))` be` x.conj().T`?

Majokux:
>should x.T in line p12 = numpy.linalg.multi_dot((x, x.T, s12)) be x.conj().T?
Yes indeed, thanks bringing this up

susilehtola:
> should `x.T` in line `p12 = numpy.linalg.multi_dot((x, x.T, s12))` be` x.conj().T`?

or just `x.H`

PR #1308, opened 2022-06-28
BOMD Cleanup and Maxwell-Boltzmann Velocity Distribution

matthew-hennefarth:
This PR cleans up the Velocity-Verlet integrator to take advantage of numpy's vector arithmetic and remove explicit for loops. Other slight changes include caching the mass of each nuclei in atomic units so that a constant lookup isn't needed at each step. When writing out the energy, it will check to see if it is a multi-state scanner type and if so, print out the energy of all states at the end of the line, similar to OpenMolcas. It will also flush these files after each time step so that if the calculation crashes, not all data is lost. It also changes the integrator base class to inherit the stream class to be able to chain commands together in a single line. 

Implementation of a Maxwell-Boltzmann velocity distribution function to sample initial velocities of MD simulations. Users can specify the temperature at which to sample. Also, a global `rng` is provided in `md/__init__.py` where users can set the global seed for reproducible calculations. Examples and a simple statistical test is provided.


matthew-hennefarth:
@fishjojo I was just wondering if there is anything that I need to do for this PR to be accepted (it has been sitting here for quite some time now). Let me know!

fishjojo (approved):

PR #1323, opened 2022-07-13
Skip logging S^2 in mcscf with 64 orbitals or more

AllanChain:
Currently, systems with 64 orbitals or more are not fully supported in FCI solvers:

https://github.com/pyscf/pyscf/blob/1de8c145abb3e1a7392df9118e8062e6fe6bde00/pyscf/fci/cistring.py#L255-L256

Thus `NotImplementedError` will be raised when trying to log spin square, prohibiting further executions:

https://github.com/pyscf/pyscf/blob/f985dde7358409e4e71095575172327af795a6cb/pyscf/mcscf/casci.py#L974-L978

It is possible to get around this by specifying a lower logging level. But that's not elegant.

This PR wraps those calls with `try ... except` and skips logging S^2 with 64 orbitals or more.


sunqm:
The energy should be printed though the S^2 calculations are skipped.

sunqm (approved):

PR #1342, opened 2022-07-23
an example script for renormalized perturbative triples correction on CCSD

seunghoonlee89:
an example script for renormalized perturbative triples correction on CCSD (R-CCSD(T)).
It is a simplest correction of CCSD(T) for strongly correlated systems.

It's detailed formulation can be found in Comput. Phys. Commun. 149, 71 (2002); https://doi.org/10.1016/S0010-4655(02)00598-2.
and a recent application can be found in JCTC 17, 3414 (2021); https://doi.org/10.1021/acs.jctc.1c00205.

### DIFF
--- a/examples/cc/25-renormalized-ccsd_t.py
+++ b/examples/cc/25-renormalized-ccsd_t.py
@@ -1,33 +1,16 @@
 #!/usr/bin/env python
-# Copyright 2014-2020 The PySCF Developers. All Rights Reserved.
-#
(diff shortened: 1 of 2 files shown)


sunqm (approved):

tberkel:
With added support for UHF references, it would be great to make this a
native method, rather than an example script.

seunghoonlee89:
I will make a new commit for UHF and GHF references later.

PR #1350, opened 2022-07-26
Fixes inconsistent energy between k-RCCSD and k-UCCSD

maxnus:
There is an inconsistency in energy between k-RCCSD and k-UCCSD, which is due to the divergent exact exchange correction of the mean field energy. More generally, I think this will also affect k-RCCSD calculation on top of non-converged HF mean-fields, such as LDA - the main issue is that the k-RCCSD kernel, opposed to all other CCSD kernels, does *not* set the attribute `e_hf`,
and the `e_tot` property thus falls back to using `self._scf.e_tot` as the mean-field energy.

The inconsistency between k-RCCSD and k-UCCSD then arises hence `self._scf.e_tot` by default includes the exxdiv correction,
whereas `eris.e_hf` does not, since the HF potential is build without this correction in the `_common_init_` of the ERIs.
For k-RCCSD@LDA the issue will likely be even worse, since E_LDA will be very different from E_HF[DM_LDA]. 

This fix will make ensure that k-RCCSD behaves like k-UCCSD (and also agrees with the Gamma-point pbc.ccsd classes, for Gamma-point only calculations. The code I used to verify this is attached below.

There is still another, related question, which is if we should not use the exxdiv correction for the HF energy part of the CCSD total energy by default (but not for the `eris.fock`, used in the CCSD eqs. and correlation energy, of course).
I think there is some justification for this, as exxdiv accelerates convergence to the thermodynamic limit, and combining a HF energy with exxdiv with a CCSD correlation energy without exxdiv is similar to combining, say, a E(HF) on a 6x6x6 k-point grid with a E(corr) calculated on a 3x3x3 grid - a somewhat common approach to converge energetics.
One big benefit of this would be that `ccsd.e_tot = hf.e_tot + ccsd.e_corr` will be true again (for converged HF calculations), as most users would expect.

[example.py.txt](https://github.com/pyscf/pyscf/files/9193487/example.py.txt)

### DIFF
--- a/pyscf/cc/ccsd.py
+++ b/pyscf/cc/ccsd.py
@@ -979,7 +979,7 @@ http://sunqm.net/pyscf/code-rule.html#api-rules for the details of API conventio
 
     @property
     def e_tot(self):
-        return (self.e_hf or self._scf.e_tot) + self.e_corr
+        return self.e_hf + self.e_corr
 
     @property
     def nocc(self):
--- a/pyscf/pbc/cc/kccsd_rhf.py
+++ b/pyscf/pbc/cc/kccsd_rhf.py
@@ -608,6 +608,11 @@ class RCCSD(pyscf.cc.ccsd.CCSD):
             # eris = self.ao2mo()
             eris = self.ao2mo(self.mo_coeff)
         self.eris = eris
+
+        self.e_hf = getattr(eris, 'e_hf', None)
+        if self.e_hf is None:
+            self.e_hf = self._scf.e_tot
+
         if mbpt2:
             self.e_corr, self.t1, self.t2 = self.init_amps(eris)
             return self.e_corr, self.t1, self.t2


verena-neufeld:
Hi Max, thanks for pointing this out. In fact, I think I had noticed this also but sadly didn't address it further as you just did. I personally would prefer keeping the exxdiv affected HF energy in the total CCSD energy, i.e. changing the behaviour of KUCCSD rather than KRCCSD. I prefer if the total CCSD energy is the sum of HF energy and CCSD corr energy pyscf gives me along the way when calculating and therefore I prefer the behaviour of KRCCSD over KUCCSD, i.e. I think total CCSD energy should differ when I do exxdiv=None or when I do exxdiv="ewald".

Majokux:
>  I personally would prefer keeping the exxdiv affected HF energy in the total CCSD energy, i.e. changing the behaviour of KUCCSD rather than KRCCSD

Yes that would be my preference as well - unfortunately, the way that k-RCCSD "achieved" this (by using `cc._scf.e_tot`) is not very robust at all

verena-neufeld:
I had a closer look to follow what you said. It seems like the issue arises from the fact that `pbc/cc/rccsd_rhf.py` does not call  the `ccsd` function in `cc/ccsd/ccsd.py` as for example `pbc/cc/rccsd_uhf.py` does (via a wrapper) but instead has its own `ccsd` function. This can probably be improved I think to keep the code cleaner (i.e. make `pbc/cc/rccsd_rhf.py` call the `ccsd` function in `cc/ccsd/ccsd.py`, too, via a wrapper). Why do you think using modifying `cc/ccsd/ccsd.py` to use `cc._scf.e_tot` immediately is not a good idea (instead of trying to get `e_hf` from `eris.e_hf` first)? A solution might be to modify the `ccsd` function in `cc/ccsd/ccsd.py` to not check `e_hf` from `eris.e_hf` first but I haven't checked what other implications this might have. The deviation in behaviour is of course, as you said, that `eris.e_hf` does not contain the effects of `exxdiv` unless the user set `keep_exxdiv` to True.

verena-neufeld:
I checked and this usage of `eris.e_hf` first before trying `cc._scf.e_tot` was introduced in commit `db64a063f42fee15cbb7dc1b348c2ded88291387` to deal with cases where the SCF was not converged. In `eris`, the HF energy was recomputed, so this non convergence issue was solved. However, in pbc cases, `eris` often ignores `exxdiv` and therefore its recomputed HF energy is probably not be used as the HF energy. Maybe there should be a function that recomputes the HF energy refactored out, in the `eris` class this function can be called with `exxdiv` removed by default, and in the main CCSD kernels, this class can be called with `exxdiv` left in.

maxnus:
Hi Verena,

I think not/never using `cc._scf.e_tot` is the right call - it's not just about non-converged HF, but also DFT, where `e_tot` is the KS energy.

I also don't really think this mean-field energy belongs into the `_ChemistERIs` object - I would suggest to add a function to the CCSD class itself, which returns Hartree-Fock potential and energy (based on the density in `cc._scf.make_rdm1()`)
The ERIs can then also use this HF potential in `_common_init_`, as needed since they also store the (active-space part of the) Fock matrix.

It's also easy to go from a HF potential with exxdiv to one without, you just need to shift the diagonal of the occupied block by a factor of the Madelung constant (at least that's true for `'ewald'`), so we can use the potential with exxdiv to calculate the energy, and then remove it (unless `keep_exxdiv == True`) to pass it to the ERIs.

verena-neufeld:
I agree, I think a function in the CCSD class, or maybe actually in the HF class (?) so it is accessible for other post HF methods, to get the HF energy (containing vhf, dm and finally e_tot evaluation) is a good idea, with `lib.temporary_env` appropriately placed to remove the effect of `exxdiv` if required. This can be called from the `eris` class (`exxdiv` effect usually removed), as well as the `ccsd` classes (`exxdiv` effect not removed) for example. I think we shouldn't use the madelung constant to move from removed `exxdiv` to applied `exxdiv`, if we can avoid it, as non ewald `exxdiv` don't use the madelung constant.

sunqm (approved):

verena-neufeld:
Haha seems like this was merged was faster than our conversation :D . I would prefer if the exxdiv stays in e_tot in CCSD, so I guess I'll make a new PR which does the above mentioned refactoring and then restored the exxdiv behaviour for KRCCSD, but also brings it to KUCCSD, etc :)

maxnus:
I think merging this is good, the exxdiv was already removed in Gamma-point and KUCCSD, as you said, now it's consistent (and for HF mean-fields you can relatively simply use `hf.e_tot + cc.e_corr` to get the exxdiv corrected energy.

But yes, I'm all for making the change to an exxdiv corrected `cc.e_tot`.

sunqm:
I guess I misunderstood the issue. When looking at cc.e_tot, I guess most people would think it is `e_hf + e_tot` . So probably the exxdiv should be consistently contained in `cc.e_tot` since it is contained in scf total energy by defaults. Otherwise it may get more confusion that HF energy is higher than `cc.e_tot`.

maxnus:
@sunqm Yes, I agree. I think @verena-neufeld will make a PR to get this behavior.

tberkel:
Without adding any suggestion to refactoring, I want to add that I *think*
there is now consensus that the CC correlation energy should never be
computing with orbital energies containing an exxdiv correction. If others
agree, then any cleanup of the code shouldn't even allow other choices
(level_shift exists if needed). However, this has an annoying artifact: the
initial CC energy won't match the "correct" MP2 correlation energy (which I
think people agree should be calculated with orbital energies that *do*
contain an exxdiv correction). I suppose we could make the initial t2 guess
include some kind of exxdiv. Also, I imagine a similar issue pops up for
CCSD(T).

maxnus:
Hi Tim,
I agree that this is an issue and should be looked at. I have a few questions:
>I want to add that I *think*
there is now consensus that the CC correlation energy should never be
computing with orbital energies containing an exxdiv correction

When you say "orbital energies" do you mean the `Fock[occ,vir] * T1` contribution to Ecorr(CCSD) for a non-HF reference?
For this term I assumed it would not matter if done with or without exxdiv, since the latter only affects the occupied-occupied block, i.e. only modifies the mean-field energy itself?

There is of course a difference when it comes to which Fock matrix is used for the T-amplitude updates; have you found that this should better be the uncorrected Fock matrix (as it is in the code right now)?

> However, this has an annoying artifact: the
initial CC energy won't match the "correct" MP2 correlation energy (which I
think people agree should be calculated with orbital energies that *do*
contain an exxdiv correction). I suppose we could make the initial t2 guess
include some kind of exxdiv

I actually thought the MP2 intial guess *does* include the exxdiv correction, since it uses `eris.mo_energy` (rather than `eris.fock`) to construct the amplitudes and the latter should be exxdiv corrected? This might be only true for the `'ewald'` correction though.

tberkel:
Hi Max,

Yes, I mean the second part (i.e., "which" Fock matrix, and whether Foo has
corrections) and indeed believe that the "uncorrected" one is appropriate
to use in CC. Roughly, if you consistently track the appearance of all
leading-order G=0 divergences, e.g., in ERIs like (pp|qq), then, in
addition to those appearing in the Fock eigenvalues, there are also a set
of corrections that appear in the T amplitude updates, and it turns out
that they all cancel. So whether you correct for this *everywhere* or
*nowhere*, you get the same answer; correcting only in the Fock eigenvalues
is inconsistent. I have a rough set of notes I'm happy to send you
separately if you're interested.

Regarding MP2, yes, I think you're right -- it's been a while since I've
looked at the code! It looks like everything behaves the way I would
suggest (with the `ewald` type correction playing two roles, i.e. to make
the initial guess the correct MP2 one and to improve convergence as a kind
of modified preconditioner/level shift).

maxnus:
>Yes, I mean the second part (i.e., "which" Fock matrix, and whether Foo has
corrections) and indeed believe that the "uncorrected" one is appropriate
to use in CC. Roughly, if you consistently track the appearance of all
leading-order G=0 divergences, e.g., in ERIs like (pp|qq), then, in
addition to those appearing in the Fock eigenvalues, there are also a set
of corrections that appear in the T amplitude updates, and it turns out
that they all cancel. So whether you correct for this *everywhere* or
*nowhere*, you get the same answer; correcting only in the Fock eigenvalues
is inconsistent. I have a rough set of notes I'm happy to send you
separately if you're interested.

This is quite interesting. If you say correcting *everywhere* and *nowhere* get's the same answer, why do you favor the uncorrected approach? It seem's that the corrected approach might be more in-line with the way we do MP2? 

There is also the dependence on the HF orbitals, rather than just the orbital energies:
with correct *nowhere*, do you mean that the HF itself is converged without an exxdiv correction?

tberkel:
> This is quite interesting. If you say correcting everywhere and nowhere get's the same answer, why do you favor the uncorrected approach? It seem's that the corrected approach might be more in-line with the way we do MP2.

> [edit: It came to me that "correcting everywhere" probably requires changes in additional terms in the CCSD amplitude update, not currently implemented in PySCF...]

Yes, exactly (the edit). If you get the same answer, then might as well use the simpler version. 😅 

> There is also the dependence on the HF orbitals, rather than just the orbital energies:
with correct nowhere, do you mean that the HF itself is converged without an exxdiv correction?

Right, I'm strictly assuming HF orbitals throughout, and note that the HF orbitals are identical with `exxdiv=None` and `exxdiv='ewald'` (but not for other `exxdiv`). So when I say "correct nowhere", I mean you have a converged HF solution with either of those two options, such that the orbitals are identical (but the HF total energy and HF occupied orbital energies are different by terms proportional to the Madelung constant).

Issue #1330, opened 2022-07-20
Save iteration count (cycle) to CCSD object


kfyharukz:
Hi,
In the current codes, the iteration count of CCSD isn't saved; however, it's useful, for example, to judge how complex the target problem is.
I think it's not difficult and can be enabled by small changes [like this code](https://github.com/kfyharukz/pyscf/compare/master...save-cycle-count).

So, can I send a PR?

Majokux:
I personally think it's a good idea and it also applies to all other iterative methods, SCF, CASSCF, etc

However, I think it would be good to bundle such results up a bit more to avoid too many return values and stateful output attributes in the calculation object.
This could be done similar to the [OptimizeResult](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.OptimizeResult.html#scipy.optimize.OptimizeResult) object in SciPy; I would suggest somthing akin to
```
class IterationResults(typing.NamedTuple):
    converged: bool = False
    cycle: int = 0
    status: Optional[str] = None
```
which can be instantiated on CCSD/SCF/....`.__init__` and updated in the respective kernel.

kfyharukz:
It seems good.  I'll prepare PR that way. 😃

sunqm:
Using PR #1351, you can now access the .cycles attribute of the CCSD object to obtain the iteration count.

PR #1351, opened 2022-07-27
Save cycle count

kfyharukz:
ref. #1330 

This PR enables to save iteration cycle count to the calculation object.

I implemented it based on the feedback in #1330, once on CCSD.
It looks clear but is breaking change for the interface of the calculation object.  What do you think about making destructive changes? @Majokux @hebrewsnabla


Majokux:
Thanks for opening this.
I would suggest some modifications:
* Instead of `itr_info`, I would suggest to name the attribute `iterinfo`
* We definitely still want to have access to `ccsd.converged` - but we do not want to store it as an independent boolean attribute, so we can add it as a property to the CCSD class:
```
@property
def converged(self):
   return self.iterinfo.converged
```
* The kernel should either return a new `IterInfo` object, or update the existing one by accessing `self`. The later approach would make it a non-pure function (function with side effects), although I believe this is already the case because of the DIIS object (?).
* It's not necesary to have both `cycle` and `istep` in the current kernel - you can simply use the value of `istep+1` after the `for`-loop to initialize the `IterInfo` object.

It would be good to hear the opinion of one of the core developers if this effort to save the iteration count is a good idea.

kfyharukz:
@Majokux  Thanks for your useful comments! I rewrote the PR reflecting them.

> It would be good to hear the opinion of one of the core developers if this effort to save the iteration count is a good idea.

Do you know who to talk to?

sunqm (commented):
It is acceptable to update the iterinfo in the kernel function. Changing the returns of kernel function may break someone's code (for example pyscf/cc/rccsd_slow.py), although it does not break other modules in pyscf.

It is not necessary to create a new module to provide the IterationInfo class. This class can be defined in cc module.

verena-neufeld:
Thanks! I think this is a good idea for storing meta data about the convergence and iterations, etc, of CCSD, SCF, etc! I personally like having this class abstracted out outside of CCSD, so that it could also be used for SCF etc. I have two comments:
1) I think the cycle attribute can be updated at every step, so that if the calculation gets stopped earlier in a clean way, the numbers of cycles that were done can be accessed.
2) I am in favour of not changing the output of kernel but instead the user can access cc.iterinfo (for backwards compatibility).

kfyharukz:
>I think this is a good idea for storing meta data about the convergence and iterations, etc, of CCSD, SCF, etc! I personally like having this class abstracted out outside of CCSD, so that it could also be used for SCF etc.

Yes, it's why I created new module.

I've rewritten my PR to ensure the backwards compatibility kernel func,  and also counting up cycle variable in `for` loop.

sunqm (approved):

sunqm (approved):

sunqm:
I’ve simplified the cycle count in CCSD by removing the `IterInfo` class. Now, you can access the cycle count directly via `CCSD(mf).run().cycles`. As for the IterInformation class, I believe it could be made more general to store additional information like cycles, convergence status, energy, etc. In this case, such a summary or metadata class would be like the scf_summary attribute in the SCF class. If you’re interested in pursuing this refactoring/design, please feel free to open a new issue so we can discuss it further.

PR #1357, opened 2022-08-05
evaluate HF energy in CCSD and MP2 using mo occ and mo coeff

verena-neufeld:
This reevaluates the HF energy as a property in the CCSD and MP2 calculations using MO coeffs and MO occ. This is useful for the evaluation of the total CCSD/MP2 energy. `e_hf` is now no longer an attribute of `eris`.

- First it tries to use the coeffs and occ from the mean field object, if they are not set, it uses the CCSD/MP2 coeffs and occ.
- The code is duplicated, it appears twice, once in cc/ccsd.py, the other time in mp/mp2.py. I wonder whether I could change that.
- It is now necessary to specify `get_veff()` and `make_rdm1()` for the mf object, otherwise the subsequent CCSD/MP2 calculation will fail in the HF energy evaluation; see changes in test `mp/tests/mp2.py` and others.

This will affects the total energies of some k points post HF methods, e.g. KUCCSD, when exxdiv is not None. While previously, some effects of exxdiv where not present in the total KUCCSD energy (e.g. E_KUCCSD != E_HF + E_KUCCSD_corr), now E_KUCCSD = E_HF + E_KUCCSD_corr. See discussion in PR #1350 . 

Before merging, please let me know whether the style is good and whether there are some edge cases I can test that I might have forgotten, that are important. Obviously, the short code snippet to evaluate the HF energy was adapted from other parts of pyscf that evaluate HF energies.


hebrewsnabla:
I understand that it makes sense to re-evaluate the `e_hf` of periodic CC/MP2. But for what reason this PR need to apply the same thing for non-periodic CC/MP2 (instead of using `self.e_hf = self._scf.e_tot`)?

maxnus:
This looks good.

A few questions from my side:

>First it tries to use the coeffs and occ from the mean field object, if they are not set, it uses the CCSD/MP2 coeffs and occ.

Why this choice, instead of using the CC's `mo_coeff` first? Can it really happen that mf doesn't have `mo_coeff` and `mo_occ` defined?

>The code is duplicated, it appears twice, once in cc/ccsd.py, the other time in mp/mp2.py. I wonder whether I could change that.

I think it's worth avoiding that, as is already done for some of the `get_frozen_mask` etc between MP2 and CCSD. You don't have to make it a module function to reuse the code, you can simply have `e_mf = MP2.e_mf` as a line within the CCSD class, with the MP2 class imported at the top of the module.

While I'm generally a big fan of properties, I wouldn't use one here for `e_mf` - the reason is that `e_tot` is a property of `e_mf` -
If a user decides to access `cc.e_tot` twice or multiple times in their script (and they have every right to do so: it looks like a harmless attribute access), this will access `e_mf` multiple times and trigger a full HF potential rebuild every time.

One way would be to cache this property with one of the functools caches - but we don't really use this that much in PySCF anywhere else.
Another way would be to make `e_hf` a regular attribute, calculated on `__init__` - but this would make the `__init__` slow, and I prefer if class instantiation is always fast. So I would vote for the following: have `e_mf` be a regular attribute, set to `None` on instantiation, same as `e_corr`. In the kernel, if `mycc.e_mf` is None, it is calculated via your functions, renamed, say, `get_e_mf` (and without the property) and stored.

maxnus:
>I understand that it makes sense to re-evaluate the e_hf of periodic CC/MP2. But for what reason this PR need to apply the same thing for non-periodic CC/MP2 (instead of using self.e_hf = self._scf.e_tot)?

It's not just about the divergent exchange correction in solilds - you can perform a CCSD calculations on top of DFT, but in this case the mean-field energy is not `mf.e_tot` (the Kohn-Sham energy), but you have to evaluate the Hartree-Fock energy of the Kohn-Sham system.

Another case are non-converged HF calculations: in this case you can either use a set of orbitals C, defining the density-matrix D and the resulting Fock matrix, i.e. {C, F[C]} or you use a Fock matrix and the orbitals resulting from it's diagonalization, i.e. {F, C[F]}. At self-consistency both choices should be the same. For non converged SCF, the SCF procedure will terminate with a solution set
{F, C[F]}, but for the post-SCF calculations it might be better to start with {C, F[C]}. At least that's why I am assuming @sunqm made this change in [db64a0](https://github.com/pyscf/pyscf/commit/db64a063f42fee15cbb7dc1b348c2ded88291387)

So the best way would be if we can detect if the mf defines a converged HF solution or not, because in the former case we can instead use `cc._scf.e_tot`, as you said, and save a Fock build for molecules. However, to check this, I can only think of evaluating the commutator FDS - SDF, which requires a Fock build anyway! If one really wants to avoid this, one would probably have to introduce a new flag `converged_hf` or `brillouins_theorem` to the mf class.

verena-neufeld:
Thanks for the suggestions!
- I have introduced a new class attribute, `_e_hf` which is initialised to `None` and then set the first time property `e_hf` is called. The main code to evaluate the HF energy is now only in the MP2 module to reduce code duplication.
- CISD now also uses `e_hf`.
- For the scanner, I had to add `self._e_hf = None` to the reset function. I wonder whether this covers all cases where this should be reevaluated.
- I'm not sure whether the SCF MO coeffs and occ should take priority or the ones that might be passed to the post HF method (and might be different). In same custom hamiltonian cases, no SCF evaluation was done and a post HF calculation was directly started, that is why if SCF MO coeffs and occ are not set, it uses the post HF MO coeffs and occ.

sunqm:
get_e_hf method is taking mo_coeff of underlying SCF object as absent value. Imaging the case the mo_coeff is not derived from canonical HF calculations, e_hf should be evaluated with mo_coeff of ccsd methods. I think it's better to issue an explicit call in build or kernel method of CC class to generate e_hf.

sunqm:
get_e_hf should be consistent with the ao2mo method because the CC calculations are based on the integrals generated by ao2mo. `self.mo_coeff`, which was passed to ao2mo, is also the argument for get_e_hf. `self.mo_coeff` for CC class should be initialized properly before calling the kernel function.

verena-neufeld:
I've made changes, thanks for all the suggestions!
Now e_hf is no longer a property. self.e_hf is evaluated in the kernel and whenever eris is None and calculated (and it makes sense to need e_hf there). I've tried to tie self.e_hf to eris by doing this, so they use the same coeffs etc. However, I'm still not sure whether this is the best way.

pyscf/cc/ccsd.py
```diff
@@ -1042,6 +1046,7 @@ def init_amps(self, eris=None):
         time0 = logger.process_clock(), logger.perf_counter()
         if eris is None:
             eris = self.ao2mo(self.mo_coeff)
+            self.e_hf = self.get_e_hf(mo_coeff=self.mo_coeff)
```
sunqm, line 1049: if eris is specified in input, self.e_hf may not be properly created.

pyscf/cc/ccsd.py
```diff
@@ -1126,7 +1127,9 @@ def ccsd_t(self, t1=None, t2=None, eris=None):
         from pyscf.cc import ccsd_t
         if t1 is None: t1 = self.t1
         if t2 is None: t2 = self.t2
-        if eris is None: eris = self.ao2mo(self.mo_coeff)
+        if eris is None:
+            eris = self.ao2mo(self.mo_coeff)
+            self.e_hf = self.get_e_hf(mo_coeff=self.mo_coeff)
```
sunqm, line 1132: self.e_hf may not properly initialized if eris is specified in input

pyscf/cc/qcisd.py
```diff
@@ -410,10 +410,6 @@ def qcisd(self, t1=None, t2=None, eris=None):
         if eris is None:
             eris = self.ao2mo(self.mo_coeff)
```
sunqm, line 412: Is self.e_hf initialization missed?

pyscf/cc/qcisd_slow.py
```diff
@@ -176,10 +176,6 @@ def qcisd(self, t1=None, t2=None, eris=None):
         if eris is None:
             eris = self.ao2mo(self.mo_coeff)
```
sunqm, line 178: Is self.e_hf initialization missed?

pyscf/ci/cisd.py
```diff
@@ -1004,7 +1007,9 @@ def get_init_guess(self, eris=None, nroots=1, diag=None):
             MP2 energy and initial guess(es) for CISD coefficients.
 
         '''
-        if eris is None: eris = self.ao2mo(self.mo_coeff)
+        if eris is None:
+            eris = self.ao2mo(self.mo_coeff)
+            self.e_hf = self.get_e_hf(mo_coeff=self.mo_coeff)
```
sunqm, line 1012: self.e_hf may not be properly created if input includes eris

pyscf/mp/mp2.py
```diff
@@ -39,6 +39,7 @@ def kernel(mp, mo_energy=None, mo_coeff=None, eris=None, with_t2=WITH_T2, verbos
 
     if eris is None:
         eris = mp.ao2mo(mo_coeff)
+        mp.e_hf = mp.get_e_hf(mo_coeff=mo_coeff)
```
sunqm, line 42: this function should not change mp

pyscf/pbc/cc/kccsd.py
```diff
@@ -391,7 +391,9 @@ def init_amps(self, eris):
         return self.emp2, t1, t2
 
     def ccsd(self, t1=None, t2=None, eris=None, **kwargs):
-        if eris is None: eris = self.ao2mo(self.mo_coeff)
+        if eris is None:
+            eris = self.ao2mo(self.mo_coeff)
+            self.e_hf = self.get_e_hf()
```
sunqm, line 396: self.e_hf may not be created if eris is specified

pyscf/pbc/cc/kccsd.py
```diff
@@ -417,7 +419,9 @@ def ccsd_t(self, t1=None, t2=None, eris=None):
         from pyscf.pbc.cc import kccsd_t
         if t1 is None: t1 = self.t1
         if t2 is None: t2 = self.t2
-        if eris is None: eris = self.ao2mo(self.mo_coeff)
+        if eris is None:
+            eris = self.ao2mo(self.mo_coeff)
+            self.e_hf = self.get_e_hf()
```
sunqm, line 424: self.e_hf may not be created if eris is specified

pyscf/pbc/cc/kccsd_rhf.py
```diff
@@ -628,7 +626,9 @@ def ccsd_t(self, t1=None, t2=None, eris=None):
         from pyscf.pbc.cc import kccsd_t_rhf
         if t1 is None: t1 = self.t1
         if t2 is None: t2 = self.t2
-        if eris is None: eris = self.ao2mo(self.mo_coeff)
+        if eris is None:
+            eris = self.ao2mo(self.mo_coeff)
+            self.e_hf = self.get_e_hf()
```
sunqm, line 631: self.e_hf may not be created if eris is not None

pyscf/cc/ccsd.py
```diff
@@ -700,7 +701,9 @@ def energy(mycc, t1=None, t2=None, eris=None):
     '''CCSD correlation energy'''
     if t1 is None: t1 = mycc.t1
     if t2 is None: t2 = mycc.t2
-    if eris is None: eris = mycc.ao2mo()
+    if eris is None:
+        eris = mycc.ao2mo()
+        mycc.e_hf = mycc.get_e_hf()
```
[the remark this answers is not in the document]
verena-neufeld: I've deleted the e_hf evaluation here.

pyscf/cc/ccsd.py
```diff
@@ -46,6 +46,7 @@ def kernel(mycc, eris=None, t1=None, t2=None, max_cycle=50, tol=1e-8,
     log = logger.new_logger(mycc, verbose)
     if eris is None:
         eris = mycc.ao2mo(mycc.mo_coeff)
+        mycc.e_hf = mycc.get_e_hf(mo_coeff=mycc.mo_coeff)
```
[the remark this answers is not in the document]
verena-neufeld: I've deleted the e_hf evaluation here.

pyscf/cc/ccsd.py
verena-neufeld: if self.e_hf is None, e_hf will be evaluated new, otherwise e_hf = self.e_hf. e_hf is needed in this function.

pyscf/cc/ccsd.py
verena-neufeld: I've deleted the self.e_hf evaluation here.

pyscf/cc/qcisd.py
verena-neufeld: I've added self.e_hf evaluation.

pyscf/cc/qcisd_slow.py
verena-neufeld: I've added self.e_hf evaluation.

pyscf/ci/cisd.py
verena-neufeld: I've deleted the self.e_hf evaluation.

pyscf/mp/mp2.py
verena-neufeld: I've deleted the mp.e_hf evaluation.

pyscf/pbc/cc/kccsd.py
verena-neufeld: I've deleted the self.e_hf evaluation here.

pyscf/pbc/cc/kccsd.py
verena-neufeld: I've deleted the self.e_hf evaluation here.

pyscf/pbc/cc/kccsd_rhf.py
verena-neufeld: I've deleted the self.e_hf evaluation here.

verena-neufeld:
Thanks @sunqm , I have now deleted a lot of e_hf evaluations. They are now only in the class method def ccsd and related, as well as the init amps function where e_hf is needed in the function. As long as the user calls the post HF method in a classical way, e.g. via kernel(), self.e_hf should be set. I still wonder whether this is sufficient/best. I'm happy to change things.

sunqm (approved):

PR #1376, opened 2022-08-20
Make UCASSCF work with state_average

hczhai:
Currently, the following code (UCASSCF with state_average) does not work because ``make_rdm12s`` is missing in ``mcscf.addons.state_average``. This pull request adds ``make_rdm12s`` in ``mcscf.addons.state_average`` and ``mcscf.addons.state_average_mix`` to allow this.

    from pyscf import scf, gto, mcscf
    import numpy as np
    mol = gto.M(atom="F 0 0 0", basis='cc-pvdz-dk', verbose=3, spin=1)
    mf = scf.UHF(mol).run(conv_tol=1e-14)
    mc = mcscf.UCASSCF(mf, 4, (4, 3)).state_average_(np.ones(3) / 3.0).run()

### DIFF
--- a/pyscf/mcscf/addons.py
+++ b/pyscf/mcscf/addons.py
@@ -954,6 +954,24 @@ def state_average(casscf, weights=(0.5,0.5), wfnsym=None):
             rdm2 = numpy.einsum ('r,rpqst->pqst', self.weights, rdm2)
             return rdm1, rdm2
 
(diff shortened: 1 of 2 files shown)


sunqm:
A small test to cover the new code needs to be added

fishjojo (approved):

Issue #1405, opened 2022-09-06
PySCF 2.1.0 wheels for MacOS and python > 3.7 don't install


manoelmarques:
The pyscf 2.1.0 distribution for MacOS changed the scipy requirements from scipy (!=1.5.0,!=1.5.1) to scipy (<=1.1.0).
However there is no scipy (<=1.1.0) distribution for python > 3.7 under MacOS.
For the MacOS, under python 3.8, it reverts to pyscf 2.0.1 and works, however for python 3.10, it can't install the only wheel existent for python 3.10 and fails.

Scipy <= 1.1.0 is very old and while there are wheels for python <=3.7, there aren't any for python > 3.7

So it seems to me that the current MacOS wheels for python > 3.7 don't work.

sunqm:
scipy dependency check is removed in pyscf-2.2

PR #1407, opened 2022-09-06
Loosen version requirments for scipy (issue #1405)

sunqm:
A new version needs to be released


fishjojo (approved):

dasmy:
Thanks for the change. Using `pip install git+https://github.com/sunqm/pyscf@scipy-deps`, I could finally install psycf again.

PR #1416, opened 2022-09-09
WIP: GHF real2complex stability, complex GHF internal stability and other improvements

### LINKED ISSUE #1363: Is it possible to check real-to-complex instability for GHF?
In some cases, breaking the symmetry of a real GHF solution by adding a few `0.1j` to the dm (or rotate it with complex coefficients) helps us find a lower complex solution. But currently, it seems the `stability.ghf_stability` cannot detect such instability. I'm wondering if this can be regarded as a kind of instability and if so how to detect it.

hebrewsnabla:
real2complex stability check for real GHF was added in this PR (may resolve #1363). The basic idea is to diagonalize A-B. I didn't find any paper to support this approach, but found that in [GQCP source code](https://github.com/GQCG/GQCP/blob/develop/gqcp/include/QCModel/HF/StabilityMatrices/GHFStabilityMatrices.hpp).

Some other improvements:
* allow user modifying Davidson solver tol in stability. This is important in testing UHF stability for some hard case like ozone (with only 1e-4 tol, severe unreproducibility is observed)
* move stable_opt_internal from example to scf/stability

Still, I am not sure this real2complex implementation is right. ~~And it remains a question that how can we do an internal stability check for complex GHF? GQCP's approach is to diagonalize ((A B) (B* A*)), but I don't know whether that's right.~~


yangdatou:
JCP 66, 3045 (1977); doi: 10.1063/1.434318 shall be a good reference for this topic.

hebrewsnabla:
There's still some discrepancy between scf.stability and scf.stability_slow. Running a recently-added example scf/17-stability-h2 will give (value with * is newly implemented by this PR):

| eigenvalue of H | RHF in. | RHF r2c | RHF ex. | UHF in. | UHF r2c | UHF ex. | GHF in. | GHF r2c |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | 
| stability(davidson) | 2.316 | 0.472 | 0.298 | 1.158 | 0.472 | 0.298 | 0.596 | 0.403* |
| stability_slow(exact) | 0.579 | 0.472 | 0.298 | 0.298 | 0.472 | 0.298 | 0.298* | 0.298* |

Some of them differ by a factor 4 or 2, which is not a big problem, I think. But I still can't tell which implementation of GHF r2c is wrong, or both.

ajz34:
@hebrewsnabla Simply change `hermi=2` to `hermi=0` may resolve the problem of GHF r2c value (from wrong 0.403 to 0.298). Code of concern is at this place:
https://github.com/hebrewsnabla/pyscf/blob/554932ae46797572862693a9eb3fc957be2ef091/pyscf/scf/stability.py#L260-L262

It seems that current implementation of `pyscf.scf.ghf.get_jk` only handles `hermi=0` or `hermi=1` (symmetric). Behavior of `hermi=2` in this function is the same to `hermi=1`. But `dm1` defined in `hop_real2complex` is anti-symmetric.

> Some of them differ by a factor 4 or 2, which is not a big problem, I think.

By comparison to results from one other software (all RHF/UHF, internal/r2c/external are available, but without GHF), I guess eigenvalues of `stability_slow(exact)` is correct. Maybe some constants can be multiplied to `hessian_x` functions.

hebrewsnabla:
I'll split this (too complicated) pr into a few small ones and start with some simple improvement pr like adjusting tol and nroot. So close this first.

Issue #1371, opened 2022-08-17
Features requests


sunqm:
Based on the discussions in the 2022 PySCF developers' meeting, requested features are collected in the list below. If you are looking for features not available in PySCF or you are interested in contributing new features, this list can be used as a reference. 

[Names] in the "[]" at the end of each statement are the people who are currently developing or have plans to develop the feature. The list will be updated roughly monthly.

People
======

AYS = Alex Sokolov or Sokolov group
CL = Chenghan Li
GC = Garnet Chan or Chan group
HY = Hongzhou Ye
HZ = Huanchen Zhai
JK = Janus Kristen
JY = Junjei Yang
KL = Ke Liao
MRH = Matthew R Hermes
NR = Nick Rubin
QS = Qiming Sun
SL = Seunghoon Lee
SS = Sandeep Sharma or Sharma group
TB = Tim Berkelbach or Berkelbach group
TZ = Tianyu Zhu
XW = Xiao Wang
XZ = Xing Zhang
YZ = Yu Zhang
ZC = Zhihao Cui

Features
========

* GPU plugins for pyscf [QS/GC]
  - Hartree-Fock (RHF, UHF, GHF) on GPU [QS/GC]
  - Release pipelines of gpu plugins, mpi plugins [QS]
  - DFT on GPU [QS]
  - RCCSD on GPU [QS]
  - AO evaluation on GPU would benefit PyQMC [QS]
* Integration of SymTensor/ctf-tensor/ctf-cc as addons [GC needs help with PR]
* Support vendors other than NVIDA [JK]
* HF and DFT nuclear gradients on GPU
* Extension of molecular GPU HF to PBC
* pbc.df module in mpi4pyscf needs to be tested and updated against pyscf v2
* Benchmark GPU dispatch from JAX PySCFAD for many methods
* df ao2mo on GPU
* FCI on GPU
* FCI with MPI
* plane wave-based HF and post-HF [HY]
* Integral-direct DFT, HF and MP2 with Gaussian density fitting [HY]
* PBC integrals evaluated in real space (the RSJK module) [QS]
* fast FFTDF J-build [GC/XZ]
* fast exchange from robust pseudospectral method [SS]
* Match popular PW-based DFT code
  - k-point symmetry
  - PW basis
* DFT
  - k-point symmetry
  - Non-local xc (e.g., VV10)
  - Pseudopotential (e.g., UPF with GTOs)
  - 1st derivative: forces, stress and geom opt
[...]

PR #1432, opened 2022-09-21
Enable molecular nuc grad with DF ERIs for ROHF, ROKS, CASSCF, SA-CASSCF

MatthewRHermes:
Also add unit tests for molecular nuclear gradients with density-fitted two-electron integrals for ROHF, ROKS, UHF, UKS, CASSCF, and SA-CASSCF methods. This involved significant refactoring of df/grad/rhf.py, which ended up including some performance optimizations I had been using over the past couple of years in _mrh_.

This was mentioned at the 2022 developer's meeting (issue #1371). BTW, although ROHF and ROKS were "implemented" already, they weren't actually working because of an oversight involving the make_rdm1e method. IDK if this is already an issue somewhere.

### DIFF
--- a/pyscf/df/grad/rhf.py
+++ b/pyscf/df/grad/rhf.py
@@ -405,7 +405,7 @@ def _cho_solve_rhojk (mf_grad, mol, auxmol, orbol, orbor):
     class get_rhok_class (object):
         def __init__(self, my_f):
             self.f_rhok = my_f
-        def __call__(set_id, p0, p1):
+        def __call__(self, set_id, p0, p1):
             buf = numpy.empty((p1-p0,nocc[set_id],nao))
             col1 = 0
             for istep in range(nsteps):
--- a/pyscf/df/grad/rhf.py
+++ b/pyscf/df/grad/rhf.py
@@ -38,10 +38,10 @@ from functools import reduce
 from itertools import product
 from pyscf.ao2mo import _ao2mo
 
-def get_jk(mf_grad, mol=None, dm=None, hermi=0, with_j=True, with_k=True, ishf=True):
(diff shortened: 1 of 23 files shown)


pyscf/df/grad/rhf.py
```diff
@@ -241,6 +291,127 @@ def get_int3c(shls_slice=None):
...
+            t2 = logger.timer_debug1 (mf_grad, 'df grad cache D_Pmi (m <-> i transpose upon retrieval)', *t2)
+        int3c = v = None
+    rhoj = scipy.linalg.cho_solve(int2c, rhoj.T).T
+    int2c = None
+    t1 = logger.timer_debug1 (mf_grad, 'df grad vj and vk AO (P|Q) D_Q = (P|mn) D_mn solve', *t1)
+    def get_rhok(set_id, p0, p1):
+        buf = numpy.empty((p1-p0,nocc[set_id],nao))
+        col1 = 0
+        for istep in range(nsteps):
+            dat = f_rhok['%s/%s'%(set_id,istep)][p0:p1]
```
sunqm, line 409: The temporary file object `f_rhok` may be closed when exiting `_cho_solve_rhojk`. It needs to be attached to an object that can survive as long as the `get_rhok` function.

sunqm (approved):

pyscf/df/grad/rhf.py
MatthewRHermes: Does ccbf8de fix this?

sunqm: I was wrong for the first version you implemented. f_rhok was held in the closure of get_rhok. You can keep either version as you like.

MatthewRHermes: I prefer the new version, since it's feels like good practice to keep the temporary file potentially accessible as a tag IMO.

Issue #1422, opened 2022-09-13
cell.precision for 2D systems changed in 2.1


maxnus:
We have some tests failing downstream since 2.1, which seem to suddenly require a much lower `cell.precision` to pass.
This affects a 3x2x1 supercell of a square lattice of He atoms (example script below).

With version 2.0 the results are:
```
k-HF:  converged=  True  E=-2.855155880475
G-HF:  converged=  True  E=-2.855155880468
```

With version 2.1 the results are:

```
k-HF:  converged=  True  E=-2.855155880613
G-HF:  converged= False  E=-2.855140202195
```
with the following warning for the Gamma-point supercell calculation:

> WARN: 0.008379 error found in overlap integrals. cell.precision  or  cell.rcut  can be adjusted to improve accuracy.

Only when lowering `cell.precision` to `1e-15` will this warning dissappear, and I get

```
k-HF:  converged=  True  E=-2.855155880468
G-HF:  converged=  True  E=-2.855155880468
```

It appears that some truncation or resolution parameter is now underestimated in this 2D supercell system, since we shouldn't need to go to such a low precision.
 
This doesn't appear to be an issue when using 3D PBC.
(edit: although possibly related to #1409)
 
 Script:
 [run.py.txt](https://github.com/pyscf/pyscf/files/9557081/run.py.txt)

verena-neufeld:
Hi @maxnus , thanks for opening this issue. Has it been resolved now?

sunqm:
Pyscf-2.6 and newer have a tighten threshold on integral precision. The gamma point and kpoints calculation should both provide `E=-2.855155881`

Issue #1366, opened 2022-08-12
ke_guess from estimate_ke_cutoff_for_eta too low


maxnus:
In the PBC calculation below the nuclear potential has an imaginary part at k-points which are self-conjugated (i.e. k = G-k) for some inverse lattice vector G). At these k-points the nuclear potential should be real.

This is however not the case, even with a `cell.precision` of `1e-11`. To remove the imaginary part one has to use `1e-12`, which seems really quite low.  

It seems that the kinetic energy cutoff [here](https://github.com/pyscf/pyscf/blob/2bd16d760ddcbbcec4ca5d7b7b480b953743459b/pyscf/pbc/df/aft.py#L160) is not estimated conservatively enough.

This may be related to the fact that the unit cell below us quite "sharp", with lattice vectors far from being orthogonal (I don't know if a cubic cell was assumed for the ke estimate, just a guess).

```
import numpy as np
import pyscf
import pyscf.pbc
import pyscf.pbc.df
import pyscf.lib


cell = pyscf.pbc.gto.Cell()
a  = 4.1705
cell.a = a * np.asarray([
    [1, 0.5, 0.5],
    [0.5, 1, 0.5],
    [0.5, 0.5, 1.0]])
cell.atom = [
        ['Ni', [0, 0, 0]],
        ['Ni', [a, a, a]]]
cell.basis = 'def2-svp'
cell.output = 'pyscf.txt'
cell.verbose = 10
cell.exp_to_discard = 0.1
cell.precision = 1e-10
#cell.precision = 1e-12
cell.build()

kmesh = [2,1,1]
kpts = cell.make_kpts(kmesh)

df = pyscf.pbc.df.GDF(cell, kpts)
vnuc = pyscf.lib.asarray(df.get_nuc(kpts))
for k in range(len(kpts)):
    im = np.linalg.norm(vnuc[k].imag)
    print("Imaginary part of V(nuc) at kpt= %3d:  %.3e" % (k, im))
```

maxnus:
Similarly, the Ewald parameters used for the nuclear-nuclear energy [here](https://github.com/pyscf/pyscf/blob/2bd16d760ddcbbcec4ca5d7b7b480b953743459b/pyscf/pbc/gto/cell.py#L758-L759)
are not well converged with `precision = 1e-8` for this system.

sunqm:
With the updates in PR #1402 #1429 , the required accuracy can be reached.

PR #1441, opened 2022-09-25
Improve pbc integral accuracy and various improvements and bugfixes

sunqm:
* Improve ewald function accuracy (#1366)
* Fix bug in get_lattice_Ls for 2D systems (#1422)
* Improve mesh grids quality for numerical integration under PBC
* Improve the implementation for Makov Payne correction
* Improve the treatment of absent methods for Mole and Cell class
* Add hybrid functional KGKS

### DIFF
--- a/pyscf/pbc/gto/cell.py
+++ b/pyscf/pbc/gto/cell.py
@@ -739,25 +739,27 @@ def get_ewald_params(cell, precision=None, mesh=None):
         ew_cut = cell.rcut
         ew_eta = np.sqrt(max(np.log(4*np.pi*ew_cut**2/precision)/ew_cut**2, .1))
     else:
-        if mesh is None:
-            mesh = cell.mesh
         mesh = _cut_mesh_for_ewald(cell, mesh)
-        Gmax = min(np.asarray(mesh)//2 * lib.norm(cell.reciprocal_vectors(), axis=1))
-        log_precision = np.log(precision/(4*np.pi*(Gmax+1e-100)**2))
-        ew_eta = np.sqrt(-Gmax**2/(4*log_precision)) + 1e-100
+        ke_cutoff = pbctools.mesh_to_cutoff(cell.lattice_vectors(), mesh).min()
+        charge = cell.atom_charges().sum()
+        log_precision = np.log(precision / (charge*16*np.pi**2))
+        ew_eta = np.sqrt(-ke_cutoff/(2*log_precision))
         ew_cut = _estimate_rcut(ew_eta**2, 0, 1., precision)
     return ew_eta, ew_cut
 
+MESH_FOR_EWALD = np.array([15] * 3)
+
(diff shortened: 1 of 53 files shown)


pyscf/pbc/gto/cell.py
```diff
@@ -909,7 +922,7 @@ def make_kpts(cell, nks, wrap_around=WRAP_AROUND, with_gamma_point=WITH_GAMMA,
         kpts = libkpts.make_kpts(cell, kpts, space_group_symmetry, time_reversal_symmetry, symmorphic)
     return kpts
 
-def get_uniform_grids(cell, mesh=None, **kwargs):
+def get_uniform_grids(cell, mesh=None, wrap_around=True):
```
fishjojo, line 925: I think setting `wrap_around` default as `True` may break other codes following the previous convention.

pyscf/pbc/scf/hf.py
```diff
@@ -237,27 +237,30 @@ def dip_moment(cell, dm, unit='Debye', verbose=logger.NOTE,
...
-        #ke_cutoff = gto.estimate_ke_cutoff(cell, 1e-5)
-        #grids.mesh = tools.cutoff_to_mesh(a, ke_cutoff)
+
     if rho is None:
         rho = numint.NumInt().get_rho(cell, dm, grids, kpt, cell.max_memory)
 
+    # With the optimal origin of the unti cell, the net dipole in the unit
+    # cell should be strictly zero. However, the integral grids are often not
+    # enough to produce the zero dipole. Errors are caused by the sub-optimal
+    # origin and the numerial integration.
```
fishjojo, line 247: I think the dipole moment calculation should follow the implementation in [here](https://github.com/pyscf/properties/blob/92a4df7645033349232c4776355f8e53a9256987/pyscf/pbc/prop/polarizability/rhf.py#L237), although the accuracy of k-point derivatives need to be improved...

fishjojo (approved):

pyscf/pbc/gto/cell.py
sunqm: When quantities and integrals are perfectly converged, there's no difference if the grids are inside the unit cell or wrapped around against the point (0,0,0) . If quantities are barely converged, it may have small difference to the other convention. This setting favors the systems in which atoms are placed near the edge of the cell, e.g. a 2D slab.

pyscf/pbc/scf/hf.py
sunqm: The definition of dipole in the cell is vague. It depends on how the box of the cell is constructed. The dipole computed in this function is very likely different to the one you referred. It optimizes the cell center so as to minimize the dipole in the cell. This routine is used in this module to compute the Makov-Payne correction for charged systems. I'm not sure if this is the correct way to define dipole in a PBC system. Is the one you referred the better definition for dipole in pbc systems? If so could you please update the dipole function in a new PR?

PR #1443, opened 2022-09-26
bug fix number of k points in each direction

verena-neufeld:
This fixes a bug affecting for example the madelung constant evaluation and exxdiv=vcut_ws, etc. It was noticed for shifted k point meshes.

The unique function might consider two k point coordinates as distinct even though they represent the same k point coordinate and only differ very slightly due to floating point noise. Consider 0.055000000001 and 0.0549999999999 (made-up example), which might get rounded to 0.06 and 0.05 respectively and therefore considered different, even though they belong to the same k point. This fix looks at the differences between coordinates directly and is hopefully a little more robust.

How to set the minimum tolerance for the difference is a question. If it is too low, floating point deviations might cause two coordinates being considered distinct even though they are not. If it is too high, we limit the size of k point meshes. Currently, it is set at 1e-5.

Credits for code and discussion goes to @tberkel and @hongzhouye .

### DIFF
--- a/pyscf/pbc/tools/pbc.py
+++ b/pyscf/pbc/tools/pbc.py
@@ -468,9 +468,9 @@ def madelung(cell, kpts):
         return 2*cell.omega/np.pi**0.5-np.einsum('i,i,i->', ZSI.conj(), ZSI, coulG*weights).real
 
 
-def get_monkhorst_pack_size(cell, kpts):
+def get_monkhorst_pack_size(cell, kpts, tol=1e-5):
     kpts = np.reshape(kpts, (-1,3))
-    min_tol = 1e-5
+    min_tol = tol
     assert kpts.shape[0] < 1/min_tol
     if kpts.shape[0] == 1:
         Nk = np.array([1,1,1])
--- a/pyscf/pbc/tools/pbc.py
+++ b/pyscf/pbc/tools/pbc.py
@@ -469,8 +469,17 @@ def madelung(cell, kpts):
 
 
 def get_monkhorst_pack_size(cell, kpts):
-    skpts = cell.get_scaled_kpts(kpts).round(decimals=6)
-    Nk = np.array([len(np.unique(ki)) for ki in skpts.T])
+    kpts = np.reshape(kpts, (-1,3))
+    min_tol = 1e-5
+    assert kpts.shape[0] < 1/min_tol
+    if kpts.shape[0] == 1:
+        Nk = np.array([1,1,1])
+    else:
+        tol = max(10**(-int(-np.log10(1/kpts.shape[0]))-2), min_tol)
+        skpts = cell.get_scaled_kpts(kpts)
+        Nk = np.array([np.count_nonzero(abs(ski[1:]-ski[:-1]) > tol) + 1
+                       for ski in np.sort(skpts.T)])
+    assert np.prod(Nk) == kpts.shape[0]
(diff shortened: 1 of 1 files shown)


verena-neufeld:
Meshes like this:
```
kpts = array([[ 0.26682728,  0.62113383,  0.52914209],
       [-0.26682728, -0.62113383, -0.52914209]])
```
were not considered by us when checking that prod(Nk) = length of kpts array. Here, Nk = [2,2,2], giving the unique k points in each direction. There are only two k points though, not eight.
A solution is to delete the check (not ideal but can't see any solution at this moment).

maxnus:
I'm slightly confused about the intial error - your examples should both round to 0.055000 = 0.055 when using `round(decimals=6)`.
Of course if you have enough leading zeros, at some point you would get different rounding. k-points can become small in big unit cells, but `cell.get_scaled_kpts` returns them dimensionless, so they shouldn't be that small?

verena-neufeld:
Good point! This is an actual example that I had commented somewhere else: due to numerical noise, two scaled k point coordinates that should be the same number, were 0.05167749999999998 and 0.05167750000000004, being rounded to 0.051677 and 0.051678 respectively and therefore found not to be the same.

maxnus:
Right! And I guess just reducing `decimals` to say 10 wouldn't be a robust solution either, since for another grid you might have a 5 in the 10th decimal position!

tberkel:
Indeed. Assuming they come from a uniform mesh (as the function name
suggests), then you only need to compare k-points to the order of
1/nkpts to decide if they are the same kpt or not.

verena-neufeld:
The tests currently fail due to two k points like this:
```
kpts = array([[ 0.26682728,  0.62113383,  0.52914209],
       [-0.26682728, -0.62113383, -0.52914209]])
```
However, they don't represent a uniform monkhorst mesh, so maybe the function `get_monkhorst_pack_size` should not be called? It happens in multigrid DFT. I'm having a closer look.

verena-neufeld:
@sunqm is it correct that multigrid DFT evaluates the madelung constant with Nk = [2,2,2], even though only two k points are present?

sunqm:
> @sunqm is it correct that multigrid DFT evaluates the madelung constant with Nk = [2,2,2], even though only two k points are present?

madelung constant is evaluated against [2,2,2] . The code searches in each direction the number of different k points. Multigrid DFT uses the same code as the regular DFT code

sunqm:
> Investigating test failures. Meshes like this:
> 
> ```
> kpts = array([[ 0.26682728,  0.62113383,  0.52914209],
>        [-0.26682728, -0.62113383, -0.52914209]])
> ```
> 
> were not considered by us when checking that prod(Nk) = length of kpts array. Here, Nk = [2,2,2], giving the unique k points in each direction. There are only two k points though, not eight. A solution is to delete the check (not ideal but can't see any solution at this moment).

The check prod(Nk) == len(kpts) may be skipped here. When k point symmetry is considered, it may also happen that the prod(Nk) != len(kpts)

tberkel:
It’s a bit dangerous to infer the Monkhorst Pack mesh from only a limited
set of k-points, but I suppose as long as it’s always used carefully, the
assert can be removed.

verena-neufeld:
Thanks for the comments. I've removed the assert.

fishjojo (approved):

PR #1446, opened 2022-09-27
a fix for 4C-TDDFT using GGA functionals in multi-collinear approach

Haskiy:
It leads to a error for 4C-TDDFT using GGA functionals in multi-collinear approach.


sunqm:
Another similar bug is in dft.numint2c module. Could you also fix the numint2c module?

sunqm (approved):

Haskiy:
The second commit is a fix for the potential bug for 2c-TDDFT calculation using GGA functionals in multi-collinear approach as you said above, but I cannot give a suitable test to present this bug. In addition, a same usage of numpy. hstack() function lies in dft.numit module.

sunqm:
In numint.py, the density is at most 2-dimension array. numpy.hstack with default options works fine.

PR #1448, opened 2022-09-27
Moment resolved GF-CCSD

obackhouse:
This PR implements the moment-resolved approximate GFCCSD solver developed by @ghb24 and I, as detailed in [arXiv:2206.13198](https://arxiv.org/abs/2206.13198). This method involves constructing moments of the single-particle Green's function at the level of CCSD and using an effective block Lanczos solver to generate an approximate Green's function, explicitly represented by a pole structure. This can be used to efficiently perform GFCCSD calculations with no algorithmic dependence on the frequency grid or choice of broadening.

Changes:
 - Adds `pyscf/cc/momgfccsd.py`, with the class `MomGFCCSD` to control calculations.
 - Adds `pyscf/cc/test/test_momgfccsd.py` with unit tests.
 - Adds `pyscf/cc/5...` detailing some example calculations and more details such as generating self-energies.

The expressions for the bra and ket states required to build the moments are the same as those required in traditional GFCCSD codes. I adapted these expressions from the stale branch `kccgf`, which implements a more standard GFCCSD algorithm using a GCROT solver. I would be happy to revive this branch and include another class and kernel with this solver if anyone is interested.


ssh2 (approved):
This PR looks good to me

PR #1455, opened 2022-10-11
MAINT: Add starting configurations for f-in-core ECPs

bogdanoff:
PR defines  core configurations for 4f and 5f in-core ECPs. As of now the usage of such ECP, e.g. [https://www.tc.uni-koeln.de/cgi-bin/pp.pl?language=en,format=molpro,element=Gd,job=getecp,ecp=ECP54MWB](https://www.tc.uni-koeln.de/cgi-bin/pp.pl?language=en,format=molpro,element=Gd,job=getecp,ecp=ECP54MWB), will raise an error.
Cases when `nelec_core` is 46, 54 or 78 require knowledge of the atom type, as ECPs with different core configurations exist.

### DIFF
--- a/pyscf/gto/ecp.py
+++ b/pyscf/gto/ecp.py
@@ -37,6 +37,7 @@ import ctypes
 import numpy
 from pyscf import lib
 from pyscf.gto import moleintor
+from pyscf.data.elements import ELEMENTS
 
 libecp = moleintor.libcgto
 libecp.ECPscalar_cache_size.restype = ctypes.c_int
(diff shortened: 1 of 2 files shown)


sunqm (approved):

hebrewsnabla:
Hi Nikolay,

It seems that doing SCF with f-in-core atoms still raises error, because the atom symbol is not passed to ecp.core_configuration in minao init guess

https://github.com/pyscf/pyscf/blob/9ae825adb22174542deae3c2b6d250d93e770f53/pyscf/scf/hf.py#L367

And after some grep, it looks like in entire pyscf there's a couple of calls to this function without atom_symbol specified, should they all need to be updated?

bogdanoff:
Hi Shirong,
Thank you for checking this. Indeed I recall problems with SCF initial guess, which I get around with `mf.init_guess='1e'`. I think you are right, that more changes have to be done here and there.

PR #1456, opened 2022-10-11
Small fixes

bogdanoff:
Handling of the arccos argument in `pyscf/symm/Dmatrix.py` is the main purpose of this PR.
But I also push some small changes here and there.

PR #1474, opened 2022-10-19
Fixes numerical domain issue in get_euler_angles

maxnus:
The value `zz` can be slightly outside the interval [-1,1] due to finite precision, causing `numpy.arccos(zz)` to return `nan`.
This fixes this issue.


maxnus:
Fixed in [b6682d1](https://github.com/pyscf/pyscf/commit/b6682d1877b79e28c36ea930649281ed4c2ea5ad)
@bogdanoff Sorry, I didn't see you already had PR #1456 for this

PR #1481, opened 2022-10-21
Fix B3LYP functional.

### LINKED ISSUE #1480: B3LYP is wrong in PySCF
The B3LYP functional is wrong in PySCF. The correct functional is provided by libxc as HYB_GGA_XC_B3LYP. The functional is supposed to have the RPA variant of VWN, not the recommended VWN functional (VWN5); see e.g. https://doi.org/10.1016/S0009-2614(97)00207-8

susilehtola:
Closes #1480


sunqm:
I think it is acceptable to change the default B3LYP functional. I added an option to configure the default B3LYP. @fishjojo @gkc1000 what do you think? Can we break the backward consistency?

fishjojo:
> I think it is acceptable to change the default B3LYP functional. I added an option to configure the default B3LYP. @fishjojo @gkc1000 what do you think? Can we break the backward consistency?

@sunqm, I think it is fine as long as one can restore the previous convention.

fishjojo (approved):

PR #1498, opened 2022-11-03
Create esp.py

swillow:
Estimate Electrostatic potential (ESP) per each grid and atomic point charges.


sunqm:
I think it is appropriate to put this PR in pyscf-forge repo first.

swillow:
Okay.

PR #1504, opened 2022-11-10
Fully spin-polarized UMP2, DF-UMP2, UCCSD, & UCCSD(T) bug fixes

### LINKED ISSUE #1500: Bad workspace sizes for DF-UMP2 w/o beta electrons
Fully spin-polarized calculations have workspace sizing problems with DF-UMP2. A minimal example of the problem is:

```
from pyscf.gto import Mole
from pyscf.scf import UHF
from pyscf.mp.dfump2_native import DFUMP2

mol = Mole()
mol.atom = '''
H   0.0   0.0   0.0
H   0.0   0.0   0.8
'''
mol.basis = 'cc-pVDZ'
mol.spin = 2
mol.build()

mf = UHF(mol).run()
DFUMP2(mf).run()
```

I tinkered with some quick fixes that avoided the sections causing the errors, but it doesn't seem like a completely trivial problem to fix.

godotalgorithm:
Fixes #1500. This is by no means a thorough attempt to fix this type of problem throughout the code, just all of the places that my present workflow is touching. With the UCCSD(T) fix, the problem could alternatively be fixed higher up in the logic, and that might be a slightly cleaner fix than what I have done. My understanding is that many of the CCSD(T) corrections (e.g. aaa, bbb, aab, etc.) are trivially zero in the absence of alpha or beta electrons, and so the low-level C routines that calculate them can be skipped altogether. My fix does not skip trivial aaa/bbb calculations as the low-level routines still seem to work in the trivial case, but the aab-type corrections must be skipped because they lack low-level checks for no alpha/beta electrons and will crash.

### DIFF
--- a/pyscf/cc/uccsd.py
+++ b/pyscf/cc/uccsd.py
@@ -511,9 +511,12 @@ def _add_vvvv(mycc, t1, t2, eris, out=None, with_ovvv=False, t2sym=None):
             u2bb[otrilb[1],otrilb[0]] = u2tril.transpose(0,2,1)
             u2bb[otrilb] = u2tril
 
-        u2ab = _ao2mo.nr_e2(buf[-nocca*noccb:].reshape(nocca*noccb,nao**2), mo,
-                            (0,nvira,nvira,nvira+nvirb), 's1', 's1')
-        u2ab = u2ab.reshape(t2ab.shape)
+        if nocca*noccb > 0:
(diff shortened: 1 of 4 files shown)


sunqm (approved):

godotalgorithm:
FYI, with the very large timing discrepancy between the linting and building steps in your CI, you might want to look into making the CI steps conditional on successful linting. This will be a bit less wasteful when a PR fails the linting but triggers all of the builds anyways, like mine just did.

PR #1514, opened 2022-11-13
Fix memory leaks in C kernels for CCSD(T)

### LINKED ISSUE #1513: Memory leak in CCSD(T) calculations
This is a follow-up issue on #1501, which was conflating problems in PySCF with a variety of user errors. Now that I'm more familiar with my workflow and have some high-throughput CCSD(T) calculations working, I can see more clearly that memory is really leaking. While I can avoid the memory problem by separating my workflow into one CCSD(T) calculation per Python script, I'd certainly prefer it if memory wasn't leaking in the future.

A minimal example demonstrating the memory leak on Linux (the problem is harder to see on non-Linux systems where `lib.current_memory()` doesn't work):
```
from pyscf.gto import Mole
from pyscf.scf import RHF
from pyscf.cc import RCCSD
from pyscf.lib import current_memory

import tracemalloc
tracemalloc.start()

mol = Mole()
mol.atom = '''
He  0.0 0.0 0.0
'''
mol.basis = 'def2-QZVPP'
mol.build()

mf = RHF(mol)
mf.verbose = 0
mf.kernel()

cc = RCCSD(mf)
cc.direct = True
cc.verbose = 0
cc.kernel()

while True:
    t = cc.ccsd_t()

    # total memory usage
    print("total memory: %.1f MiB" % current_memory()[0])

    # in-Python memory usage
    snapshot = tracemalloc.take_snapshot()
    top_stats = snapshot.statistics('lineno')
    total = sum(stat.size for stat in top_stats)
    print("in-Python memory: %.1f MiB" % (total / 1048576))
```

In this simple calculation, about 1 MB of non-Python memory is leaking every 4 CCSD(T) calculations. For larger calculations, the amount of leaked memory is much more substantial. The fact that the CCSD(T) correction is trivial in this 2-electron example isn't preventing the leak from happening.

I'm still trying to understand and fix this problem myself, and I'll post any other information that I gather, but I will probably end up triaging the problem by wrapping my long PySCF scripts as a sequence of short scripts, which will prevent leaked memory from building up.

godotalgorithm:
Fixes #1513. There were a bunch of `malloc` statements without corresponding `free` statements. I've done some simple testing of both CCSD(T) and UCCSD(T) calculations, and I'm not observing any more leaked memory.

### DIFF
--- a/pyscf/lib/cc/ccsd_t.c
+++ b/pyscf/lib/cc/ccsd_t.c
@@ -421,6 +421,7 @@ void CCsd_t_contract(double *e_tot,
         *e_tot += e;
 }
         free(permute_idx);
+        free(jobs);
 }
 
 void QCIsd_t_contract(double *e_tot,
@@ -470,6 +471,7 @@ void QCIsd_t_contract(double *e_tot,
         *e_tot += e;
 }
         free(permute_idx);
+        free(jobs);
 }
 
 
@@ -645,6 +647,7 @@ void CCsd_t_zcontract(double complex *e_tot,
         *e_tot += e;
 }
         free(permute_idx);
+        free(jobs);
 }
 
 void QCIsd_t_zcontract(double complex *e_tot,
@@ -697,6 +700,7 @@ void QCIsd_t_zcontract(double complex *e_tot,
         *e_tot += e;
 }
         free(permute_idx);
+        free(jobs);
 }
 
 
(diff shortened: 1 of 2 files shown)


sunqm (approved):

tberkel:
Hi Jonathan, just want to say thanks for your diligence in documenting, diagnosing, and fixing this bug. Much appreciated!

godotalgorithm:
No problem. I'm happy to contribute to the health of open-source software in chemistry. Also, I'm trying to generate some data on a deadline, and I decided that fixing problems in PySCF was likely to be faster than migrating my workflow to some other code that might very well have its own distinct set of problems waiting for me.

Issue #1517, opened 2022-11-22
Problems with FCI in Dooh symmetry


susilehtola:
I am trying to run some FCI calculations in Dooh symmetry. The following input
```
import numpy
from pyscf import gto, scf, fci

mol = gto.Mole()
# Singlet bond length 2.913
mol.build(
    atom = 'Li 0 0 0; Li 0 0 2.913',
    unit = 'angstrom',
    basis = 'cc-pVDZ',
    spin = 0,
    verbose = 4,
    symmetry = True,
    symmetry_subgroup = 'Dooh'
)
mf = scf.RHF(mol)
mf.kernel()

ci_energies = {}
for sym in ['A1g', 'A1u', 'E1ux', 'E1uy', 'E1gx', 'E1gy', 'E2gx', 'E2gy', 'E2ux', 'E2uy', 'E3ux', 'E3uy', 'E3gx', 'E3gy']:
    mci = fci.FCI(mol, mf.mo_coeff, singlet=False)
    mci.wfnsym = sym
    e, civec = mci.kernel(nroots=3)
    ci_energies[sym] = e
    print('{}: {}'.format(sym,e))
```
gives the following energies
```
A1g: [-14.90048097 -14.82656573 -14.80619407]
A1u: [-14.86012998 -14.83774646 -14.75794947]
E1ux: [-14.84879138 -14.80549729 -14.74183726]
E1uy: [-14.84879138 -14.80549729 -14.74183726]
E1gx: [-14.79618456 -14.78343722 -14.75798428]
E1gy: [-14.79618456 -14.78343722 -14.75798428]
E2gx: [-14.90048097 -14.82656573 -14.80619407]
E2gy: [-14.7461423  -14.74326066 -14.64015681]
E2ux: [-14.86012998 -14.83774646 -14.75794947]
E2uy: [-14.705979   -14.7043607  -14.66552235]
E3ux: [-14.84879138 -14.80549729 -14.74183726]
E3uy: [-14.84879138 -14.80549729 -14.74183726]
E3gx: [-14.79618456 -14.78343722 -14.75798428]
E3gy: [-14.79618456 -14.78343722 -14.75798428]
```
As you can see, there is a big problem in the E2 energies: 

- the x and y data are different
- `E2gx` results in the same energies as `A1g`
- `E2ux` results in the same energies as `A1u`
- the lowest Phi state `E3ux` is lower in energy than the Delta state `E2gy`

[...]

PR #1529, opened 2022-12-05
FCI wavefunction symmetry

sunqm:
* Support Dinfh and Cinv symmetry in FCI solver
* Fix issue #1517
* Adjust FCI initial guess

### DIFF
--- a/pyscf/fci/addons.py
+++ b/pyscf/fci/addons.py
@@ -417,8 +417,9 @@ def guess_wfnsym(ci, norb, nelec, orbsym):
         Irrep ID
     '''
     neleca, nelecb = _unpack_nelec(nelec)
-    strsa = numpy.asarray(cistring.make_strings(range(norb), neleca))
-    strsb = numpy.asarray(cistring.make_strings(range(norb), nelecb))
+    strsa = strsb = numpy.asarray(cistring.make_strings(range(norb), neleca))
+    if neleca != nelecb:
+        strsb = numpy.asarray(cistring.make_strings(range(norb), nelecb))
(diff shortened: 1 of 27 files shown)


sunqm:
@susilehtola This branch should be able to produce correct symmetry.

ndattani:
@sunqm I guess your commit from yesterday ([da8a25c](https://github.com/pyscf/pyscf/pull/1529/commits/da8a25c13211f48043c38ee9abf4783bfdc05c0c)) fixes the problem?

sunqm:
Yes. This PR is able to generate correct initial guess for any irreps of the cylindrical symmetry. For a few numbers of low lying states in each symmetry sector, the code in this branch can find the correct solution. There still exist numerical issues that may break the symmetry if computing too many excited states. Different techniques are needed to guarantee the symmetry adaption in H*c contraction. This is not solved by this PR.

fishjojo (approved):

ndattani:
Thanks! Let us test it on some Delta, Phi, Gamma and Eta states and see how far we can go before the numerical issues impede us :) I'm aware of a couple papers on Gamma states that can be used as a benchmark, but Eta states are an uncharted territory from an *ab initio* perspective, as far as I know (at least for molecules containing elements early in the periodic table). 

I'm still hoping to get past [this issue](https://github.com/sunqm/qcint/issues/17) with the basis set I want to use, perhaps by first seeing if I can get the calculation to work with `libcint` instead of `qcint`. Hopefully I'll be done that next week, then can start looking at some highly excited states, though @susilehtola  may have a head start (at least for the basis sets he's using).

Issue #1531, opened 2022-12-05
cubegen Cube.read reads xs, ys, zs wrong


tillhanke:
Hey there, I am currently trying to write and read multiple cube files.
In this I realized, that my written cube files are not readable by the same Cube object. Meaning, that the Cube object gets changed while reading a cube file.

To reproduce just use the following:

```python
from pyscf.tools import cubegen
from pyscf import gto
import numpy as np

mol = gto.M(atom='''O 0.00000000,  0.000000,  0.000000
                H 0.761561, 0.478993, 0.00000000
                H -0.761561, 0.478993, 0.00000000''', basis='6-31g*')
cube = cubegen.Cube(mol)
print(cube.nx, cube.xs)  # just to see the change
data = np.random.rand(cube.nx, cube.ny, cube.nz)
cube.write(data, '/tmp/test.cube')
cube.read('/tmp/test.cube')
print("\nNew Cube:\n",cube.nx, cube.xs)  # just to see the change
```
The read function changes xs, ys and zs to the incremental vectors written to the top of the cube file (lines 4 - 6).
The cube functions `get_coords` and the like assume, that these are not vectors but lists of parts of 1. Thus the object `cube` gets useless after a read call.

PR #1532, opened 2022-12-05
Read cubefile xs, ys, zs correctly

tillhanke:
* Read cubefile nx and use that to calculate xs (y and z as well)

* Fix issue #1531


tillhanke:
I am contributing via pull request basically for the first time ever. I hope I did this the right way.

Feedback would be very welcome ;)

pyscf/tools/cubegen.py
```diff
@@ -326,12 +326,23 @@ def read(self, cube_file):
...
+            def parse_nx(data, x):
+                from pyscf.pbc.gto import Cell
                 d = data.split()
-                return int(d[0]), numpy.array([float(x) for x in d[1:]])
-            self.nx, self.xs = parse_nx(f.readline())
-            self.ny, self.ys = parse_nx(f.readline())
-            self.nz, self.zs = parse_nx(f.readline())
+                nx = int(d[0])
+                x_vec = numpy.array([float(x) for x in d[1:]])
+                assert (numpy.delete(x_vec, x) == 0).all()
```
fishjojo, line 334: This is not true if the box is not rectangular.

tillhanke: But isn't the `Cube` object only usable with rectangular boxes?
I know, that cubefiles can theoretically use non rectangular boxes. But (as far as I know) other than throwing an error at this point the `Cube` object can't handle that.

Or am I missing something?

tillhanke: As a reference I have found 
https://github.com/pyscf/pyscf/blob/3c8c7875fdc0921ee4aec0338ca1ae3413971288/pyscf/tools/cubegen.py#L247-L250

fishjojo: For periodic systems, the box is the unit cell.
https://github.com/pyscf/pyscf/blob/3c8c7875fdc0921ee4aec0338ca1ae3413971288/pyscf/tools/cubegen.py#L241-L242

tillhanke: thanks for the input! I think I fixed that now.

pyscf/tools/cubegen.py
```diff
@@ -326,12 +326,23 @@ def read(self, cube_file):
             data = f.readline().split()
             natm = int(data[0])
             self.boxorig = numpy.array([float(x) for x in data[1:]])
-            def parse_nx(data):
+            def parse_nx(data, x):
```
sunqm, line 329: the argument `x` is not needed

sunqm (approved):

PR #1550, opened 2022-12-25
PBC DF K-build with a smaller overhead

hongzhouye:
# Math

The currently implemented density matrix (DM)-based K-build in the PBC module
```
K(p,q) = \sum_{L,r,s} V(L,p,s) * V(L,q,r).conj() * D(s,r)
```
requires `O(N^3 * X)` cost in both CPU and storage, where `N` and `X` are the size of the AO and auxiliary basis sets, respectively. This algorithm can be accelerated by exploiting the low-rank structure of the DM, which leads to the MO-based algorithm
```
D = A * B.conj()
X(L,p,i) = \sum_{s} V(L,p,s) * A(s,i)
Y(L,q,i) = \sum_{s} V(L,q,r) * B(r,i)
K(p,q) = \sum_{L,i} X(L,p,i) * Y(L,q,i).conj()
```
that now scales as `O(N^2 * O * X)` in both CPU and storage cost, where `O` is the rank of the DM. When `D` is positive semidefinite, the asymmetric algorithm above becomes symmetric.

The speedup of the MO-based algorithm is roughly `N/O`.

For routine K-build tasks such as those in SCF and TD-SCF calculations, `O` is strictly `Nocc` and the reformulation is numerically exact. In more general cases where the DM is only approximately low-rank, truncation by singular values guarantees a controlled error in the K-matrix elements.

*Note that the algorithms implemented here are different from the [occ-RI-K algorithm](https://aip.scitation.org/doi/10.1063/1.4923369). The latter (1) gives a `K` matrix that is different from the normal one in the virtual-virtual block and (2) is best implemented in an integral-direct manner which requires a structural change of existing (RS)DF module.*

# Major changes

The following functions
- `get_k_kpts` from `pbc/df/df_jk.py`
- `get_jk` from `pbc/df/df_jk.py`

are updated with MO-based K-build algorithm outlined above using GDF/RSDF/MDF (only the short-range part for MDF), which complements the existing DM-based algorithm. Both the symmetric mode (via eigen decomposition) and the asymmetric mode (via SVD) are implemented, with the eigen/singular value cutoff set to `cell.precision*1e-2`. One of the three K-build schemes is automatically chosen by the following logic:

- If DM is PSD (hence must be Hermitian), use **symm MO-based**
- if DM is non-PSD (including non-Hermitian) but `rank(DM)*2 < Nao`, use **asymm MO-based**
- otherwise, use **DM-based**

The user can also skip the logic above and force the use of the original DM-based K-build by setting
```
mydf.force_dm_kbuild = True        # default is False
```

The validity of the new K-build schemes have been verified using ground-state HF (for symmetric mode) and excited-state CIS (for asymmetric mode) for both single and multiple k-points.

# Detailed cost analysis

The theoretical speedup of the MO-based K-build is `N/O` over the original DM-based one. The practical speedup is smaller than the theoretical value mainly due to I/O, but can still be a factor of 3 even for a DZ basis set.

Comparison of CPU and memory cost of different K-build
(`N`: AO basis size; `O`: rank of DM; `X`: aux basis size; `[X]`: block size of aux basis index)
| K-build scheme | CPU | Memory |
| --------------- | ---- | --------- |
| DM                     | `2 * N^3*X` | `2 * N^2*[X]` |
| MO, symm         | `2 * N^2*O*X` | `N^2*[X] + N*O*[X]` |
| MO, asymm       | `3 * N^2*O*X` | `N^2*[X] + 2 * N*O*[X]` |

# Minor changes

The timing of `get_j_kpts`, `get_k_kpts`, and `get_jk` have also been updated.

### DIFF
--- a/pyscf/pbc/df/df.py
+++ b/pyscf/pbc/df/df.py
@@ -131,6 +131,8 @@ class GDF(lib.StreamObject, aft.AFTDFMixin):
     # Call _CCGDFBuilder if applicable. _CCGDFBuilder is slower than
     # _RSGDFBuilder but numerically more close to previous versions
     _prefer_ccdf = False
+    # If True, force using denisty matrix-based K-build
+    force_dm_kbuild = False
 
     def __init__(self, cell, kpts=numpy.zeros((1,3))):
         self.cell = cell
@@ -160,7 +162,6 @@ class GDF(lib.StreamObject, aft.AFTDFMixin):
         self.blockdim = getattr(__config__, 'pbc_df_df_DF_blockdim', 240)
         self.linear_dep_threshold = LINEAR_DEP_THR
         self._j_only = False
-        self.force_dm_kbuild = False # use mo coeff for K-build if possible
 # If _cderi_to_save is specified, the 3C-integral tensor will be saved in this file.
         self._cderi_to_save = tempfile.NamedTemporaryFile(dir=lib.param.TMPDIR)
 # If _cderi is specified, the 3C-integral tensor will be read from this file
--- a/pyscf/pbc/df/df.py
+++ b/pyscf/pbc/df/df.py
@@ -160,6 +160,7 @@ class GDF(lib.StreamObject, aft.AFTDFMixin):
         self.blockdim = getattr(__config__, 'pbc_df_df_DF_blockdim', 240)
         self.linear_dep_threshold = LINEAR_DEP_THR
         self._j_only = False
+        self.force_dm_kbuild = False # use mo coeff for K-build if possible
 # If _cderi_to_save is specified, the 3C-integral tensor will be saved in this file.
         self._cderi_to_save = tempfile.NamedTemporaryFile(dir=lib.param.TMPDIR)
 # If _cderi is specified, the 3C-integral tensor will be read from this file
--- a/pyscf/pbc/df/df_jk.py
+++ b/pyscf/pbc/df/df_jk.py
@@ -768,12 +768,14 @@ def get_jk(mydf, dm, hermi=1, kpt=numpy.zeros(3),
         tspans[2] += tock - tick
 
         LpqR = LpqI = pLqR = pLqI = None
+
+    tick = numpy.asarray((logger.process_clock(), logger.perf_counter()))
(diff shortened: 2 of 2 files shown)


sunqm:
* When the dm is passed from scf driver, MO coefficients are normally attached to the dm matrix. So `hasattr(dm, 'mo_coeff')` can be checked before calling other treatments

hongzhouye:
> * When the dm is passed from scf driver, MO coefficients are normally attached to the dm matrix. So `hasattr(dm, 'mo_coeff')` can be checked before calling other treatments

@sunqm Sounds good I can restore that option.

hongzhouye:
> * When the dm is passed from scf driver, MO coefficients are normally attached to the dm matrix. So `hasattr(dm, 'mo_coeff')` can be checked before calling other treatments

@sunqm Hi Qiming, I just added support for mo-tagged dm for k-points case in `get_k_kpts`. I didn't change the single k-point version (i.e., `get_jk`) because the reshaping [here](https://github.com/pyscf/pyscf/blob/master/pyscf/pbc/scf/hf.py#L653) seems to always remove the tag. Please let me know what you think.

ndattani:
Regarding the three failed CI runs, it seems to be the same problem as described [here](https://github.com/pyscf/pyscf/pull/1580#issuecomment-1404508122). Perhaps someone with permission (I don't have it) to run CI actions can re-run those three failed ones.

mkrompiec:
Just curious - is the same strategy used in the Gamma-point PBC code?

mkrompiec:
Thanks @hongzhouye, I haven't read the code properly and didn't realise it covers also the Gamma-point calculation. In that case, I can't wait for it to be merged! :)

hongzhouye:
> Thanks @hongzhouye, I haven't read the code properly and didn't realise it covers also the Gamma-point calculation. In that case, I can't wait for it to be merged! :)

@mkrompiec Thanks for being interested in it! I will try to clean it up quickly.

mkrompiec:
Any chance this will be merged in 2.2? :)

sunqm (approved):

Issue #1563, opened 2023-01-05
kccsd "outcore" version not working


fishjojo:
PySCF version: [9bbfe84](https://github.com/pyscf/pyscf/commit/9bbfe84fbc575a10978aa029783db3821335268c)

Input:
```
from pyscf.pbc import gto, scf, cc

cell = gto.Cell()
cell.atom = """
Si  0.0 0.0 0.0
Si  1.3467560987 1.3467560987 1.3467560987
"""
cell.basis = 'gthszv'
cell.pseudo  = 'gth-pade'
cell.a = [[0.0, 2.6935121974, 2.6935121974], [2.6935121974, 0.0, 2.6935121974], [2.6935121974, 2.6935121974, 0.0]]
cell.verbose = 5
cell.max_memory=200
cell.build()

nk = [2,2,2]
kpts = cell.make_kpts(nk, with_gamma_point=True)

kmf = scf.KRHF(cell, kpts, exxdiv=None).density_fit()
kmf.kernel()

kcc = cc.KRCCSD(kmf)
kcc.kernel()
```

Output:
```
using HDF5 ERI storage
    CPU time for CCSD integral transformation    418.67 sec, wall time      9.99 sec
Init t2, MP2 energy (with fock eigenvalue shift) = -0.0522329613819593
    CPU time for init mp2      5.52 sec, wall time      0.33 sec
Init E_corr(CCSD) = -0.0522329613819593
cycle = 1  E_corr(CCSD) = -0.020076051595837  dE = 0.0321569098  norm(t1,t2) = 1.21987
    CPU time for CCSD iter     39.29 sec, wall time      8.97 sec
cycle = 2  E_corr(CCSD) = -0.277986995364487  dE = -0.257910944  norm(t1,t2) = 8.72543
    CPU time for CCSD iter     36.34 sec, wall time      7.97 sec
cycle = 3  E_corr(CCSD) = -0.112187689811058  dE = 0.165799306  norm(t1,t2) = 71.3722
    CPU time for CCSD iter     37.38 sec, wall time      8.24 sec
cycle = 4  E_corr(CCSD) = -0.0835189447058978  dE = 0.0286687451  norm(t1,t2) = 10.7327
    CPU time for CCSD iter     37.16 sec, wall time      8.29 sec
cycle = 5  E_corr(CCSD) = -0.0810260231511605  dE = 0.00249292155  norm(t1,t2) = 2.604
[...]

Issue #1470, opened 2022-10-17
KHF_ksymm changes number of electrons


maxnus:
In this NiO system, the number of electrons, as measured as the trace of the 1-DM, changes from the initial guess (which is correct), to a different number when calling `mf.kernel()`. In this example below, the output is:

```
N(elec) of system= 192.000000
N(elec) initial guess= 192.000000
N(elec) converged MF= 186.000000
```

This may be related to the fact that `KHF_ksymm` does not overwrite `get_occ` of the KHF base class -
however I'm quite confused how this would ever be working in the first place then.

[example.py](https://github.com/pyscf/pyscf/files/9800890/issue.py.txt)

fishjojo:
A fix has been added in this [branch](https://github.com/pyscf/pyscf/tree/kpoint_symm).

PR #1578, opened 2023-01-23
Various updates for k-point symmetry related stuffs

fishjojo:
This PR adds the following new features:
1. Add symmetry adapted crystalline GTOs for symmorphic space groups. For non-symmorphic space groups, the factor system for the projective representation is not 1, which can be added in the future if needed.
2. Allow using symmetry adapted GTOs in PBC SCF calculations.
3. Add k-point symmetry adapted KRCCSD. The current implementation, at the best case, reduces both the memory footprint and the CPU time by a factor of the order of the point group. This code is meant to be a template, and k-point symmetry can be applied to other methods following the same strategy. A benchmark is shown below (run on 16 CPU cores). To further improve the performance, we will need an efficient sparse tensor contraction engine which also considers space group symmetry.
image
image

In addition, this PR also fixes the following issues:
1. fix occupation number error in pbc.khf_ksymm.get_occ (issue #1470)
2. fix dimension error in pbc.df.df.CDERIArray (issue #1563)

### DIFF
--- a/pyscf/pbc/cc/kccsd_rhf.py
+++ b/pyscf/pbc/cc/kccsd_rhf.py
@@ -34,6 +34,7 @@ from pyscf.pbc.cc import kintermediates_rhf as imdk
 from pyscf.lib.parameters import LOOSE_ZERO_TOL, LARGE_DENOM  # noqa
 from pyscf.pbc.lib import kpts_helper
 from pyscf.pbc.lib.kpts_helper import gamma_point
+from pyscf.pbc.df import GDF, RSGDF
 from pyscf.pbc.df import df
 from pyscf import __config__
 
@@ -499,9 +500,11 @@ class RCCSD(pyscf.cc.ccsd.CCSD):
 
     def __init__(self, mf, frozen=None, mo_coeff=None, mo_occ=None):
         assert (isinstance(mf, scf.khf.KSCF))
-        pyscf.cc.ccsd.CCSD.__init__(self, mf, frozen, mo_coeff, mo_occ)
+        # mf.to_khf converts mf to a non-symmetry object
+        pyscf.cc.ccsd.CCSD.__init__(self, mf.to_khf(), frozen, mo_coeff, mo_occ)
         self.kpts = mf.kpts
-        self.khelper = kpts_helper.KptsHelper(mf.cell, mf.kpts)
+        self.khelper = kpts_helper.KptsHelper(mf.cell, mf.kpts,
+                                              init_symm_map=False)
         self.ip_partition = None
         self.ea_partition = None
         self.direct = True  # If possible, use GDF to compute Wvvvv on-the-fly
@@ -511,13 +514,14 @@ class RCCSD(pyscf.cc.ccsd.CCSD):
         self.keep_exxdiv = False
 
         keys = set(['kpts', 'khelper', 'ip_partition',
-                    'ea_partition', 'max_space', 'direct'])
+                    'ea_partition', 'max_space', 'direct',
+                    'keep_exxdiv'])
         self._keys = self._keys.union(keys)
         self.__imds__ = None
 
     @property
     def nkpts(self):
-        return len(self.kpts)
+        return getattr(self.kpts, 'nkpts', len(self.kpts))
 
     get_normt_diff = get_normt_diff
     get_nocc = get_nocc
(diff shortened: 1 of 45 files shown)


pyscf/pbc/symm/symmetry.py
```diff
@@ -100,6 +92,41 @@ def make_Dmats(cell, ops, l_max=None):
...
+        for x in range(3):
+            while True:
+                tmp = ft[:,x] * mesh1[x]
+                if (abs(tmp - tmp.round()) > tol).any():
+                    mesh1[x] = mesh1[x] + 1
+                else:
+                    break
+
+        if not return_mesh:
+            logger.warn(cell, 'Input mesh %s has lower symmetry than the lattice.\
```
sunqm, line 122: Consider to change format to?
```
logger.warn(cell, 'Input mesh %s has lower symmetry than the lattice.\n'
'Some of the symmetry operations will be removed.\n'
'Recommended mesh is %s.', mesh, mesh1)
```

pyscf/pbc/symm/group.py
```diff
@@ -64,244 +132,331 @@ def matrix(self, a):
...
+            inverse : (n,) array of ints
+                The indices to reconstruct `conjugacy_mask` from `classes`.
+        '''
+        _, idx = np.unique(self.conjugacy_mask, axis=0, return_index=True)
+        representatives = np.sort(idx)
+        classes = self.conjugacy_mask[representatives]
+        inverse = -np.ones((self.order), dtype=int)
+        diff = (self.conjugacy_mask[None,:,:]==classes[:,None,:]).all(axis=-1)
+        for i, a in enumerate(diff):
+            inverse[np.where(a==True)[0]] = i # noqa: E712
```
sunqm, line 307: why it's `np.where(a==True)`? Isn't it the same to `np.where(a)`?

pyscf/pbc/symm/group.py
```diff
@@ -64,244 +132,331 @@ def matrix(self, a):
...
-        prod_table = np.zeros((n,n), dtype=np.int32)
-        lookup = self.lookup_table()
-        for i, g in enumerate(self.elements):
-            for j, h in enumerate(self.elements):
-                prod = g @ h
-                prod_table[i,j] = lookup[hash(prod)]
-        return prod_table
+        if self._multiplication_table is None:
+            prod = self.elements[:,None] * self.elements[None,:]
+            _table = [self.hash_table[hash(gh)] for gh in prod.flatten()]
```
sunqm, line 251: Multiplication table requires elements to be unique. Is it possible collision in hash table?

sunqm (approved):

pyscf/pbc/cc/kccsd_rhf.py
```diff
@@ -950,12 +955,18 @@ def _init_df_eris(cc, eris):
...
+    ao_loc = None
+#    with df.CDERIArray(cc._scf.with_df._cderi) as cderi_array:
+#        for ki in range(nkpts):
+#            for kj in range(nkpts):
+#                Lpq = cderi_array[ki,kj]
+    for ki, kpti in enumerate(kpts):
+        for kj, kptj in enumerate(kpts):
+            kpti_kptj = np.array((kpti, kptj))
+            # This loader is compatible with the old GDF format
+            with df._load3c(cc._scf.with_df._cderi, 'j3c', kpti_kptj) as j3c:
```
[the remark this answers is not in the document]
fishjojo: It seems RSGDF (the one implemented by Hongzhou) uses the old GDF file format, and is not compatible with CDERIArray. We probably should update CDERIArray to make it also work for the old GDF file format.

pyscf/pbc/symm/group.py
fishjojo: fixed

pyscf/pbc/symm/symmetry.py
fishjojo: fixed

pyscf/pbc/lib/kpts.py
```diff
@@ -966,19 +992,83 @@ def make_gdf_kptij_lst_jk(self):
...
+            assert (table > -1).all()
+            self._addition_table = table
+        return self._addition_table
+
+    @property
+    def inverse_table(self, tol=KPT_DIFF_TOL):
+        if self._inverse_table is None:
+            diff = -self.kpts_scaled[:,None,:] - self.kpts_scaled[None,:,:]
+            diff = round_to_fbz(diff, tol=tol)
+            idx = np.where(np.sum(abs(diff), axis=-1) < tol)
```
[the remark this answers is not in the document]
fishjojo: updated

pyscf/pbc/symm/group.py
fishjojo: Added a sanity check function to make sure the elements form a group.

PR #1580, opened 2023-01-24
Fix issue 1577

xwang862:
CCSD(T) fails after reading FCIDUMP. When determining orbital symmetries, it attempts to rebuild overlap matrix with `mol.intor_symmetric('int1e_ovlp')` ([code](https://github.com/pyscf/pyscf/blob/c0c4dbcb97b5725b5bbec3257b78f18c63663c76/pyscf/symm/addons.py#L69)), which will return an empty array because `mol` has no atom information. However, we have a working `get_ovlp()` in the mean-field object recovered from FCIDUMP ([code](https://github.com/pyscf/pyscf/blob/c0c4dbcb97b5725b5bbec3257b78f18c63663c76/pyscf/tools/fcidump.py#L343)). 

This pull request simply gets overlap from `mycc._scf.get_ovlp()` and provides it to `label_orb_symm()`.

### DIFF
--- a/pyscf/cc/ccsd_t.py
+++ b/pyscf/cc/ccsd_t.py
@@ -137,8 +137,9 @@ def _sort_eri(mycc, eris, nocc, nvir, vvop, log):
     nmo = nocc + nvir
 
     if mol.symmetry:
+        ovlp = mycc._scf.get_ovlp()
         orbsym = symm.addons.label_orb_symm(mol, mol.irrep_id, mol.symm_orb,
-                                            eris.mo_coeff, check=False)
+                                            eris.mo_coeff, s=ovlp, check=False)
         orbsym = numpy.asarray(orbsym, dtype=numpy.int32) % 10
     else:
         orbsym = numpy.zeros(nmo, dtype=numpy.int32)


ndattani:
Thanks for this fix! It seems to essentially be what Ty suggested in this comment: 

> "when reading the FCIDUMP, it doesn't need to/doesn't have enough info to generate the AO overlap matrix. so it leaves it blank and when it tries to form `s_mo` it fails. It seems like this is a bug, as it should just be able to fallback to making `s_mo` using [`mf.get_ovlp`](https://github.com/pyscf/pyscf/blob/master/pyscf/tools/fcidump.py#L343) which would have been defined in `to_scf`."

That comment was [made here](https://mattermodeling.stackexchange.com/q/10277/5).

xwang862:
Oh I didn't notice Ty has already suggested this 😂

ndattani:
Interesting that it [failed to build h5py](https://github.com/pyscf/pyscf/actions/runs/4001217802/jobs/6867228812#step:4:289) in all three of the cp37-cp37m builds. The code change in this pull request shouldn't cause the building of h5py to fail.

hebrewsnabla:
> Interesting that it [failed to build h5py](https://github.com/pyscf/pyscf/actions/runs/4001217802/jobs/6867228812#step:4:289) in all three of the cp37-cp37m builds. The code change in this pull request shouldn't cause the building of h5py to fail.

The fail is due to the ci cannot find an aarch64 wheel for h5py 3.8. Lokks like a coincidence: the wheel was uploaded to pypi just a few hours after the ci run 🤣 . I suggest simply re-run it.

ndattani:
I tried to re-run those three failed actions, but it looks like it needs to be done by @xwang862, or maybe someone with higher-level privileges in this repository.

Issue #1609, opened 2023-02-23
Issue with x2c and using cc-pCVXZ basis sets in at least PySCF 1.7.6 and newer


iliamazin:
Hi all,

I've come across what I believe to be a bug when trying to run a relativistic exact two-component PySCF calculation using a core-valence correlation consistent basis set (cc-pCVDZ) at the SCF/mean-field level. I have included two input files that do (c6h4_load.py) and do not (c6h4_parse.py) have the bug I describe below:

When running a calculation in which I request to use the cc-pCVDZ basis set, the output file lists the correct number of primitive GTOs, and the initial energy in the SCF procedure is reasonable (approx. -230 a.u.), but then the first step of the SCF cycle has an energy that is two orders of magnitude greater (approx. -50000 a.u.) and the SCF procedure does not converge.

The issue only occurs if I load the cc-pCVDZ basis set as it is defined within PySCF. Two actions fix this problem:
1) Switching to the non-core-valence variant (cc-pCVDZ -> cc-pVDZ); SCF converges normally without any complications.

2) If I manually specify the SAME cc-pCVDZ basis set and use the 'parse' function from the 'gto' module, the same exact x2c mean-field calculation no longer produces these massive energy errors in the SCF energy and convergence is smooth and quick.

The exact same input files provided below, run in PySCF 1.5.3 (compiled) and PySCF 1.6.2 (compiled), DO NOT have this issue. So it seems like something that could have been introduced around version 1.7?

The versions of PySCF I have found to be displaying this behavior:
PySCF 1.7.6 (conda)
PySCF 2.0.0a (compiled)
PySCF 2.0.1 (conda)
PySCF 2.1.1 (conda)
PySCF 2.1.1 (compiled)
PySCF 2.1.1 (compiled w/ OMP_NUM_THREADS=1)

Please let me know if you need any additional info from me! Thanks in advance.

Best,
Ilia Mazin

[x2c_basis_set_test.zip](https://github.com/pyscf/pyscf/files/10818364/x2c_basis_set_test.zip)

ndattani:
Without downloading your .zip file (and extracting it, etc.), my questions are:
 - were you uncontracting the basis sets? For X2C calculations the basis sets are meant to be uncontracted (although there's also basis sets with names like cc-pVDZ-X2C which are contracted specifically for X2C).
[...]

PR #1620, opened 2023-03-10
Fix Mole.decontract_basis method (Fix issue #1609)

sunqm:


xubwa:
Hi Qiming, since you are fixing this function, it should be proper to bring up a related problem. In x2c, when this decontract function is called to generate contraction coeffs and unc mole object, basis contracted for different j value is not treated properly, either a decontract function for 2c basis might be  implemented separately.

sunqm:
This PR solves only the non-relativistic case. The j distinguished de-contraction may deserve a new feature request issue I think. Would you mind creating one with an example, so that others can find the problem?

xubwa:
Sure I can do that and probably with a fix.

fishjojo (approved):

Issue #1451, opened 2022-10-06
kMP2 with RSGDF failure


verena-neufeld:
When using RSGDF with kMP2, e.g. by modifying example `pbc/22-k_points_mp2.py` to use RSGDF density fitting, I get an error in line 201 of `pbc/mp/kmp2.py`, in `_init_mp_df_eris`, `RuntimeError: cderi data not generated or format incompatible`.

PR #1628, opened 2023-03-14
pbc.df.CDERIArray compatibility improved

sunqm:
This PR updates pbc.df CDERIArray class. It can read the cderi file which is generated with rs_density_fit() method. This PR can fix issue #1451


verena-neufeld:
Thanks @sunqm ! I have just tested this on example 22-k_points_mp2.py and it runs even when RSGDF is specified. Thanks for all these bug fixes!

fishjojo (approved):

PR #1632, opened 2023-03-15
fix mp scs test according to instructions in error message

verena-neufeld:
I've added the symmetry instructions in the cell build as given by the failing test's error message. Not sure specifying "symmorphic=False" for example is necessary, I just followed the error message here.


verena-neufeld:
Not sure what causes the mcscf test failures here

sunqm (approved):

PR #1635, opened 2023-03-16
lo/boys: Use same centers for r2 and r integrals, fix charge center evaluation

wangenau:


wangenau:
Please see issue https://github.com/pyscf/pyscf/issues/1636 for the context of this PR.

sunqm (approved):

PR #1670, opened 2023-03-25
Frequency integration grid scaling factor as input to RPA calculations

EgorTrushin:
The frequency integration grid used in RPA calculations is determined by two parameters: number of points and scaling factor. However, only the number of frequency points served as an input parameter that could be varied in the RPA calculation. I have added the possibility to vary the scaling factor as well.

### DIFF
--- a/pyscf/gw/rpa.py
+++ b/pyscf/gw/rpa.py
@@ -39,7 +39,7 @@ einsum = lib.einsum
 # core routines, kernel, rpa_ecorr, rho_response
 # ****************************************************************************
 
-def kernel(rpa, mo_energy, mo_coeff, Lpq=None, nw=None, x0=None, verbose=logger.NOTE):
+def kernel(rpa, mo_energy, mo_coeff, Lpq=None, nw=40, x0=0.5, verbose=logger.NOTE):
     """
     RPA correlation and total energy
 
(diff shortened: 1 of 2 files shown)


pyscf/gw/rpa.py
```diff
@@ -39,14 +39,14 @@
 # core routines, kernel, rpa_ecorr, rho_response
 # ****************************************************************************
 
-def kernel(rpa, mo_energy, mo_coeff, Lpq=None, nw=None, verbose=logger.NOTE):
+def kernel(rpa, mo_energy, mo_coeff, Lpq=None, nw=None, x0=None, verbose=logger.NOTE):
```
sunqm, line 42: `x0=0.5` by default ?

sunqm (approved):

pyscf/gw/rpa.py
EgorTrushin: x0=0.5 is the default value in the corresponding grid generation subroutine [def _get_scaled_legendre_roots(nw, x0=0.5)] and probably needs to be kept at that value to avoid breaking test calculations.

sunqm: That's what I mean. I guess I didn't make a clear statement. Since the subsequent function calls do not test if x has a valid value, I recommend x0=0.5 instead of x0=None here as the default value.

EgorTrushin: I added default values in kernels for both nw and x0 in rpa.py and urpa.py

Issue #470, opened 2019-12-18
pyscf parses sys.argv when setting max_memory on a Mole object


eljost:
Dear developers,
while writing & executing some unittests for my own code that uses pyscf I encountered the following issue:

Setting `max_memory` on a `mol` object leads to a call of `cmd_args()` in `[...]/pyscf/gto/mole.py`. When I execute such code with `pytest -m [some mark]` then pyscf will parse sys.argv[1:] in `cmd_args()` and the mark supplied to pytest with `-m [some mark]` will be interpreted as memory argument py pyscf.

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../anaconda3/lib/python3.6/site-packages/pyscf/gto/mole.py:2033: in build
    _update_from_cmdargs_(self)
../../../anaconda3/lib/python3.6/site-packages/pyscf/gto/mole.py:3103: in _update_from_cmdargs_
    opts = cmd_args.cmd_args()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    def cmd_args():
        '''
        get input from cmdline
        '''
        parser = argparse.ArgumentParser()
        parser.add_argument('-v', '--verbose',
                            action='store_false', dest='verbose', default=0,
                            help='make lots of noise')
        parser.add_argument('-q', '--quiet',
                            action='store_false', dest='quite', default=False,
                            help='be very quiet')
        parser.add_argument('-o', '--output',
                            dest='output', metavar='FILE', help='write output to FILE')
        parser.add_argument('-m', '--max-memory',
                            action='store', dest='max_memory', metavar='NUM',
                            help='maximum memory to use (in MB)')
    
        (opts, args_left) = parser.parse_known_args()
    
        if opts.quite:
            opts.verbose = pyscf.lib.logger.QUIET
    
        if opts.verbose:
            opts.verbose = pyscf.lib.logger.DEBUG
    
        if opts.max_memory:
>           opts.max_memory = float(opts.max_memory)
E           ValueError: could not convert string to float: 'pyscf'
```

Here is code that reproduces the issue.
```
[...]

PR #1673, opened 2023-03-27
Parsing of command line arguments disabled by default

mkrompiec:
Changed the default value of `parse_arg` to `False` in `Mole.build()` and `Cell.build()`.


mkrompiec:
Solves #470

fishjojo (approved):

sunqm:
What about using a global config to control whether to parse cli arguments?

sunqm:
> This would be more elegant and would preserve backward compatibility for those who use CLI parsing. But does anyone use this functionality?

We are not sure how people use the library. A sudden change of default setting without notification may break others workflow. In this case, it is not difficult for us to maintain the backward compatibility.

PR #1677, opened 2023-03-28
Allow Gaussian MM charges

fishjojo:
This PR allows to use Gaussian MM charges in QMMM calculations. The nuclear gradients w.r.t MM atoms are also included. This will be used for interfacing PySCF with LAMMPS.
The code was written by Chenghan @MoleOrbitalHybridAnalyst, and was modified by me.
Future work is to perform QMMM calculations with PBC.


fishjojo:
I just noticed that in libcint, `PTR_FRAC_CHARGE` and `PTR_ZETA` are reserved at the same location in `_atm`. This seems inconvenient when the charges are Gaussians with fractional total charges. Is it okay to assign them to two columns?

zc62:
@fishjojo Does this PR support MM gradients when `charge_model` is `'point'`?

fishjojo:
@zc62 yes, the 3 center integral will be less efficient than the 2 center integral. We can add the point charge version to the main code.

sunqm (approved):

Issue #1684, opened 2023-03-29
Issue with cmake during installation of pyscf


vandan-revanur:
Currently if cmake is not installed locally, the installation of pyscf does not work and gives the following error.

## To reproduce:

`pip install pyscf`

## Error

```
Processing /Users/vandanrevanur/personal/codes/chemistry/pyscf
  Preparing metadata (setup.py) ... done
Requirement already satisfied: numpy!=1.16,!=1.17,>=1.13 in /Users/vandanrevanur/.virtualenvs/pyscf/lib/python3.10/site-packages (from pyscf==2.2.0) (1.24.2)
Requirement already satisfied: scipy!=1.5.0,!=1.5.1 in /Users/vandanrevanur/.virtualenvs/pyscf/lib/python3.10/site-packages (from pyscf==2.2.0) (1.10.1)
Requirement already satisfied: h5py>=2.7 in /Users/vandanrevanur/.virtualenvs/pyscf/lib/python3.10/site-packages (from pyscf==2.2.0) (3.8.0)
Building wheels for collected packages: pyscf
  Building wheel for pyscf (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [7 lines of output]
      scipy>1.1.0 may crash when calling scipy.linalg.eigh. (Issues https://github.com/scipy/scipy/issues/15362 https://github.com/scipy/scipy/issues/16151)
      running bdist_wheel
      running build
      running build_ext
      Configuring extensions
      cmake -S/Users/vandanrevanur/personal/codes/chemistry/pyscf/pyscf/lib -Bbuild/temp.macosx-12-x86_64-cpython-310
      error: command 'cmake' failed: No such file or directory
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for pyscf
  Running setup.py clean for pyscf
Failed to build pyscf
Installing collected packages: pyscf
  Running setup.py install for pyscf ... error
  error: subprocess-exited-with-error
  
  × Running setup.py install for pyscf did not run successfully.
  │ exit code: 1
  ╰─> [9 lines of output]
      scipy>1.1.0 may crash when calling scipy.linalg.eigh. (Issues https://github.com/scipy/scipy/issues/15362 https://github.com/scipy/scipy/issues/16151)
      running install
[...]

PR #1685, opened 2023-03-29
Add cmake as a dependancy in the setup.py

vandan-revanur:
This PR fixes issue: #1684

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -157,6 +157,6 @@ setup(
     cmdclass={'build_ext': CMakeBuildExt},
     install_requires=['numpy>=1.13,!=1.16,!=1.17',
                       _scipy_version,
-                      'h5py>=2.7'],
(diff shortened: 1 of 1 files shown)


sunqm:
cmake is not an install dependency

vandan-revanur:
> cmake is not an install dependency
As far as I understand it is a dependancy. And without it the pyscf installation does not succeed.

sunqm:
cmake may be considered a setup dependency. For setup dependencies, setup.cfg or project.toml are often the best place to add the configurations.

Honestly I doubt if it cmake should be managed by python package systems. Cmake is not a python project and it is often available on many systems. Installing the cmake package hosted on pypi requires a bunch of extra dependency which may not be met on many platforms. This can cause unnecessary install fails when installing pyscf.

I'll close this PR until a more safe and clean solution is found to manage cmake in python.

PR #1702, opened 2023-04-13
Fix github action

sunqm:
codecov was removed from pypi


fishjojo:
codecov can still be used with actions: https://github.com/codecov/codecov-action

fishjojo (approved):

PR #1704, opened 2023-04-14
smearing for ROHF and fix_spin=True

mkrompiec:
For ROHF and spin>0, PySCF's smearing function produces incorrect results (and drives spin to 0), because it assumes the meaning of the occupation vector as in RHF. I propose a solution where the smearing function is applied separately to alpha and beta occupations which are later added up to produce the ROHF occupation vector. This is obviously wrong for low multiplicities and high smearing sigma, but it's "almost" accurate for high enough multiplicity and low enough sigma.
This is my test case:
```
from pyscf import gto, scf
from pyscf.scf.addons import smearing_

mol = gto.M(
    atom = '''
Fe      -1.301259682     -0.271152073      0.000000000
Fe       0.741101210     -0.199078285      1.084788423
Fe       0.741101210     -0.199078285     -1.084788423

''',basis="lanl2dz",ecp="lanl2dz",symmetry=False,unit='Angstrom',
            spin=4,charge=0,verbose=4)
myhf = scf.ROHF(mol)
myhf.max_cycle = 600 

smearhf = smearing_(myhf, sigma=0.005, fix_spin=True)
smearhf.kernel()
```

### DIFF
--- a/pyscf/pbc/scf/addons.py
+++ b/pyscf/pbc/scf/addons.py
@@ -57,7 +57,7 @@ def project_mo_nr2nr(cell1, mo1, cell2, kpts=None):
 
 def smearing_(mf, sigma=None, method=SMEARING_METHOD, mu0=None, fix_spin=False):
     '''Fermi-Dirac or Gaussian smearing'''
-    from pyscf.scf import uhf
+    from pyscf.scf import uhf, rohf
     from pyscf.scf import ghf
     from pyscf.pbc.scf import khf
     mf_class = mf.__class__
@@ -65,9 +65,12 @@ def smearing_(mf, sigma=None, method=SMEARING_METHOD, mu0=None, fix_spin=False):
     is_ghf = isinstance(mf, ghf.GHF)
     is_rhf = (not is_uhf) and (not is_ghf)
     is_khf = isinstance(mf, khf.KSCF)
+    is_rohf = isinstance(mf, rohf.ROHF)
+    if is_rohf:
+        is_rhf = False 
 
-    if fix_spin and not is_uhf:
-        raise KeyError("fix_spin only supports UHF.")
+    if fix_spin and not (is_uhf or is_rohf):
+        raise KeyError("fix_spin only supports UHF and ROHF.")
     if fix_spin and mu0 is not None:
         raise KeyError("fix_spin does not support fix mu0")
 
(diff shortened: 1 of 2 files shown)


fishjojo:
I don't think smearing should be applied to ROHF, and the idea of `fix_spin` was merely to conserve the crystal symmetry in UHF calculations. The proposed solution can be a workaround if one really want smearing for ROHF while not altering the spin too much, but this may fail in many cases. @sunqm what do you think?

mkrompiec:
Fix_spin preserves the multiplicity defined in the Mole object and smearing improves convergence and makes the result less dependent on the initial guess for high-spin systems. I agree that smearing in ROHF is not well defined (because we have only one occupation number vector, being the sum of alpha and beta occupations), but for sufficiently high multiplicities this is approximately correct.

sunqm:
> I don't think smearing should be applied to ROHF, and the idea of `fix_spin` was merely to conserve the crystal symmetry in UHF calculations. The proposed solution can be a workaround if one really want smearing for ROHF while not altering the spin too much, but this may fail in many cases. @sunqm what do you think?

I don't quite understand why fix_spin is inappropriate for ROHF. I always considered ROHF as a special UHF, which has an extra constraints on orbital degeneracy. If smearing can be applied to molecular UHF, seems the same can be done on ROHF. 

Actually, the issue in smearing I'm concerned is whether it is a right method for molecular calculations at first place. If smearing for molecules is acceptable, the proposed solution for ROHF is acceptable IMO

fishjojo:
Because unless you allow different occupations for alpha and beta spin
orbitals, which I think is not the case in ROHF, fix_spin won’t actually
fix the spin.

Qiming Sun ***@***.***>于2023年4月29日 周六下午4:05写道：

mkrompiec:
But it does fix the spin for ROHF, with my modifications to the smearing
function (please see for yourself). If the multiplicity is high enough and
electronic temperature low enough (both conditions not difficult to meet),
alpha and beta occupations are, for all intents and purposes, effectively
independent in ROHF.

fishjojo:
Please fix the lint error, and please also add a small test. Thanks.

mkrompiec:
Thanks, done!

mkrompiec:
The test demonstrates usefulness of this trick for molecules. 
First, I run a ROHF calculation on Fe2 (spin=6) and get E=-244.9642035. 
Then I run ROHF with smearing and use the computed RDM as the initial guess to another ROHF (without smearing), yielding a lower energy than the 1st one: -244.9808750.

mkrompiec:
Very strange, the result of the non-smeared ROHF calculation depends on the python version.

mkrompiec:
I hope the py3.7 tests pass now, I'm unable to run them on my computer (I can't install scipy on py3.7 on Mac M1).

mkrompiec:
Apologies, I guess I misunderstood the rounding off in assertAlmostEqual. The test now passes in my local python 3.9 environment on Mac m1 and should pass on CI.

fishjojo (approved):

PR #1706, opened 2023-04-15
Support complex-valued h1e in fci_slow.absorb_h1e

kevinsung:
Makes `fci_slow.absorb_h1e` accept a complex-valued `h1e`. I made the same change to `direct_nosym.absorb_h1e`. May I ask, why do both functions exist? They appear identical.

### DIFF
--- a/pyscf/fci/direct_nosym.py
+++ b/pyscf/fci/direct_nosym.py
@@ -118,7 +118,7 @@ def absorb_h1e(h1e, eri, norb, nelec, fac=1):
     '''
     if not isinstance(nelec, (int, numpy.number)):
         nelec = sum(nelec)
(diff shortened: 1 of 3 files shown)


sunqm (approved):

sunqm:
Probably one is copied from the other one. It was there for clarity at beginning since fci_slow.py was not really considered as a submodule of fci module. Those *_slow.py files were put in the source code to demonstrate how the relevant modules were implemented. Maybe these files should be moved to examples, or somewhere in the documentation in the future.

kevinsung:
Got it. Yeah, the only reason I'm using `fci_slow` is that `fci_slow.contract_2e` supports complex two-body tensors.

PR #1734, opened 2023-05-15
Implementation of APC active space selection

dking072:
This branch implements the "approximate pair coefficient" (APC) ranked-orbital scheme for active space selection, developed in the paper

A Ranked-Orbital Approach to Select Active Spaces for High-Throughput Multireference Computation
https://doi.org/10.1021/acs.jctc.1c00037

It was recently used (and developed further) to automatically benchmark multiconfigurational vertical excitations with PySCF:

Large-Scale Benchmarking of Multireference Vertical-Excitation Calculations via Automated Active-Space Selection
https://doi.org/10.1021/acs.jctc.2c00630

I have implemented the active space selection scheme trying to follow the conventions of mcscf/avas.py, which is a comparable contribution.

mcscf/apc.py contains the code implementing the method
mcscf/test/test_apc.py contains unit tests
examples/mcscf/80-apc.py contains minimal working examples

I believe this will be a helpful contribution for users wishing to utilize the mcscf module and associated methods.


pyscf/mcscf/apc.py
```diff
@@ -0,0 +1,410 @@
...
+        condition1 = (nactel > 0)
+        condition2 = (nactel < 2*len(active_idx))
+        condition3 = (len(active_idx) >= num_os)
+        
+        if (condition1 and condition2 and condition3):
+            return True
+        else:
+            self.log.debug("Active space is not reasonable!")
+            if not condition1:
+                self.log.debug("Condition 1 not met")
```
sunqm, line 96: logger print also `nactel, len(active_idx), num_os` as the reason why conditions are not met.

pyscf/mcscf/apc.py
```diff
@@ -0,0 +1,410 @@
...
+        assert(len(entropies) == orbs.shape[1])
+        
+        self.log = logger.new_logger(lib.StreamObject,verbose)
+        self.orbs = orbs
+        self.occ = np.array(occ)
+        self.entropies = entropies
+        self.max_size = max_size
+        self.verbose = verbose
+        
+    def _ncsf(self,nactel,norbs):
```
sunqm, line 65: What's the difference between _ncsf and _calc_ncsf? It's better to have some comments here to explain the difference

pyscf/mcscf/apc.py
```diff
@@ -0,0 +1,410 @@
...
+        #Check that we have a full set of orbitals:
+        assert(orbs.shape[0] == orbs.shape[1])
+        assert(len(occ) == len(entropies))
+        assert(len(entropies) == orbs.shape[1])
+        
+        self.log = logger.new_logger(lib.StreamObject,verbose)
+        self.orbs = orbs
+        self.occ = np.array(occ)
+        self.entropies = entropies
+        self.max_size = max_size
```
sunqm, line 62: ensure data type
```
self.entropies = np.asarray(entropies)
```

pyscf/mcscf/apc.py
```diff
@@ -0,0 +1,410 @@
...
+            tries = 0
+
+            while not active_space_is_reasonable:
+                try:
+                    dropped_idx = ranked_active_idx[tries] #Move to next possibility
+                    dropped_idx_entropy = np.round(entropies[dropped_idx],4)
+                    dropped_idx_occ = int(occ[dropped_idx])
+                except IndexError:
+                    log.error("Not enough orbitals to choose a reasonable active space!")
+                    assert(active_space_is_reasonable)
```
sunqm, line 155: active_space_is_reasonable will never be False when reaching this line. What is the reason to assert here?

pyscf/mcscf/apc.py
```diff
@@ -0,0 +1,410 @@
...
+        if len(os_idx) > 0:
+            log.info("Singly occupied orbitals found, setting them to have entropy of max + 0.01...")
+        
+        #Start with all orbitals in active space:
+        active_idx = list(range(len(entropies)))
+        inactive_idx = []
+        secondary_idx = []
+        
+        #Size constraint:
+        if isinstance(self.max_size,tuple):
```
sunqm, line 120: Including other iterables `isinstance(self.max_size, (tuple, list, np.array))` ?

pyscf/mcscf/apc.py
```diff
@@ -0,0 +1,410 @@
...
+    >>> myapc = apc.APC(mf,max_size=2)
+    >>> ncas,nelecas,casorbs = myapc.kernel()
+    >>> mc = mcscf.CASSCF(mf, ncas, nelecas).run(casorbs)
+    """
+
+    def __init__(self,mf,max_size=(8,8),n=2,eps=1e-3,verbose=4):
+        self.log = logger.new_logger(lib.StreamObject,verbose)
+        self.mf = mf
+        self.n = n
+        self.eps = eps
```
sunqm, line 249: check `eps > 0` ?

pyscf/mcscf/apc.py
dking072: Done

pyscf/mcscf/apc.py
dking072: Added an assertion statement 👍

pyscf/mcscf/apc.py
dking072: Done.

pyscf/mcscf/apc.py
dking072: I think it should always be False here, so it's meant as a way to stop execution. Have replaced with
"raise RuntimeError("Not enough orbitals to choose a reasonable active space!")"

pyscf/mcscf/apc.py
dking072: Done.

pyscf/mcscf/apc.py
dking072: Have added comments 👍

pyscf/mcscf/apc.py
dking072: Added 👍

dking072:
In addition, I needed to update the test and example files as the geometries I needed to update the units of the input molecules to bohr instead of angstroms.

dking072:
Hi, looking for an update on this PR?

dking072:
I made it so that singly occupied orbitals are explicitly contribute alpha electrons in the chosen active space. Previously this had equivalent behavior to the old implementation

nactel = int(np.sum(np.array(occ)[active_idx]))
alpha = int(nactel//2 + nactel%2)
beta = int(nactel//2)

for scfs with 0 or 1 singly occupied orbitals, but the old code would choose something like a ((2,1),3) active space if it chose the 3 singly occupied orbitals which seemed confusing (even though this wouldn't effect calculations as spin symmetry is enforced in the mc calculation).

sunqm (approved):

PR #1746, opened 2023-05-27
Implement interface to set custom parameters of XC functionals in Libxc

Dayou-Zhang:
In this pull request, I added a callback interface in `libxc_itrf.c` to allow using custom parameters in XC functionals. The user can adjust the parameters in the `xc_func_type` struct right before Libxc evaluates the functional through the callback interface. A fully commented example is supplied.


susilehtola:
This is not the way to set libxc parameters. You are not supposed to touch the Libxc internal arrays. Also, your example is hardcoded to a single functional, which sets a bad example. We have generic functionality to do this. Use the function `xc_func_set_ext_params` to set the external parameters. Moreover, you need to set the external parameters not just where the functions is evaluated, but also where PySCF queries various properties of the functional, such as exact exchange coefficients. At present these modifications will yield incorrect behavior.

Several people have expressed interest in being able to set libxc functional parameters in PySCF. For instance, I have been in touch with @ychen428 since early 2023 who has a proper interface to the external parameter setters. Per my request, @ychen428 promised to open up a pull request "shortly" several months ago, but I don't see one up yet. @ychen428 please open up a pull request; like we discussed by email the interface may need some discussion with the PySCF main developers since it's not obvious what is the best way to store the parameters in PySCF.

Dayou-Zhang:
Thanks for pointing me towards the right direction. I didn't know Libxc has such a convenient method to set external parameters.

I made a new implementation of the interface to set external parameters. This should be much cleaner than what is previously done. In the new implementation, no manual adjustment of HF exchange percentage is needed.

susilehtola:
> Moreover, you need to set the external parameters not just where the functions is evaluated, but also where PySCF queries various properties of the functional, such as exact exchange coefficients. At present these modifications will yield incorrect behavior.

This is still missing.

Dayou-Zhang:
In the current implementation, the external parameters are set when PySCF queries properties of the functional.

Here is a summary of how this is achieved. In `LIBXC_hybrid_coeff`, `LIBXC_nlc_coeff`, `LIBXC_rsh_coeff`, and `LIBXC_eval_xc`, the `xc_func_init` function call is replaced by `_xc_func_init`, which sets the custom parameters after the `xc_func_type` struct is initialized. Therefore, when PySCF queries the parameters such as the HF exchange percentage, the library will return the updated values. I also added an option to clear the lru cache so that PySCF will call the library again to read the updated values after the external parameters have changed.

Can you please let me know which part I am missing?

pyscf/lib/dft/libxc_itrf.c
```diff
@@ -362,6 +363,124 @@ static void _eval_xc(xc_func_type *func_x, int spin, int np,
...
 }
 
+typedef struct {
+        int id;
+        double *params;
+        size_t size;
+} ext_params;
+
+static ext_params *custom_ext_params = NULL;
+static int custom_ext_params_count = 0;
```
susilehtola, line 373: A few issues here. First, the name of the struct is too close to libxc's ext_params and the two can be confused. The struct isn't used for the external parameters, instead it is used as temporary storage.

`custom_ext_params_count` is also a bad variable name: it's not a counter but the size of the custom_ext_params array.

```suggestion
typedef struct {
    int id; /* LibXC functional ID */
    double *ext_params; /* LibXC functional external parameters */
    size_t N; /* Number of parameters */
} ext_params_cache_t;

/* Storage array for the external parameters */
static ext_params_cache_t *ext_params_cache = NULL;
static int ext_params_cache_size = 0;
```

pyscf/lib/dft/libxc_itrf.c
```diff
@@ -362,6 +363,124 @@ static void _eval_xc(xc_func_type *func_x, int spin, int np,
...
+
+int set_ext_params(int id, double *params, size_t size)
+{
+        // Update existing entry if exists
+        for (int i = 0; i < custom_ext_params_count; i++) {
+                ext_params *p = &custom_ext_params[i];
+                if (p->id == id) {
+                        if (p->size != size) {
+                                fprintf(stderr, "Inconsistent parameter size. Expected %lu but got %lu\n", p->size, size);
```
susilehtola, line 382: "Parameter size" is misleading, say "number of parameters"

pyscf/lib/dft/libxc_itrf.c
```diff
@@ -362,6 +363,124 @@ static void _eval_xc(xc_func_type *func_x, int spin, int np,
...
+int remove_ext_params(int id)
+{
+        for (int i = 0; i < custom_ext_params_count; i++) {
+                if (custom_ext_params[i].id == id) {
+                        free(custom_ext_params[i].params);
+                        for (int j = i; j < custom_ext_params_count - 1; j++) {
+                                custom_ext_params[j] = custom_ext_params[j + 1];
+                        }
+                        custom_ext_params_count--;
+                        return 0;
```
susilehtola, line 425: The deallocater isn't working properly. You also need to reallocate the cache size.

pyscf/lib/dft/libxc_itrf.c
```diff
@@ -362,6 +363,124 @@ static void _eval_xc(xc_func_type *func_x, int spin, int np,
...
+                        printf("%3i %14.6e %8s %s\n", j, p->params[j],
+                                xc_func_info_get_ext_params_name(func.info, j),
+                                xc_func_info_get_ext_params_description(func.info, j));
+                }
+                xc_func_end(&func);
+                putchar('\n');
+        }
+}
+
+static int _xc_func_init(xc_func_type *func, int xc_id, int spin)
```
susilehtola, line 466: This is also named too closely to `xc_func_init`, it should be clearer that this is a function in the PySCF wrapper. Use e.g. `xc_init_wrapper()`

susilehtola:
> Can you please let me know which part I am missing?

I do see now that the minute change of adding `_` in the function calls changes the behavior. This is not sufficiently clear. I left other comments on code clarity. Please also fix your C indentation to a tab width of 4 characters instead of 8.

pyscf/lib/dft/libxc_itrf.c
Dayou-Zhang: Fixed according to suggestions. However, I think the name `params` is clear enough in the context; we don't need to change it to `ext_params`.

pyscf/lib/dft/libxc_itrf.c
Dayou-Zhang: Fixed

pyscf/lib/dft/libxc_itrf.c
Dayou-Zhang: Fixed

pyscf/lib/dft/libxc_itrf.c
Dayou-Zhang: I renamed it as `init_xc_and_set_ext_params()`. This should be clear and distinguishable from Libxc functions.

Dayou-Zhang:
Thanks for the suggestions. I have fixed the issues mentioned above.

> Please also fix your C indentation to a tab width of 4 characters instead of 8.

I used 8-space indentation because `libxc_itrf.c` used 8-space indentation before I made any changes. To make the whole file consistent, I reformatted the entire file with 4-space indentation in my last commit.

sunqm:
Modifying the internal parameters of libxc through PySCF's libxc interface IMO is not a good way to use libxc. The pyscf libxc interface is supposed to combine different libxc functionals at Python level. In this scenario, it is better to use pylibxc library and call its APIs to modify functionals. An adapter may need to be implemented for calling pylibxc in the dft module.

susilehtola:
@sunqm an issue is that pylibxc is overengineered in terms of an interface and we are thinking about disbanding it in favor of a more minimal interface similar to the C interfacing used in PySCF...

However, it is true that many codes will want a more abstract, object-oriented interface to Libxc; this could be done by another project in a higher-level language like C++ with Python interfaces.

Dayou-Zhang:
Here is a new implementation where the external parameters are cached in Python. The parameters are cached in the `NumInt` object. Parameter editing can be done via a `dict`-like interface accessible from `mf.ext_params_dict`. When user sets a new external parameter, an array of struct `ext_params_cache_t` is created via `ctypes`. The struct array is passed to the C library as a pointer while calling `LIBXC_hybrid_coeff`, `LIBXC_nlc_coeff`, `LIBXC_rsh_coeff`, and `LIBXC_eval_xc`. This implementation will not set parameters persistently and allows multiple `mf` objects with different external parameters set.

I made modifications in the base classes `KohnShamDFT` and `_NumIntMixin`. Hopefully this would work for all variations of DFT, but I have not done a thorough test yet.

sunqm:
This feature looks make the interface to libxc unnecessarily complicated. I think it needs more consideration whether and how to handle this feature. I will close the PR for the moment.

PR #1754, opened 2023-06-05
Fixed the numerical issue in df hessian when the overlap matrix is ill-conditioned

wxj6000:
- The overlap matrix of the auxiliary basis is highly ill-conditioned, especially when omega is not zero. The ill-condition error was not raise with np.linalg.inv. Using pinv instead.
- The test cases are updated accordingly. Because of this improvement, the test case of range-separated hybrid functional with auxbasis_response = 2 can pass. And the agreement with hessian w/o df is improved from 1e-3 to 1e-4.


susilehtola:
It looks like `df/incore.py` similarly employs `numpy.linalg.inv(j2c)` which gives the wrong result when the auxiliary basis is not well-conditioned. That bug should also be fixed, no?

wxj6000:
> 

`numpy.linalg.inv(j2c)` only appears in the test section of `df/incore.py`. The overlap matrix in that test is well-conditioned I think. In other places, it is not necessary to calculate the inverse of j2c explicitly (as far as I know). One can solve a linear equation with cholesky decomposition. [Scipy](https://github.com/scipy/scipy/blob/v1.10.1/scipy/linalg/_basic.py#L359) will raise an error if the matrix is ill-conditioned. Then it is switched to an [eigh solver](https://github.com/pyscf/pyscf/blob/master/pyscf/df/incore.py#LL219C9-L219C9).

fishjojo (approved):

Issue #1762, opened 2023-06-12
CASSCF kernel function impurity leads to persistent _max_stepsize reduction in repeated kernel calls


MatthewRHermes:
The following:
```
from pyscf import gto, scf, mcscf, lib
xyz='''C 0 0 0,
       O 1 0 0'''
mol = gto.M (atom=xyz, basis='cc-pvdz', symmetry=False, verbose=lib.logger.DEBUG, output='CO.log')
mf = scf.RHF (mol).run ()
mo = mf.mo_coeff.copy ()
mc = mcscf.CASSCF (mf, 8, 10).run () # default convergence requested
print ("mc.converged =", mc.converged, "; mc.e_tot =", mc.e_tot, '; mc._max_stepsize =', mc._max_stepsize)
mc.run (mo, conv_tol_grad=1e-16) # impossibly tight convergence requested
print ("mc.converged =", mc.converged, "; mc.e_tot =", mc.e_tot, '; mc._max_stepsize =', mc._max_stepsize)
mc.run (mo, conv_tol_grad=None) # default convergence requested
print ("mc.converged =", mc.converged, "; mc.e_tot =", mc.e_tot, '; mc._max_stepsize =', mc._max_stepsize)
```
results in the following:
```
overwrite output file: CO.log
mc.converged = True ; mc.e_tot = -112.81902947003246 ; mc._max_stepsize = 0.02
mc.converged = False ; mc.e_tot = -112.81902947761458 ; mc._max_stepsize = 6.565139347890731e-25
mc.converged = False ; mc.e_tot = -112.74869243452127 ; mc._max_stepsize = 2.742069470056062e-39
```
That is, the third calculation, which should be identical to the first, is unable to converge, due to the effects of the `_max_stepsize` member variable which is gradually constricted to essentially zero over the course of the futile second calculation. The issue is resolved by resetting `_max_stepsize` manually:
```
from pyscf import gto, scf, mcscf, lib
xyz='''C 0 0 0,
       O 1 0 0'''
mol = gto.M (atom=xyz, basis='cc-pvdz', symmetry=False, verbose=lib.logger.DEBUG, output='CO.log')
mf = scf.RHF (mol).run ()
mo = mf.mo_coeff.copy ()
mc = mcscf.CASSCF (mf, 8, 10)
print ("The initial value of mc._max_stepsize is", mc._max_stepsize)
mc.run () # default convergence requested
[...]

PR #1763, opened 2023-06-12
Reset CASSCF stepsize scheduler in kernel

MatthewRHermes:
Address an impurity in the CASSCF kernel functions by resetting the class member _max_stepsize (issue #1762). This variable is set to a number that gets smaller and smaller as kernel proceeds, and if it's too small then nothing will ever converge.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -81,7 +81,7 @@ Kyle Bystrom
 Minye Zhang
 Yusuke Kimura
 Unathi Koketso Skosana
-Matthew Hennefarth
+Matthew R Hennefarth (University of Chicago)
 Will Wheeler
 Basil Ibrahim
 Junjie Yang
--- a/pyscf/df/grad/rhf.py
+++ b/pyscf/df/grad/rhf.py
@@ -244,7 +244,7 @@ def get_j(mf_grad, mol=None, dm=None, hermi=0):
 
     # (P|Q)
     int2c = auxmol.intor('int2c2e', aosym='s1')
-    rhoj = scipy.linalg.solve(int2c, rhoj.T, sym_pos=True).T
+    rhoj = scipy.linalg.solve(int2c, rhoj.T, assume_a='pos').T
     int2c = None
 
     # (d/dX i,j|P)
--- a/pyscf/lib/CMakeLists.txt
+++ b/pyscf/lib/CMakeLists.txt
@@ -18,7 +18,6 @@ project (pyscf)
 if (NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE RELWITHDEBINFO)
 endif()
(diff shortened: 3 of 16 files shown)


MatthewRHermes:
Here I choose to explicitly reinitialize `_max_stepsize` as close as possible to the scheduler calls in all relevant kernel functions (even though, as I say in #1762, I suspect the "problem" is a typo in only one of them), because I think it's better to just be 100% certain that this is always being reset.

sunqm:
What about adding a reset method in CASSCF class and placing these initialization there? This method calls the reset method of CASCI class and cleanup attributes like `_max_stepsize`. Also, the reset method should be called in the CASSCFS_Scanner .

MatthewRHermes:
CASCI.reset calls _scf.reset, which erases the cached _eri array, which is necessary for potential energy scans or geometry optimizations, but doesn't that throw the baby out with the bathwater for the case in #1762? To fix that example, CASSCF.reset would have to be called in every kernel run, which doesn't look like how reset is used elsewhere.

Now that I'm saying this, why is casscf._max_stepsize an attribute of casscf at all, given the way it's used? Especially since the scheduler's return is assigned to `max_stepsize` in the local namespace of the calling kernel function? Wouldn't it be more parsimonious to just compare `self.max_stepsize` directly to `envs['max_stepsize']` in the scheduler?

MatthewRHermes:
I suppose the floating max_stepsize is convenient to keep around as metadata, but one could still tag it in a "read-only" sort of way, comparing `self.max_stepsize` to `envs['max_stepsize']` to set its value and then only copying that value to `self._max_stepsize`.

MatthewRHermes:
Like, how about this? The scheduler tags the constricted _max_stepsize on the method instance, and a very advanced user can use this information, but the scheduler itself manipulates the local max_stepsize from the caller scope directly, so that the tag is implicitly reset every kernel call without needing to write an initialization line or a reset function or anything else. Also, the UCASSCF scheduler function imports the RCASSCF function so they will always be identical.

MatthewRHermes:
I suspect that GCISD test failure for Python 3.11 is unrelated to this PR.

MatthewRHermes:
Since @sunqm asked, I also implemented a reset function that additionally resets the _max_stepsize tag. I also took the opportunity to make sure UCASSCF follows the RCASSCF behavior in the scanner and to make sure that CASCI reset gets called in its as_scanner method.

MatthewRHermes:
I'll just add that one reason I like my current solution is that it moves the CASSCF method class closer to the standard that

> Intermediate status at runtime are **not** supposed to be saved in the method class (in contrast to the object oriented paradigm).

From [here](https://pyscf.org/code-rule.html#api-conventions).

sunqm (approved):

PR #1765, opened 2023-06-13
KGHF Mulliken analysis

pavel-po:
Implementation of a Mulliken analysis for KGHF, which is useful for understanding characters of the found solutions in solids. The implementation is consistent with KUHF and molecular GHF Mulliken analyses and largely follows them.


pavel-po:
> Please add a test to run the new function

The existing test function for mf.analyze() is extended to include the KGHF case.

PR #1770, opened 2023-06-20
Bug fix for PBC DF

hongzhouye:
Two bugs are fixed in this PR:
1. In PR #[1620](https://github.com/pyscf/pyscf/pull/1620), the default behavior of Mole was changed to sort the input basis sets by angular momentum for each atom, which makes `mol._bas` and `mol._basis` not necessarily agree with each other. The mismatch causes bugs in various places in `pbc/df/rsdf_helper.py`, leading to inaccurate DF integrals when using RSDF. This PR fixes the bug. A minimal demo is provided below.
2. The PBC DF-based J-build has a bug in [L129](https://github.com/pyscf/pyscf/blob/master/pyscf/pbc/df/df_jk.py#L129). When an non-Hermitian density matrix is input, the J matrix is not Hermitian and aosymm must be False in the DF loop function. This leads to failure in e.g., k-based TDDFT calculations. This PR fixes the bug.

PBE total energy for water in a box (details below) with different DF methods.
| DF methods | Energy |
| ------------ | ------- |
| RSDF (before fix) | -76.3338831107 |
| RSDF (after fix) | -76.3338642530 |
| GDF (RS builder) | -76.3338642530 |
| GDF (CC builder) | -76.3338642530 |

``` Python
atom = '''
O          0.00000        0.00000        0.11779
H          0.00000        0.75545       -0.47116
H          0.00000       -0.75545       -0.47116
'''
basis = 'cc-pvdz'
a = np.eye(3) * 6
kmesh = [3,1,1]
cell = gto.M(atom=atom, a=a, basis=basis)
cell.precision = 1e-12
```

### DIFF
--- a/pyscf/pbc/df/df_jk.py
+++ b/pyscf/pbc/df/df_jk.py
@@ -121,12 +121,18 @@ def get_j_kpts(mydf, dm_kpts, hermi=1, kpts=numpy.zeros((1,3)), kpts_band=None):
     weight = 1./nkpts
     rhoR *= weight
     rhoI *= weight
-    vjR = numpy.zeros((nset,nband,nao_pair))
-    vjI = numpy.zeros((nset,nband,nao_pair))
+    if hermi == 0:
+        aos2symm = False
+        vjR = numpy.zeros((nset,nband,nao**2))
+        vjI = numpy.zeros((nset,nband,nao**2))
+    else:
+        aos2symm = True
+        vjR = numpy.zeros((nset,nband,nao_pair))
+        vjI = numpy.zeros((nset,nband,nao_pair))
     for k, kpt in enumerate(kpts_band):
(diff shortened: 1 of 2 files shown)


fishjojo (approved):

hongzhouye:
> Is it possible to add a test which can detect this kind of error?

This is a good point. For DF, what's your opinion on replacing all existing tests based on He dimer in a box with a slightly more complicated system, like H2O in a box shown here? The existing one is insufficient for detecting the bug reported here (all ERI elements from the buggy DF code are correct for He dimer, but not for H2O).

For TDSCF, it seems that there are only tests for `krhf_slow` but not for others like `k[r/u/g]hf`. I can certainly add some tests for those.

fishjojo:
I think it's fine to add a few H2O tests, and tests for kTDSCF are welcome.

PR #1771, opened 2023-06-21
Remove deprecated keyword from scipy.linalg.solve

wangenau:
* The `sym_pos` keyword will be removed in SciPy 1.11, use `assume_a` instead
  See https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.solve.html
* SciPy 1.11 could release soon since there is already a second release candidate
* The assume_a keyword exists at least since SciPy 1.0, so there should be no compatibility issue

### DIFF
--- a/pyscf/df/grad/rhf.py
+++ b/pyscf/df/grad/rhf.py
@@ -244,7 +244,7 @@ def get_j(mf_grad, mol=None, dm=None, hermi=0):
 
     # (P|Q)
     int2c = auxmol.intor('int2c2e', aosym='s1')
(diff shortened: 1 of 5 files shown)


fishjojo (approved):

ndattani:
My student Sichao Xiong also fixed this here: https://github.com/HPQC-LABS/pyscf/pull/3
We were going to make a pull request probably today, it looks like it won't be necessary anymore!

timmysilv:
hello! 👋 @fishjojo (or other maintainers) is there a plan for a bugfix release including this change sometime soon to go with the Scipy 1.11 release?

fishjojo:
@timmysilv PySCF 2.3 will be released very soon.

Issue #1758, opened 2023-06-08
Error when calculating the IAOs of a supercell with space_group_symmetry=True.


cjcscott:
Hello!
I've come across an error when trying to obtain the IAOs of a supercell generated from a cell initialised with space group symmetry.
A minimal example reproducing the issue is
```
import pyscf.lo
import pyscf.pbc
import numpy as np
​
cell = pyscf.pbc.gto.Cell()
cell.atom = ['He 0.0 0.0 0.0']
cell.a = 1.4 * np.eye(3)
cell.basis = 'def2-svp'
cell.output = 'pyscf.out'
cell.space_group_symmetry = True
cell.symmorphic = True
cell.build()
​
scell = pyscf.pbc.tools.super_cell(cell, (2,2,2))
pyscf.lo.iao.reference_mol(scell, minao="sto-3g")
```

which gives traceback

```
Traceback (most recent call last):
 File "testscript.py", line 16, in <module>
    pyscf.lo.iao.reference_mol(scell, minao="sto-3g")
 File "python3.10/site-packages/pyscf/lo/iao.py", line 135, in reference_mol
    pmol.build(False, False, basis=minao)
 File "python3.10/site-packages/pyscf/pbc/gto/cell.py", line 1516, in build
    _lattice_symm.build(space_group_symmetry=True,
 File "python3.10/site-packages/pyscf/pbc/symm/symmetry.py", line 176, in build
    self.spacegroup = space_group.SpaceGroup(cell).build(dump_info=False)
 File "python3.10/site-packages/pyscf/pbc/symm/space_group.py", line 302, in build
    self.ops = geom.search_space_group_ops(self.cell, tol=self.symprec)
 File "python3.10/site-packages/pyscf/pbc/symm/geom.py", line 81, in search_space_group_ops
    atmgrp = mole.atom_types(cell._atom, magmom=cell.magmom)
 File "python3.10/site-packages/pyscf/gto/mole.py", line 308, in atom_types
    uniq_mag = np.unique(magmom[idx])
IndexError: index 1 is out of bounds for axis 0 with size 1
```

This can be avoided by disabling space group symmetry initially or in the supercell object after generation, but I don't know whether that should be necessary. Any advice as to whether this is an intended incompatibility would be greatly appreciated.

PR #1773, opened 2023-06-23
fix supercell with symmetry

fishjojo:
Fix issue #1758.


pyscf/gto/mole.py
```diff
@@ -2561,9 +2561,9 @@ def build(self, dump_input=True, parse_arg=True,
             # number of electrons are consistent.
             self.nelec
 
-        if not self.magmom:
-            self.magmom = [0.,]*self.natm
-        if self.spin == 0 and abs(numpy.sum(numpy.asarray(self.magmom)) - self.spin) > 1e-6:
+        if self.magmom is None or len(self.magmom) <= 0:
```
sunqm, line 2564: `len(self.magmon) != self.natm`

sunqm (approved):

fishjojo:
@sunqm should this fix be added to 2.3 release as well?

PR #1786, opened 2023-07-05
Release 2.3

sunqm:
Solving the scipy compatibility issue.

Big changes include:
* Drop supports for Python 3.6
* The default parameters for B3LYP functional have been updated. It is now compatible with Gaussian
* The command line parser is disabled by default.

Other new features and bugfixes can be found in CHANGELOG.


verena-neufeld:
Thanks so much @sunqm ! 
Do we know why tests are currently failing on the main branch?

hebrewsnabla:
> The default parameters for B3LYP functional have been updated. It is now compatible with ORCA and Gaussian

It's equivalent to Gaussian's B3LYP but not ORCA's since the ORCA B3LYP uses VWN5. But we can say that 'B3LYP5' is compatible with ORCA's B3LYP.

fishjojo (approved):

PR #1790, opened 2023-07-08
Fix changelog and warn message about B3LYP

hebrewsnabla:
After #1481, current PySCF choice of B3LYP (the same as Gaussian's) is different with ORCA. See ORCA manual 9.4.2.1

> In ORCA VWN5 is chosen for the local correlation part. This choice is consistent with the TurboMole
program but not with the Gaussian program. 

https://github.com/pyscf/pyscf/blob/fc9995cad1318409303ba939d230d02196aa8109/pyscf/dft/libxc.py#L737-L739

So the changelog and libxc.py stating that PySCF's B3LYP is the same to ORCA's should be corrected.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,7 +8,7 @@ PySCF 2.3.0 (2023-07-04)
   - TDDFT gradients for triplet states
   - Support complex-valued h1e in fci_slow.absorb_h1e
 * Improved
-  - Update B3LYP functional to make it behave the same to Gaussian
+  - Update B3LYP functional to correspond to the original definition
   - Disable CLI parser by default
   - Accuracy and performance of RSDF, GDF and RSJK methods
   - get_lattice_Ls, and energy cutoff estimation
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,7 +8,7 @@ PySCF 2.3.0 (2023-07-04)
   - TDDFT gradients for triplet states
   - Support complex-valued h1e in fci_slow.absorb_h1e
 * Improved
-  - Update B3LYP functional to make it behave the same to ORCA
+  - Update B3LYP functional to make it behave the same to Gaussian
   - Disable CLI parser by default
   - Accuracy and performance of RSDF, GDF and RSJK methods
   - get_lattice_Ls, and energy cutoff estimation
--- a/pyscf/dft/libxc.py
+++ b/pyscf/dft/libxc.py
@@ -770,7 +770,7 @@ XC_CODES.update({
 # Issue 1480
 if not hasattr(__config__, 'B3LYP_WITH_VWN5'):
     warnings.warn('Since PySCF-2.3, B3LYP (and B3P86) are changed to the VWN-RPA variant, '
-                  'the same to the B3LYP functional in Gaussian and ORCA (issue 1480). '
+                  'the same to the B3LYP functional in Gaussian (issue 1480). '
                   'To restore the VWN5 definition, you can put the setting '
                   '"B3LYP_WITH_VWN5 = True" in pyscf_conf.py')
 elif getattr(__config__, 'B3LYP_WITH_VWN5', False):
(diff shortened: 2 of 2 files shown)


CHANGELOG
```diff
@@ -8,7 +8,7 @@ PySCF 2.3.0 (2023-07-04)
   - TDDFT gradients for triplet states
   - Support complex-valued h1e in fci_slow.absorb_h1e
 * Improved
-  - Update B3LYP functional to make it behave the same to ORCA
+  - Update B3LYP functional to make it behave the same to Gaussian
```
susilehtola, line 11: This is misleading. B3LYP is B3LYP. The earlier implementation did not match it.

pyscf/dft/libxc.py
```diff
@@ -770,7 +770,7 @@ def _xc_key_without_underscore(xc_keys):
 # Issue 1480
 if not hasattr(__config__, 'B3LYP_WITH_VWN5'):
     warnings.warn('Since PySCF-2.3, B3LYP (and B3P86) are changed to the VWN-RPA variant, '
-                  'the same to the B3LYP functional in Gaussian and ORCA (issue 1480). '
+                  'the same to the B3LYP functional in Gaussian (issue 1480). '
```
susilehtola, line 773: this should read: the original B3LYP functional as published by Stephens et al.

pyscf/dft/libxc.py
```diff
@@ -770,7 +770,7 @@ def _xc_key_without_underscore(xc_keys):
...
                   'To restore the VWN5 definition, you can put the setting '
                   '"B3LYP_WITH_VWN5 = True" in pyscf_conf.py')
```
susilehtola, line 775: What is the point of that keyword? Just use B3LYP5 if you want B3LYP5.

CHANGELOG
hebrewsnabla: So what's your suggestion about this changelog?

pyscf/dft/libxc.py
hebrewsnabla: After some struggling, I can understand that your 'the B3LYP functional' of ORCA refers to the one called 'B3LYP_G' in ORCA manual, does it? That makes sense, but it will be very hard to understand as a warning message.

CHANGELOG
susilehtola: Fix the definition of B3LYP to correspond to the original functional.

pyscf/dft/libxc.py
susilehtola: ?

hebrewsnabla: I mean, the B3LYP keyword (a wrong B3LYP definition) of ORCA refers to B3LYP5 (although it has a right B3LYP definition called B3LYP_G), so it's misleading to state here that current PySCF's B3LYP is the same to the one in ORCA.

susilehtola: Yes, ORCA, TURBOMOLE etc don't implement B3LYP but B3LYP5 which is a different functional. They are wrong.

hebrewsnabla: So could this conversation being resolved? Removing 'and ORCA' here is better, right?

susilehtola: It is better, but "same as B3LYP functional in Gaussian" is still not saying the main point, which is that that implementation is the correct one.

hebrewsnabla: What about "corresponding to the original definition by Stephens et al. (issue 1480) and the same as the B3LYP functional in Gaussian"

pyscf/dft/libxc.py
hebrewsnabla: Do you think something should be modified here? That looks introduced in #1481.

pyscf/dft/libxc.py
susilehtola: @hebrewsnabla yes!

pyscf/dft/libxc.py
susilehtola: Maybe @sunqm  can elucidate the introduction of that keyword? Would not a simple instruction to use the `B3LYP5` functional, instead, suffice?

sunqm (approved):

pyscf/dft/libxc.py
sunqm: This keyword allows XC parser working the same to old versions.

Issue #1187, opened 2022-02-14
exxdiv=None is not completely propagated in gamma point pbc/tdscf


btglau:
The wrapper to take pbc/scf to molecular tdscf,
https://github.com/pyscf/pyscf/blob/14b880cdc4bf7bb64f3989a147c0b04622efc498/pyscf/pbc/tdscf/rhf.py#L31

Creates a temporary environment where `exxdiv=None`, but this setting does not completely go through the molecular tdscf calculation. In particular, the `mo_energy` is taken from the existing quantity `mf.mo_energy`, instead of calculated from the diagonalization of a freshly made Fock AO matrix, like how the `common_init` works in pyscf/cc. In fact, those lines are commented out!

https://github.com/pyscf/pyscf/blob/14b880cdc4bf7bb64f3989a147c0b04622efc498/pyscf/tdscf/rhf.py#L74

sunqm:
TDDFT code calls the get_jk functions to compute the contraction between two-electron integrals and X,Y amplitudes. Setting `exxdiv=None` here is to ensure the special treatment of exxdiv being excluded because get_jk by default utilize the exxdiv settings in mf object. I guess the question here is whether the exxdiv corrections to orbital energies should be excluded when solving TDDFT response equations. To make TDA results consistent with CIS excited states, the exxdiv corrections should be excluded from orbital energy. However, DFT orbital energy gaps can be used as a good estimation for excitation energies. To make the TDDFT excited energies consistent with the orbital energy gaps, the exxdiv correction has to be kept in the orbital energy when computing the A matrix in TDA as the current implementation. @gkc1000 what do you think?

gkc1000:
I do not know if there is a completely consistent way to do this theoretically.  With correlated methods, the problem is the following: let's say the method depends on H = h + V. Then there is no exxdiv correction to H, since K does not explicitly appear. However, if we rewrite H in normal ordered form, then H = const + F + W. Then one *might* expect exxdiv to apply to the K part of F, but it actually changes the results. My guess is that if the correlated theory does not involve any (vv|vv) integrals - as is true for MP2 and TD-HF-TDA then modifying H -> const + F + exxdiv + W can in practice improve the results.

[...]

PR #1793, opened 2023-07-09
Complete implementation of pbc/tdscf

hongzhouye:
This PR completes the implementation of the `pbc/tdscf` module (thanks to @tberkel and @xwang862 for helpful discussions):
1. Support for non-zero particle-hole momentum transfer for `GDF/RSDF` (same as the `kshift` used in pbc eom-ee-ccsd)
2. Support for twisted boundary conditions (both single k-point and  k-point mesh).
4. Fix finite-size correction (1/Nk convergence to TDL and agreement with molecular code).
5. Add tests and examples.

Regarding the finite-size correction, one can show (thanks to @tberkel for discussion) that the TDSCF equations are independent of treatment of the Coulomb singularity at G = 0. As a result, one should either include or exclude such treatments everywhere. The existing pbc tdscf code (see issue #[1187](https://github.com/pyscf/pyscf/issues/1187)) uses `exxdiv = None` for the J/K build, but does not remove the Madelung constant shift in the occupied orbital energy when `exxdiv = 'ewald'`. The new implementation forces the removal of the latter inside the tdscf code. Doing so leads to `1/Nk` convergence of pbc excitation energies to the TDL as well as agreement with molecular excitation energies. This also applies to hybrid DFT. Non-hybrid DFT, which already satisfies these, is not affected by this change as no exxdiv treatment is needed.

A small change was made to the molecular `tdscf` module's init guess (which applies to pbc as well). The previous implementation uses a fixed tolerance of 1e-6 Ha for selecting degenerate particle-hole pairs as init guess vectors. This threshold is too tight and can miss roots beyond the first root. The new implementation increases the default value to 1e-3 Ha and allows user to alter it via the `deg_eia_thresh` variable. If the number of init vectors generated this way exceeds `nroots`, `trunc_workspace` is called to truncate the working space to `nroots` before entering the Davidson solver. An example is added `example/tdscf/04-tddft_missing_roots.py`.

Two questions:
1. To allow for non-Gamma single k-point calculations, I have to comment out `assert (mo_coeff.dtype == numpy.double)` a few places in the molecular code (`tdscf/rhf.py` and `tdscf/rks.py`). I checked that all existing linear algebra is correct in terms of `.conj()` and needs no changes. But I am not so sure about other parts, e.g., blocks that check and operate on symmetry. Please let me know if there are potential conflicts.
2. For a KS-DFT reference, the following two sanity checks do not pass for all xc functionals.
- First, there are two ways of running TDDFT that should give the same answer:
```
td1 = mf.TDDFT().run()
```
calls the specialized `CasidaTDDFT` code while
```
from pyscf.tdscf import rhf   # similar for pbc.tdscf.rhf and pbc.tdscf.krhf
td2 = rhf.TDHF(mf).run()
```
calls the `TDHF` code.
- Second, as box size becomes large, pbc code (both single k-point or k-points) should reproduce the molecular code results.

This [script](https://github.com/pyscf/pyscf/files/11995573/test_tddft.py.zip) examines the two sanity checks for different xc's and the results are summarized below. One can see that both the molecular code (column 1) and the k-points pbc code (column 5-7) are good, but the single k-point code (column 2-4) is problematic and this depends on xc. My suspect is that `fxc` term is calculated differently in the molecular code and the pbc code but I am not so sure.

(`mtd`, `std`, and `ktd` means molecular, single-kpt, and k-point code; `1` and `2` label the two different ways of running TDDFT calculations; `True` if two codes predict excitation energies < 0.01 eV)
| xc | `mtd1.e == mtd2.e` | `std1.e == std2.e` | `std1.e == mtd1.e` | `std2.e == mtd1.e` | `ktd1.e == ktd2.e` | `ktd1.e == mtd1.e` | `ktd2.e == mtd1.e` |
| -- | -------------------- | ------------------ | ------------------ | ------------------- | ------------------ | ----------- | ---|
| LDA | True | True | True | True | True | True | True |
| PBE | True | False | True | False | True | True | True |
| BLYP | True | False | True | False | True | True | True |
| PBE0 | True | True | False | False | True | True | True |
| B3LYP | True | True | True | True | True | True | True |


pyscf/pbc/scf/_response_functions.py
```diff
@@ -189,6 +189,26 @@ def _gen_ghf_response(mf, mo_coeff=None, mo_occ=None,
     '''
     raise NotImplementedError
 
+def _get_jk_kshift(mf, dm_kpts, hermi, kpts, kshift, with_j=True, with_k=True):
+    from pyscf.pbc.df.df_jk import get_j_kpts_kshift, get_k_kpts_kshift
+    vj = vk = None
+    if with_j:
+        vj = get_j_kpts_kshift(mf.with_df, dm_kpts, kshift, hermi=hermi, kpts=kpts)
```
sunqm, line 196: `mf.get_jk` may call the `mf.rsjk.get_jk` method. Check if mf.rsjk is initialized as the get_jk code implemented in `scf/khf.py` . If rsjk object available, an assertion to `kshift=0` may be needed I guess.

sunqm:
Looks like errors are all related to the non-hermitian diagonalization solver for std2.e . The `fxc` for pbc code are refactored in this developing branch https://github.com/sunqm/pyscf/tree/mcfun-pbc . Later we may try the new fxc code in that branch

pyscf/tdscf/rhf.py
```diff
@@ -803,12 +829,14 @@ def init_guess(self, mf, nstates=None, wfnsym=None):
             orbsym_in_d2h = numpy.asarray(orbsym) % 10  # convert to D2h irreps
             e_ia[(orbsym_in_d2h[occidx,None] ^ orbsym_in_d2h[viridx]) != wfnsym] = 1e99
 
+        e_ia_uniq = numpy.unique(e_ia)
```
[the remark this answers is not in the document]
hongzhouye: @sunqm I believe that `e_ia` here is effectively rounded by `numpy.ceil` a few lines above

pyscf/grad/test/test_tduks_grad.py
```diff
@@ -49,18 +50,18 @@ def tearDownModule():
 
 class KnownValues(unittest.TestCase):
     def test_tda_lda(self):
-        td = tdscf.TDA(mf_lda).run(nstates=3)
+        td = tdscf.TDA(mf_lda).run(nstates=nstates)
         tdg = td.nuc_grad_method()
         g1 = tdg.kernel(td.xy[2])
-        self.assertAlmostEqual(g1[0,2], -0.40279473514282405, 6)
+        self.assertAlmostEqual(g1[0,2], -0.7944872119457362, 6)
```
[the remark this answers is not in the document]
hongzhouye: @sunqm requesting more states (5 rather than 3) seems to be the cause. For example, even using the master branch which has `nstates=3` gives the following excitation energies
```
[0.12166646 0.12166646 0.80854428]
```
while changing it to `nstates=5` gives
```
[0.12166646 0.12166646 0.75668038 0.80854428 0.80854428]
```
You can see that the energy of the third state changes. In general, I found it numerically more stable to request a few more states. So I changed it to `nstates=5` in almost all tdscf-related tests. Not sure if this is a general solution or there's anything we can do on the davidson solver side.

pyscf/pbc/scf/_response_functions.py
hongzhouye: The current code checks `isinstance(mf.with_df, df.DF)`, which evaluates to `True` for `rsjk` since it uses GDF for `vnuc` build. As a result, the code will automatically cache j3c if roots with non-zero `kshift` are requested. What do you think this behavior (compared to an `assert` error)? If so I can add a warning for `rsjk`.

hongzhouye:
> Looks like errors are all related to the non-hermitian diagonalization solver for std2.e . The `fxc` for pbc code are refactored in this developing branch https://github.com/sunqm/pyscf/tree/mcfun-pbc . Later we may try the new fxc code in that branch

Sounds good! Glad to hear that we do have a suspect.

verena-neufeld:
Thanks so much @hongzhouye ! `exxdiv == "ewald"` is dealt with explicitly now, other exxdiv options might also need to be dealt with.

hongzhouye:
> Thanks so much @hongzhouye ! `exxdiv == "ewald"` is dealt with explicitly now, but other exxdiv options might also need to be dealt with.

Thanks for the suggestions @verena-neufeld ! I just changed the way exxdiv is handled. Now it is the same as CC in the sense that mo energy is always recalculated from the diagonal of fock with `exxdiv='none'`, to match the (ii|aa) type integrals which are calculated using `exxdiv='none'` as well. Please let me know if this change looks good to you.

verena-neufeld:
On the first glance, the changes look good (I don't know the details of TDSCF but now Ewald is not singled out while vcut_ws , etc, are not dealt with). 
One question: Do you mean to write `if` <something else> `mf.mo_coeff is mo_coeff`? or `==`

hongzhouye:
> On the first glance, the changes look good (I don't know the details of TDSCF but now Ewald is not singled out while vcut_ws , etc, are not dealt with). One question: Do you mean to write `if` ...something else... `mf.mo_coeff is mo_coeff`? or `==` (or actually a numpy comparison function)?

@verena-neufeld It checks if input `mo_coeff` is given. If not, `mo_coeff` will be set to `mf.mo_coeff` and that if statement will be true. If also `exxdiv is None`, there is no need to recompute the `fock` matrix and the code will return `mf.mo_energy`.

verena-neufeld:
Thanks for clarifying @hongzhouye !

pyscf/pbc/scf/_response_functions.py
sunqm: I think an error can be raised here. We just need to ensure that users do not get incorrect results. It is not a problem if this combination is not supported.

pyscf/tdscf/rks.py
```diff
@@ -56,7 +56,7 @@ def gen_vind(self, mf=None):
 
         mol = mf.mol
         mo_coeff = mf.mo_coeff
-        assert mo_coeff.dtype == numpy.double
+        # assert mo_coeff.dtype == numpy.double
```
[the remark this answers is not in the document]
hongzhouye: You are right. The code assumes that `A-B` is diagonal, which assumes `(ia|jb) = (ia|bj)`. I added the assert back in the molecular rks code and made pbc rks calls this molecular code only for non-hybrid & Gamma-point. Otherwise, the pbc tda code will be used.

pyscf/pbc/scf/_response_functions.py
hongzhouye: No problem. I added a `NotImplementedError` for rsjk in both `pbc/tdscf/krhf.py` and `pbc/scf/_response_functions.py`. Please let me know if it looks good to you.

pyscf/pbc/tdscf/rks.py
```diff
@@ -32,15 +33,16 @@ class CasidaTDDFT(rhf.TDA):
...
         return CasidaTDDFT(mf)
+    else:
+        return TDDFT(mf)
 
 dft.rks.RKS.TDA           = lib.class_as_method(rhf.TDA)
 dft.rks.RKS.TDHF          = None
-dft.rks.RKS.TDDFTNoHybrid = lib.class_as_method(TDDFTNoHybrid)
-dft.rks.RKS.CasidaTDDFT   = lib.class_as_method(CasidaTDDFT)
+dft.rks.RKS.TDDFTNoHybrid = tddft
+dft.rks.RKS.CasidaTDDFT   = tddft
```
sunqm, line 45: Let CasidaTDDFT do CasidaTDDFT. Do not need to select the correct TDDFT for user in this case

sunqm (approved):

pyscf/pbc/tdscf/rks.py
hongzhouye: Just fixed it. Now they look like
```
dft.rks.RKS.TDDFTNoHybrid = tddft
dft.rks.RKS.CasidaTDDFT   = lib.class_as_method(CasidaTDDFT)
dft.rks.RKS.TDDFT         = tddft
```

verena-neufeld:
Do some changes here affect pac stability analysis?

hongzhouye:
> Do some changes here affect pbc scf stability analysis?

In principle no as JK-build with zero momentum transfer (which is all stability analysis needs) is unchanged. It also seems to pass all tests.

sunqm:
@hongzhouye  the warnings in lint need to be fixed.

verena-neufeld:
One of the tests is currently failing with "FAILED pyscf/solvent/test/test_ddcosmo_tdscf_grad.py::KnownValues::test_b3lyp_tda - NotImplementedError: complex density matrix" On first glance, this might be related to this PR?

hongzhouye:
> One of the tests is currently failing with "FAILED pyscf/solvent/test/test_ddcosmo_tdscf_grad.py::KnownValues::test_b3lyp_tda - NotImplementedError: complex density matrix" On first glance, this might be related to this PR?

@verena-neufeld Any idea in general why a test only fails for one of the test settings?

PR #1797, opened 2023-07-14
use system temp directory by default

### LINKED ISSUE #1796: Actually use system-wide temporary directory by default
The documentation at https://pyscf.org/install.html says
> The environment variable PYSCF_TMPDIR controls which directory is used to store intermediate files and temporary data when PySCF is run; it is also commonly known as the scratch directory. If this environment variable is not set, the system-wide temporary directory TMPDIR will be used as the temp directory, instead.

This phrasing is unclear because unless the environment variable `TMPDIR` is set, pySCF will _not_ use the system-wide temporary directory (which on Linux is `/tmp`), but will use the current directory instead. This is verified by reading the source code at https://github.com/pyscf/pyscf/blob/1189489af017c8d93e1e73fc24cf07289ab9e763/pyscf/__config__.py#L11

The behavior should be changed to actually use the system-wide temporary directory if neither `PYSCF_TMPDIR` nor `TMPDIR` are set. It looks like this directory can be retrieved by calling [`tempfile.gettempdir()`](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir), which already checks for the `TMPDIR` environment variable.

kevinsung:
Fixes #1796

### DIFF
--- a/pyscf/__config__.py
+++ b/pyscf/__config__.py
@@ -1,4 +1,4 @@
-import os, sys
+import os, sys, tempfile
 
 #
 # All parameters initialized before loading pyscf_conf.py will be overwritten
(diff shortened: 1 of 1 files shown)


xubwa:
I think it would be nice to keep the original way to read "TMPDIR" variable for backward compatibility.

kevinsung:
Since this change only affects temporary files, I don't think there is any backward compatibility issue. Generally speaking, using the system-wide temporary directory for temporary files is less surprising (and more "correct") than using the current directory. Personally, I prefer not to have my current directory cluttered with temporary files that inevitably fail to get cleaned up.

kevinsung:
Oh, if you're talking about reading the `TMPDIR` environment variable, that is already read by [`tempfile.gettempdir`](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir):
> Python searches a standard list of directories to find one which the calling user can create files in. The list is:
> 1. The directory named by the TMPDIR environment variable.
> ...

sunqm (approved):

PR #1806, opened 2023-07-24
Allow pickling of pyscf.gto.Mole by serializing as JSON-formatted string

chillenb:
Problems with pickling `Mole` objects can be avoided by defining the `__getstate__` and `__setstate__` methods to use the existing JSON serialization method. This approach is convenient for people using pickling in their workflows and avoids code duplication.


sunqm (approved):

chillenb:
I just rebased off master and removed extraneous whitespace so flake8 is happy

sunqm:
Adding `__getstate__` and `__setstate__` will lead to other compatibility issues. I will refactor the relevant code in a new PR.

chillenb:
@sunqm thanks for taking the time to look at this! I guess the behavior of gto.Mole is complex.

PR #1817, opened 2023-08-02
Fix bug in symm cas methods

yangdatou:
These methods might raise error if `mo_coeff` not given.


sunqm (approved):

fishjojo:
I suggest writing pure functions. Instead of overwriting the class attribute, it may be better to fix the other parts of the code.

sunqm:
> I suggest writing pure functions. Instead of overwriting the class attribute, it may be better to fix the other parts of the code.

It is a good suggestion that can be considered in future. Rewriting other parts of the code is too much work for of this PR.

PR #1818, opened 2023-08-05
added NVT BOMD and example

aniruddha-seal:
This PR adds NVT ensemble to the md module. I have implemented the Berendsen thermostat.


pyscf/md/integrators.py
```diff
@@ -202,11 +211,14 @@ def __init__(self, method, **kwargs):
         self.verbose = self.mol.verbose
         self.steps = 1
         self.dt = 10
+        self.T = None
```
matthew-hennefarth, line 214: The temperature, tau, and temp_output don't need to go into the base `_Integrator` class but should only be placed into the subclass.

pyscf/md/integrators.py
```diff
@@ -407,6 +450,16 @@ def _write_energy(self):
 
         # If we don't flush, there is a possibility of losing data
         self.energy_output.flush()
+        
+    def _write_temp(self):
```
matthew-hennefarth, line 454: I am not sure if we should place the temperature variable into its own output file, or if we should change the `energy_output` to `data_output` where it stores all of the various quantities in a different column.

pyscf/md/integrators.py
```diff
@@ -322,6 +358,10 @@ def check_sanity(self):
         assert self.veloc is not None
         assert self.veloc.shape == (self.mol.natm, 3)
         assert self.scanner is not None
+
```
matthew-hennefarth, line 361: I recommend overloading the `check_sanity` func in the `NVTBerendson` and then calling the parent function within that. Something like:

```python
def check_sanity(self):
    assert self.T > 0
    assert self.taut is not None
    self.super().check_sanity()
```

pyscf/md/integrators.py
```diff
@@ -310,6 +341,11 @@ def dump_input(self, verbose=None):
         log.info('******** BOMD flags ********')
         log.info('dt = %f', self.dt)
         log.info('Iterations = %d', self.steps)
+        
+        if self.T is not None:
```
matthew-hennefarth, line 345: Similar to `check_sanity`, perhaps have a overridden method in the subclass like:
```python
def dump_input(self, verbose=None):
    log = logger.new_logger(self, verbose)
    self.super().dump_input(verbose=log)
    log.info('Temperature = %f', self.T)
    log.info('tau_T = %f', self.taut)
```

matthew-hennefarth:
This PR needs tests as well to ensure it conserves temperature.

xis19:
Sorry for the abruptly comment, but Berendson thermostat does not properly simulates the canonical ensemble, see, e.g. P.H. Huenenberger, Adv. Polym. Sci. 173 2005 105. An alternative approach, which is proven to be projecting to the right phase space, is the Bussi thermostat, which is close to the original Berendsen approach but with proper correction, see, e.g. G. Bussi, D. Donadio, M. Parrinello, J. Chem. Phys., 126 2007 014101. Another implementation is the Nose-Hoover approach, which is discussed in detail in Statistics Mechanics, M. Tuckerman, 2010. I suppose Bussi thermostat might be the easiest approach.

xis19:
Also, I would feel that one possible way of implementing the integrator is to apply Trotter factorization, see, e.g. formula 283 in https://web.northeastern.edu/afeiguin/phys5870/phys5870/node103.html

The implementation might look like this:

```
import abc
import dataclasses

from typing import List
from typing_extensions import override

@dataclasses.dataclass
class SimSystem:
    positions: List[float]
    velocities: List[float]
    accelerations: List[float]

class Liouville:

    def __init__(self, simulation_system, dt):
        self._system = simulation_system
        self._dt = dt

    @abc.abstractmethod
    def first_step(self):
        raise NotImplementedError()

    @abc.abstractmethod
    def second_step(self):
        raise NotImplementedError()


class VelocityVerlet_PositionUpdate(Liouville):

    def __init__(self, simulation_system, dt, hamilton):
        super.__init__(simulation_system, dt)

        self._hamilton = hamilton

    @override
    def first_step(self):
        self._system._positions += self._system.velocities * self._dt

    @override
    def second_step(self):
        # Assume system Hamilton is position-dependence only
        self._system._accelerations = hamilton.update_force(self._system.positions) / atom_mass_list


class VelocityVerlet_VelocityUpdate(Liouville):

    def __init__(self, simulation_system, dt):
        super.__init__(simulation_system, dt)

    @override
    def first_step(self):
        self._system.velocities += self._system.accelerations * self._dt / 2

    @override
    def second_step(self):
        self._system.velocities += self._system.accelerations * self._dt / 2


class Integrator:

    def __init__(self):
        self._liouvilles = []

    def add_liouville(self, liouville):
        self._liouvilles.append(liouville)

    def step(self):
        for liouville in self._liouvilles:
            liouville.first_step()

        for liouville in reversed(self._liouvilles):
            liouville.second_step()
```

I have a similar MD code implemented this way in C++ years ago, unfortunately the code belongs to the lab, so I am not allowed to share it to public.

matthew-hennefarth:
@xis19 and @erikik Thank you for the comments regarding the Berendson thermostat and its issues. I think it would be nice to implement all of the NVT integrators discussed, (Berendson, Nose-Hover, Bussi, etc) and leave it to the end user which integrator they want to use.

pyscf/md/__init__.py
```diff
@@ -45,4 +45,4 @@ def set_seed(seed):
 from pyscf.md import integrators, distributions
 
 NVE = integrators.VelocityVerlet
-
+NVT = integrators.NVTBerendson
```
matthew-hennefarth, line 48: I don't think we should make the Berendson thermostat the default for NVT given the discussion

aniruddha-seal:
Thanks for the comments! @erikik and @xis19, I will implement the Nose-Hoover Chain Thermostat. In that case as @matthew-hennefarth mentioned, PySCF will have both the thermostats that Orca and OpenMolcas allows!

xis19:
> @xis19 and @erikik Thank you for the comments regarding the Berendson thermostat and its issues. I think it would be nice to implement all of the NVT integrators discussed, (Berendson, Nose-Hover, Bussi, etc) and leave it to the end user which integrator they want to use.
> 
> For example, Orca (as [4.2.1](https://www.afs.enea.it/software/orca/orca_manual_4_2_1.pdf)) has the Berendson thermostat. OpenMolcas only has [Nose-Hoover](https://molcas.gitlab.io/OpenMolcas/sphinx/users.guide/programs/dynamix.html). It would be nice for PySCF to offer both!

Thanks for considering the alternative approach. Could you also consider using the Trotter factorization? AFAIK molecular dynamics people tend to use that since it helps simplifying certain issue, one could combine different algorithms much easier, e.g. use Bussi thermostat with Nose Hoover baristas, and also apply RATTLE-ROLL to certain molecules.

Also, IMHO we could also implement the Anderson thermostat as a reference since it is one of the earliest algorithm designed. The original paper is J Chem. Phys. 72 4 2384, it is also implemented in VASP.

matthew-hennefarth:
> Thanks for considering the alternative approach. Could you also consider using the Trotter factorization? AFAIK molecular dynamics people tend to use that since it helps simplifying certain issue, one could combine different algorithms much easier, e.g. use Bussi thermostat with Nose Hoover baristas, and also apply RATTLE-ROLL to certain molecules. Some references:
> 
> GROMACS documentation:
> 
> https://manual.gromacs.org/documentation/2019-rc1/reference-manual/algorithms/molecular-dynamics.html
> 
> LAMMPS presentation:
> 
> https://www.osti.gov/servlets/purl/1884410
> 
> Also, IMHO we could also implement the Anderson thermostat as a reference since it is one of the earliest algorithm designed. The original paper is J Chem. Phys. 72 4 2384, it is also implemented in VASP.

We can look into refactoring this MD module to the Trotter factorization. Probably better to do so sooner rather than later.

pyscf/md/integrators.py
aniruddha-seal: Done!

pyscf/md/integrators.py
aniruddha-seal: Done!

pyscf/md/integrators.py
aniruddha-seal: Done!

pyscf/md/integrators.py
aniruddha-seal: Done!

pyscf/md/__init__.py
aniruddha-seal: Changed to NVT_Berendson!

pyscf/md/integrators.py
```diff
@@ -494,3 +495,104 @@ def _next_velocity(self, next_accel):
...
+        accel : ndarray
+            Current acceleration for the simulation. Values are given
+            in atomic units (Bohr/a.u.^2). Dimensions is (natm, 3) such as
+
+             [[x1, y1, z1],
+             [x2, y2, z2],
+             [x3, y3, z3]]
+    '''
+
+    def __init__(self, method, **kwargs):
```
matthew-hennefarth, line 527: I think you need `__init__` to take `T` and `taut`. This function should look more like
```python
def __init__(self, method, T, taut, **kwargs):
    self.T = T
    self.taut = taut
    self.accel = None
    super().__init__(method, **kwargs)
```

pyscf/md/__init__.py
```diff
@@ -45,4 +45,4 @@ def set_seed(seed):
 from pyscf.md import integrators, distributions
 
 NVE = integrators.VelocityVerlet
-
+NVT_Berendson = integrators.NVTBerendson
```
matthew-hennefarth, line 48: I would just remove this from the `__init__.py` file.

pyscf/md/integrators.py
aniruddha-seal: I have added that! But should be redundant I guess!!

pyscf/md/__init__.py
aniruddha-seal: Done!

aniruddha-seal:
> This PR needs tests as well to ensure it conserves temperature.

Added unit test.

pyscf/md/test/test_NVTBerendson.py
```diff
@@ -0,0 +1,70 @@
...
+
+import unittest
+import numpy as np
+from pyscf import gto, scf
+import pyscf.md as md
+
+CHECK_STABILITY = False
+
+def setUpModule():
+    global h2o, hf_scanner, casscf_scanner
```
xis19, line 26: I would guess that a CASSCF is not necessary since static correlation is not strong on water.

matthew-hennefarth: probably copied from the [test_velocverlet.py](https://github.com/pyscf/pyscf/blob/a27ddc77bcad42b0ee8707d93acc691fe9d51e5d/pyscf/md/test/test_velocverlet.py#L25-L55) file where I was doing some benchmarks between OpenMolcas and PySCF. Not needed in this test since it is not needed.

aniruddha-seal: Corrected!

pyscf/md/test/test_NVTBerendson.py
```diff
@@ -0,0 +1,70 @@
...
+        final_coord = np.array([[0.03203339, 0.00219695, 0.04414054],
+ 				 [-0.03398398, -1.40833795, 1.18785634],
+ 				 [-0.10817007, 1.40567903, 1.10489554]])
+
+        self.assertTrue(np.allclose(driver.mol.atom_coords(), final_coord))
+        if CHECK_STABILITY:
+
+            driver.steps = 990
+            driver.kernel()
+            self.assertTrue((driver.T - 5)<= driver.temperature() <= (driver.T + 5))
```
xis19: I would also suggest 1) record the temperature per step 2) verify the fluctuation is within certain criteria; rather than just checking the final state. I also feel that 990 steps is relatively small, as in real MD simulations, people run a much longer time, e.g. 1ns. If the test time is a concern, maybe switching to some force fields like UFF would be much faster.

aniruddha-seal: I have equilibrated the system to 300 K and `init_veloc`  is set to the equlibrated velocity. So, checking the temperature of the final state should suffice to check if the canonical ensemble is being maintained. The test time will be system dependent, and the single water molecule considered here, should not require such long test time!

pyscf/md/test/test_prop.py
```diff
@@ -0,0 +1,62 @@
...
+    hf_scanner.max_cycle = 700
+
+def tearDownModule():
+    global h2o, hf_scanner
+    hf_scanner.stdout.close()
+    del h2o, hf_scanner
+
+class KnownValues(unittest.TestCase):
+
+    def test_temperature(self):
```
matthew-hennefarth, line 44: can you also add a test for a linear molecule

matthew-hennefarth (approved):
add the linear molecule temperature property test and should be good.

matthew-hennefarth (approved):

sunqm:
There are a few comments not solved. Otherwise it looks good to me. @matthew-hennefarth  please let me know when you think this PR is ready to merge.

matthew-hennefarth:
> There are a few comments not solved. Otherwise it looks good to me. @matthew-hennefarth please let me know when you think this PR is ready to merge.

@sunqm Should be good at this point!

sunqm (approved):

PR #1830, opened 2023-08-14
fix other rst issues

fchapoton:
another wave, fixing some other issues in the rst syntax of the documentation

### DIFF
--- a/pyscf/pbc/ao2mo/eris.py
+++ b/pyscf/pbc/ao2mo/eris.py
@@ -58,13 +58,16 @@ def get_mo_eri(cell, mo_coeffs, kpts=None):
 
 def get_mo_pairs_G(cell, mo_coeffs, kpts=None, q=None):
     '''Calculate forward (G|ij) FFT of all MO pairs.
+
     TODO: - Implement simplifications for real orbitals.
+
     Args:
         mo_coeff: length-2 list of (nao,nmo) ndarrays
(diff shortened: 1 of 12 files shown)


sunqm:
I don't think it is necessary to add \`\` in the doc string. Please remove them.

fchapoton:
These are necessary, because `*` has meaning in rst and a text starting with `*` should end with another `*`. Another solution would be to use `\*`.

verena-neufeld:
Thanks @fchapoton for helping with pyscf. Out of interest, what are the effects of including `` in practical terms? I don't have a strong opinion either way. When I checked the pyscf website, https://pyscf.org/pyscf_api_docs/pyscf.pbc.ao2mo.html#module-pyscf.pbc.ao2mo.eris for example renders fine for | and others but I'm a bit rusty on my rst. Thanks again!

fchapoton:
Putting part of the documentation string between `` has the consequence that this part is displayed in a different manner, and not interpreted by rst parser. A bit like "verbatim" blocks in latex.

I was using the command "tox -e rst ." (through sagemath, in fact) to run an automatic check of the rst syntax on your python files. It complains about the use of * and | that do not come in pairs, as these are delimiters in the rst syntax.

So in my opinion, there is absolutely no risk or negative side in my changes, and they allow for an automatic check for more serious failures in rst syntax.

sunqm (approved):

PR #1854, opened 2023-09-07
Simplified `symm.label_orb_symm` function

ChuixiongWu:
It seems that obtaining the mapping of molecular orbitals to symmetric orbitals does not require such a complex process?

### DIFF
--- a/pyscf/symm/addons.py
+++ b/pyscf/symm/addons.py
@@ -71,13 +71,14 @@ def label_orb_symm(mol, irrep_name, symm_orb, mo, s=None,
     norm = numpy.zeros((len(irrep_name), nmo))
     for i, csym in enumerate(symm_orb):
         moso = numpy.dot(csym.T.conj(), s_mo)
-        ovlpso = reduce(numpy.dot, (csym.T.conj(), s, csym))
-        try:
-            s_moso = lib.cho_solve(ovlpso, moso)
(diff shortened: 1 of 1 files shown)


pyscf/symm/addons.py
```diff
@@ -71,13 +71,14 @@ def label_orb_symm(mol, irrep_name, symm_orb, mo, s=None,
...
-            ovlpso[numpy.diag_indices(csym.shape[1])] += 1e-12
-            s_moso = lib.cho_solve(ovlpso, moso)
-        norm[i] = numpy.einsum('ki,ki->i', moso.conj(), s_moso).real
+        # ovlpso = reduce(numpy.dot, (csym.T.conj(), s, csym))
+        # try:
+        #     s_moso = lib.cho_solve(ovlpso, moso)
+        # except numpy.linalg.LinAlgError:
+        #     ovlpso[numpy.diag_indices(csym.shape[1])] += 1e-12
+        #     s_moso = lib.cho_solve(ovlpso, moso)
+        s_moso = numpy.dot(mo.T.conj(), csym)
```
sunqm, line 80: Both `mo` and `csym` are coefficients on AO basis. This transformation does not give any meaningful results.

sunqm:
You have to consider the basis orthogonality within each symmetry sector.

PR #1858, opened 2023-09-11
Add DCSD

verena-neufeld:
This is a draft pull request allowing the CCSD equations to be slightly modified into the DCSD equations if a DCSD class is chosen. kUDCSD is not implemented (also not planned by me). 

My questions are related to the style. I have made DCSD classes. Then
```
    @property
    def dcsd(self):
        return self.__class__.__name__[-4:] == "DCSD"
```
decides whether the corresponding CCSD equations should be modified based on the last four letters of the DCSD class being "DCSD", e.g. RDCSD.

I know that this is not 100% ideal but it seemed cleanest. Please let me know what suggestions you have about this and other aspects of this Wip PR.

### DIFF
--- a/pyscf/cc/ccsd.py
+++ b/pyscf/cc/ccsd.py
@@ -48,6 +48,7 @@ def kernel(mycc, eris=None, t1=None, t2=None, max_cycle=50, tol=1e-8,
         eris = mycc.ao2mo(mycc.mo_coeff)
     if t1 is None and t2 is None:
         t1, t2 = mycc.get_init_guess(eris)
+        t1 = numpy.zeros_like(t1)
     elif t2 is None:
         t2 = mycc.get_init_guess(eris)[1]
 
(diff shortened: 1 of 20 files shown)


tberkel:
For whatever it's worth, the existing standard for modified CC methods had
been to pass a kwarg (mbpt2 or cc2). Admittedly, I'm not sure I love this
choice, especially as more modifications get added to CC. Similarly, the
use of lots of `if` statements in CC to manage CC2, and now DCSD, is
beginning to feel clunky, although I appreciate that it prevents code
duplication. CC2 especially needs a refactor, since it doesn't require the
same integrals/intermediates as full CCSD.

verena-neufeld:
Thanks Tim! DCSD is very similar to CCSD, so I'm hesitant to write new t amp update equations for it. I agree, some refactoring would be useful. We also have ccsd.py, rccsd.py, rccsd_slow.py, all implementing restricted CCSD which can get confusing (but that is another topic maybe). Do you have suggestions? If you prefer, I can keep passing dcsd as a bool variable immediately instead of inferring from the class name.

tberkel:
I agree, for DCSD, it probably makes sense to keep it embedded in the CCSD
update. Rather than checking the class name, perhaps its better to be more
explicit and just have DCSD classes that inherit from CCSD classes but set
a "private" attribute like `self._dcsd = True` and then check this in the
update code? CC2 and MBPT2 probably need to be refactored, and we might do
that soon in our group. (Honestly, the whole cc module would benefit from a
refactor ... 😅)

fishjojo:
try `isinstance(mycc, DCSD)` and make all (R-, U-, K-,) DCSD classes inherit from DCSD.

Issue #1801, opened 2023-07-22
Can't pickle PySCF object?


sunnowrun:
I want to save the PySCF variables to the file, but I when I run the code below, I got error and don't know how to solve it.
`import pyscf`
`import pickle`

`geom = 'Li 0 0 0; H 0 0 1'`
`basis = 'sto-3g'`
`mol = pyscf.gto.M(atom = geom, basis = basis)`

`with open ('filename', 'wb') as f:`
`	pickle.dump(mol, f)`
Error:
`Traceback (most recent call last):`
`  File "pyscf_usage.py", line 12, in <module>`
`    pickle.dump(mol, f)`
`TypeError: cannot pickle '_io.TextIOWrapper' object`

hebrewsnabla:
You can use `pyscf.lib.chkfile.save_mol` to save mol object to a hdf5 file.

sunnowrun:
But I also want to save mf and other results.
If I separately save these different type objects, the saved results will become very messy.
I want to save them in a consistent way, can I modified the mol/mf object make it suitable to pickle?

hebrewsnabla:
> But I also want to save mf and other results. If I separately save these different type objects, the saved results will become very messy. I want to save them in a consistent way, can I modified the mol/mf object make it suitable to pickle?

`pyscf.scf.chkfile.dump_scf` saves the mol and some scf's arrays into one hdf5 file. This hdf5 file is interoperable so you can dump any other arrays into it with `lib.chkfile.save('chkfile', {'array_name': array})`

sunnowrun:
> > But I also want to save mf and other results. If I separately save these different type objects, the saved results will become very messy. I want to save them in a consistent way, can I modified the mol/mf object make it suitable to pickle?
> 
> `pyscf.scf.chkfile.dump_scf` saves the mol and some scf's arrays into one hdf5 file. This hdf5 file is interoperable so you can dump any other arrays (or dict) into it with `lib.chkfile.save('chkfile', 'array_name', array)`

Why so inconvenient?

hebrewsnabla:
I think that's convenient enough. Usually `dump_scf` is enough for dumping a scf object. Do you have anything else to be dumped?

sunnowrun:
> I think that's convenient enough. Usually `dump_scf` is enough for dumping a scf object. Do you have anything else to be dumped?

[...]

Issue #1737, opened 2023-05-18
Inconsistent UHF conversion functions in open- and periodic boundary conditions 


chrislzhao:
Hi, 

I found an issue when converting to UHF object in the MP2 function. For both periodic systems at Gamma point and molecular systems, it is the molecular MP2 function that will be used when doing a MP2 calculation. When the SCF object is, say ROHF, it will be converted to UHF by the `to_uhf` function. However, the implementation of the `to_uhf` function is not consistent with molecular SCF v.s. periodic SCF, in which the latter requires an input of a SCF object while the former does not. Due to this problem, the code will fail when one tries to do a PBC-MP2 with PBC-ROHF, as when it calls `to_uhf` it complains about the missing argument.

sunqm:
The conversion function is messy in the main branch. I have them refactored in a development branch https://github.com/sunqm/pyscf/tree/mcfun-pbc . You can try this branch if you need the fix for now. The fix depends on PR #1622. So it may take some time to merge that branch

verena-neufeld:
Hi @chrislzhao , thanks! Has this been solved now by PR #1859 ?

sunqm:
#1859 has unified the conversion functions between SCF methods.

PR #1859, opened 2023-09-11
New features and code refactoring

sunqm:
Add new features:
* multi-collinear functional for PBC DFT
* non-local functional (VV10) for PBC DFT
* "undo" method for dynamic classes. This method can revert the action for conversion methods such as "density_fit()", "x2c()", "newton()", "as_scanner()", etc.
* Merged basis parser for molecular GTO basis and GTH basis. GTH basis can be assigned to Mole object directly.
* Merged ECP  and PP parser. PP can be assigned to Mole object directly.

Refactor:
* The layout of the Mole class and Cell class. Remove the inheritance between Cell and Mole.
* The layout of various SCF classes and the conversion methods (to_ks, to_hf, to_uhf, to_ghf, etc.) between different SCF objects.
* The layout of various MCSCF classes.
* New style to generate dynamic class for DF methods, X2C, QMMM, SOSCF, Solvent model, StateAverageMCSCF, Scanner methods, etc.
* SCF smearing method.
* Make Mole and Cell object picklable (issue #1801, #1806)
* Fix various deprecated warnings from numpy


verena-neufeld:
I have only looked through parts of this. Thanks so much Qiming! The CC refactoring with a base class is a good idea. Do you think this is also a good idea for MP2 (if I missed this, please let me know)? When you add commented out tests, do you intend for them to stay commented? Again, thanks!

sunqm:
> I have only looked through parts of this. Thanks so much Qiming! The CC refactoring with a base class is a good idea. Do you think this is also a good idea for MP2 (if I missed this, please let me know)? When you add commented out tests, do you intend for them to stay commented? Again, thanks!

Sure. The existing class hierarchy may not be convenient enough for the complex features. Please feel free to refactor them, like the MP2 class. I didn't update throughout all modules. I only updated the modules which seems too messy.

Which tests do you mean that was commented out? I moved some code under `__name__ == '__main__'` to unittests so that they will be run by the pytest pipeline.

pyscf/dft/test/test_r_numint.py
```diff
@@ -116,7 +116,75 @@ def test_vxc1(self):
         res = mf._numint.get_vxc(mol, mf.grids, mf.xc, dm, spin=0)
         self.assertAlmostEqual(res[1], -8.631807003163278, 11)
 
-#    def test_fxc(self):
```
verena-neufeld, line 119: This is a file I meant that has commented out tests

verena-neufeld:
Thanks @sunqm , that makes sense! I have commented in a file that has commented out tests.

verena-neufeld:
This PR might help issue #1737

fishjojo (approved):

Issue #1833, opened 2023-08-15
pycodestyle is not happy


fchapoton:
Here are the current warnings:
```

34      E111 indentation is not a multiple of 4
3       E114 indentation is not a multiple of 4 (comment)
53      E115 expected an indented block (comment)
3       E116 unexpected indentation (comment)
1       E117 over-indented
4       E121 continuation line under-indented for hanging indent
437     E122 continuation line missing indentation or outdented
1       E123 closing bracket does not match indentation of opening bracket's line
37      E124 closing bracket does not match visual indentation
94      E126 continuation line over-indented for hanging indent
1543    E127 continuation line over-indented for visual indent
563     E128 continuation line under-indented for visual indent
147     E129 visually indented line with same indent as next logical line
1179    E201 whitespace after '('
1537    E202 whitespace before ']'
8712    E203 whitespace before ','
1502    E211 whitespace before '['
1619    E221 multiple spaces before operator
544     E222 multiple spaces after operator
3140    E225 missing whitespace around operator
20486   E226 missing whitespace around arithmetic operator
44      E227 missing whitespace around bitwise or shift operator
207     E228 missing whitespace around modulo operator
69286   E231 missing whitespace after ','
1268    E241 multiple spaces after ','
2056    E251 unexpected spaces around keyword / parameter equals
1199    E261 at least two spaces before inline comment
444     E262 inline comment should start with '# '
2805    E265 block comment should start with '# '
157     E266 too many leading '#' for block comment
5       E272 multiple spaces before keyword
55      E275 missing whitespace after keyword
186     E301 expected 1 blank line, found 0
3466    E302 expected 2 blank lines, found 1
174     E303 too many blank lines (2)
587     E305 expected 2 blank lines after class or function definition, found 1
296     E306 expected 1 blank line before a nested definition, found 0
8       E401 multiple imports on one line
117     E402 module level import not at top of file
1       E502 the backslash is redundant between brackets
1388    E701 multiple statements on one line (colon)
2       E703 statement ends with a semicolon
6       E722 do not use bare 'except'
72      E731 do not assign a lambda expression, use a def
164     W291 trailing whitespace
1       W292 no newline at end of file
26      W293 blank line contains whitespace
169     W391 blank line at end of file

```

verena-neufeld:
Thank you! Can this be closed now or is it not all fixed yet?

fchapoton:
Not all fixed, no. There remains *many* warnings. I only fixe the few easiest ones.

[...]

PR #1872, opened 2023-09-19
add pycodestyle to linter workflow

fchapoton:
also refreshing the linter on the way

This may help to make progress on #1833

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -4,17 +4,32 @@ on: [push, pull_request]
 
 jobs:
   flake:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
-      - name: Setup Python
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.7
-      - uses: actions/checkout@v2
-      - name: Install flake8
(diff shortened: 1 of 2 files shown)


fchapoton:
the existing flake linter reveals an issue

verena-neufeld:
Is your plan to merge this after the current issues highlighted by pycodestyle have been fixed so we can test new code? That might be nicer so that the merge doesn't immediately lead to new test failures on the main branch.

fchapoton:
as you can see in the modified file, pycodestyle is only testing for the E111 warning, which is already fixed, and the pycodestyle linter is green. The plan would be to activate other fixes at the same time that they are fixed, to avoid moving backward.

fchapoton:
could someone please help fix the flake8 warning

```
pyscf/scf/addons.py:125:9: F811 redefinition of unused 'get_grad' from line 73
Error: Process completed with exit code 1.
```

verena-neufeld:
That sounds like a good plan!

fchapoton:
this should be ready to go, please approve and merge if you think this is a good idea

fchapoton:
the CI failure seems not to be related

sunqm (approved):

PR #1883, opened 2023-10-01
fix and activate pycodestyle W293

fchapoton:
This is fixing the pycodestyle warning W293 about spaces in empty lines, and adding the check to the linter.


fchapoton:
test failure seems not to be related to the changes made here

sunqm (approved):

Issue #973, opened 2021-06-05
pyscf_lib_placeholder.so is written into plist but isn't installed


yurivict:
```setup.py install --record {plist}``` writes ```lib/python3.8/site-packages/pyscf_lib_placeholder.so``` into plist but doesn't install it.

FreeBSD ports framework complains:
```
====> Checking for pkg-plist issues (check-plist)
===> Parsing plist
===> Checking for items in STAGEDIR missing from pkg-plist
===> Checking for items in pkg-plist which are not in STAGEDIR
Error: Missing: %%PYTHON_SITELIBDIR%%/pyscf_lib_placeholder.so
===> Error: Plist issues found.
```

verena-neufeld:
Hi @yurivict , thanks for having raised this issue. Since this was a few releases ago, has this been resolved by now?

sunqm:
Fix in PR #1900

Issue #934, opened 2021-05-05
A missing file is placed into the PLIST


yurivict:
```
===> Checking for items in pkg-plist which are not in STAGEDIR
Error: Missing: %%PYTHON_SITELIBDIR%%/pyscf_lib_placeholder.so
===> Error: Plist issues found.
```

```pyrthon setup.py install --record plist``` produces the plist file with ```pyscf_lib_placeholder.so``` that is not installed.

verena-neufeld:
Hi @yurivict , thanks for having raised this issue. Since this quite a few releases ago, has it been resolved?

sunqm:
Fix in PR #1900

PR #1900, opened 2023-10-14
Improve wheel

sunqm:
* Fix issue #973 #934 
* Improve bdist wheel compilation. Release py3 type wheel to support all Python3 versions.
* Remove some CI test pipelines


fishjojo (approved):

hebrewsnabla:
Does the modification to `build-wheels.sh` mean that there will be only py39 wheel published for linux in next release?

sunqm:
No. It just uses py39 toolkit to perform the build process. The new script builds a general wheel that supports all python3 platforms.

PR #1905, opened 2023-10-17
Allow to turn off AO symmetry for PBC KS calculations

fishjojo:


sunqm:
Please add a test to cover `use_ao_symmetry=False`

sunqm (approved):

PR #1916, opened 2023-10-23
native support dispersion correction

wxj6000:
- support d3 and d4 via dftd3 and dftd4 respectively
- support scf, gradient, hessian


pyscf/scf/hf.py
```diff
@@ -933,6 +936,43 @@ def get_veff(mol, dm, dm_last=None, vhf_last=None, hermi=1, vhfopt=None):
...
+
+    # for dftd4
+    elif disp_version[:2].upper() == 'D4':
+        from pyscf.data.elements import charge
+        atoms = numpy.array([ charge(a[0]) for a in mol._atom])
+        coords = mol.atom_coords()
+        try:
+            from dftd4.interface import DampingParam, DispersionModel
+        except ImportError:
+            print("cannot find dftd3 in the current environment. Please install dftd3")
```
hebrewsnabla, line 968: should be dftd4

pyscf/scf/dispersion.py
```diff
@@ -0,0 +1,80 @@
...
+    mol = mf.mol
+    if disp_version is None:
+        return 0.0
+    if isinstance(mf, KohnShamDFT):
+        method = mf.xc
+    else:
+        method = 'hf'
+
+    # for dftd3
+    if disp_version[:2].upper() == 'D3':
```
sunqm, line 39: `disp_version` may be `None`, as the default value in `__init__` of SCF object. Similar problems are in the grad, hessian and other relevant modules.

wxj6000: Right. `disp_version` is `None` by default. That is, there is no dispersion correction by default. The dispersion correction is added when `disp` is assigned to `mf` explicitly, such as `mf.disp = "d3bj"`.

sunqm (approved):

PR #1922, opened 2023-10-27
apply "ruff --fix check --select=C40 pyscf"

fchapoton:
to fix many little  details about list comprehension, dict, sets, etc


verena-neufeld:
Thanks @fchapoton for cleaning up pyscf! Out of interest, why are {} and {:} preferable over set() and dict()?

fchapoton:
well, usually shorter and convenient

sunqm:
As long as python supports the `set()` and `dict()` function, changing to set comprehension and dict comprehension does not look a necessary to me.

fchapoton:
ok, so should I just close the pull request ?

sunqm:
I'll merge this PR. For future PRs, unless updating the deprecated syntax, simply replacing one syntax by another is not encouraged.

PR #1930, opened 2023-10-31
4th, and 5th order XC derivatives

sunqm:
* Add a patch to xcfun to compute up to 5th order XC derivatives
* Update xcfun_itrf for up to 5th order derviatives
* Update libxc_itrf to support 4th order XC derivatives
* Add the transform_xc function to support arbitrary order XC derivatives
* Update conda build script


susilehtola:
What do you need 5th order derivatives for, should we also generate these in libxc in a later release...?

sunqm:
> What do you need 5th order derivatives for, should we also generate these in libxc in a later release...?

It is mainly for the non-collinear/multi-collinear functional methods. The multi-collinear functional requires a higher order derivative than the regular xc functional derivatives. Constructing the 4th order multi-collinear derivative requires 5th order XC derivative. TDDFT gradients  with multi-collinear functional are under development. It requires the 4th order XC derivative. I just add an additional feature in case developing TDDFT hessian in the future. I guess 5th order in libxc is not needed for the moment.

fishjojo (approved):

Issue #1961, opened 2023-11-17
Re-ordering basis functions introduces mismatch


kevfocke:
#1620 introduced ordering basis functions if they are not ordered correctly. When using `molden.load` to load molecules from molden files, the MO coefficients are not ordered accordingly, which leads to a mismatch and therefore wrong results. 
One example would be the minimally augmented basis sets (like ma-def2-SVP) from ORCA, which have some s-type AOs appended after the p-type AOs from the standard def2-SVP.

The code has changed significantly since #1620, which makes fixing the problem a bit harder.

hebrewsnabla:
I tested an example with aug-cc-pvtz using pyscf 2.2.0, which is released before 1620, and it shows that the basis is still re-ordered. So I guess this re-ordering is not a feature introduced in recent versions but much earlier?

verena-neufeld:
Hi @kevfocke , thanks! Has it been fixed now and can this issue be closed?

kevfocke:
@verena-neufeld Yes, sure, thank you!

PR #1963, opened 2023-11-18
Fix basis order issue in molden.load (issue #1961)

sunqm:


pyscf/gto/mole.py
```diff
@@ -503,11 +509,6 @@ def converter(symb, raw_basis):
                     _basis.extend(nparray_to_list(rawb))
         else:
             _basis = nparray_to_list(raw_basis)
-
-        # Sort basis accroding to angular momentum. This is important for method
-        # decontract_basis, which assumes that basis functions with the same
-        # angular momentum are grouped together. Related to issue #1620 #1770
-        _basis = sorted([b for b in _basis if b], key=lambda b: b[0])
```
kevfocke, line 510: Oh wow, I didn't find this line when I was looking for it. Thank you so much!

pyscf/tools/molden.py
```diff
@@ -239,8 +239,16 @@ def read_one_bas(lsym, nb, fac=1):
...
-    mol.basis = envs['basis'] = basis
     mol.atom = [atoms[i] for i in atom_seq]
+    uniq_atoms = {a[0] for a in mol.atom}
+
+    # To avoid the mol.build() sort the basis, disable mol.basis and set the
+    # internal data _basis directly. It is a workaround to solve issue #1961.
+    # Mole.decontract_basis function should be rewritten to support
+    # discontinuous bases that have the same angular momentum.
+    mol.basis = {}
+    _basis = gto.mole._parse_default_basis(basis, uniq_atoms)
```
kevfocke, line 250: I guess this somewhat limits what one can do with objects retrieved from molden files, depending on how many functions depend on the sorted basis functions and how common these non standard sortings are.

pyscf/tools/test/test_molden.py
kevfocke, line 1: The test seems sensible and simple, although it also fails, when both the basis and the mo_coeffs would hypothetically be sorted the same way successfully. So if that is ever implemented, one has to keep in mind that the tested behavior is not necessarily the perfect behavior.

kevfocke (approved):
Thank you so much! This is a simple fix without much additional code. 
I tested it for my use case and it works as expected.

pyscf/tools/molden.py
sunqm: The case of un-sorted basis was considered here. Calling mol.build() will enforce a sort in the function. However, assigning the ._basis attribute of the Mole object directly will skip the sorting. Probably a better solution is to use an input parameter in Mole object to control whether to sort the basis. Introducing the new parameters may require other updates in Mole class. To solve the issue 1961, we can use this temporary solution for the moment

fishjojo (approved):

Issue #1984, opened 2023-12-02
Gradient of Pipek-Mezey orbital localization is off by a factor of 2


hongzhouye:
Hi, I found that the `get_grad` function of `lo.pipek.PipekMezey` gives gradients that seem to be off by a factor of 2 compared to numerical gradients. Is there a reason for that?

For example, these are results for water/sto-3g.
```
Exponent = 2
        PySCF     PySCFx2   Num grad 
0      -0.00000  -0.00001  -0.00001
1       0.00000   0.00000  -0.00000
2       0.00000   0.00000  -0.00000
3      -0.00000  -0.00000  -0.00000
4       0.00000   0.00000   0.00000
5       0.00000   0.00000  -0.00000
6       0.00000   0.00001   0.00001
7       0.00000   0.00001   0.00001
8      -0.00000  -0.00000  -0.00000
9      -0.00000  -0.00000  -0.00000
10     -0.14417  -0.28833  -0.28833
11     -0.01879  -0.03759  -0.03759
12      0.00000   0.00000  -0.00000
13      0.19842   0.39684   0.39684
14     -0.15528  -0.31056  -0.31056
15     -0.14417  -0.28833  -0.28833
16     -0.01879  -0.03759  -0.03759
17     -0.00000  -0.00000  -0.00000
18     -0.19842  -0.39684  -0.39684
19     -0.15528  -0.31056  -0.31056
20      0.00000   0.00000  -0.00000

Exponent = 4
        PySCF     PySCFx2   Num grad 
0      -0.00001  -0.00002  -0.00002
1       0.00000   0.00000  -0.00000
2       0.00000   0.00000  -0.00000
3      -0.00000  -0.00000  -0.00000
4       0.00000   0.00000   0.00000
5       0.00000   0.00000  -0.00000
6       0.00001   0.00003   0.00003
7       0.00001   0.00003   0.00003
8      -0.00000  -0.00000  -0.00000
9      -0.00000  -0.00000   0.00000
10     -0.28698  -0.57397  -0.57396
11     -0.03749  -0.07497  -0.07497
12      0.00000   0.00000  -0.00000
13      0.39428   0.78856   0.78855
14     -0.30901  -0.61802  -0.61802
15     -0.28698  -0.57397  -0.57396
16     -0.03749  -0.07497  -0.07497
17     -0.00000  -0.00000  -0.00000
18     -0.39428  -0.78856  -0.78855
19     -0.30901  -0.61802  -0.61802
[...]

PR #1648, opened 2023-03-20
Boys localization jacobi sweep

SUSYUSTC:
In many systems, I can find that Jacobi sweep (i.e. exact optimization in a two-orbital subspace each time) can lead to a lower cost function than the current optimization result of the Boys localization, so I hope to have this pull request to add the functionality of Jacobi sweep as an option. There's also a related issue about Boys localization at https://github.com/pyscf/pyscf/issues/820.
Here's an example showing that Jacobi sweep gives more localized orbitals in a 3-water system (valid for the modified code).

```
import pyscf
import pyscf.lo

atom = '''
O 2.07091 -0.46191 -0.00993;
H 2.56267 -0.75858 0.76451;
H 2.70113 -0.40578 -0.73813;
O 0.0 0.91527 -0.05817;
H -0.87302 0.35992 0.01707;
H 0.87302 0.35993 0.01707;
O -2.07092 -0.4619 -0.00993;
H -2.70115 -0.40575 -0.73811;
H -2.56265 -0.75862 0.76451;
'''


mol = pyscf.gto.M(atom=atom, basis='cc-pVTZ')

rhf = pyscf.scf.RHF(mol).density_fit()
rhf.run()
C = rhf.mo_coeff

print('jacobi sweep')
boys = pyscf.lo.Boys(mol, C.copy()[:, 3:15])
boys.jacobi = True
Cloc1 = boys.kernel(verbose=4)

print('original optimization')
boys = pyscf.lo.Boys(mol, C.copy()[:, 3:15])
Cloc2 = boys.kernel(verbose=4)
```

The output I get is 
```
converged SCF energy = -228.137524073691
jacobi sweep
Set conv_tol_grad to 0.000316228
macro= 1  f(x)= -387.32837908711  res= 0.30631398498355
macro= 2  f(x)= -388.24630408055  res= 0.090436807320824
macro= 3  f(x)= -388.57519368992  res= 0.043949165770448
macro= 4  f(x)= -388.60894029858  res= 0.023533942030739
macro= 5  f(x)= -388.61551796336  res= 0.0079208038076226
macro= 6  f(x)= -388.61629827424  res= 0.0027455288434352
macro= 7  f(x)= -388.61639111717  res= 0.000889717664449
macro= 8  f(x)= -388.616401279  res= 0.00028671113884098
macro= 9  f(x)= -388.61640239729  res= 9.1383331136231e-05
macro= 10  f(x)= -388.61640252295  res= 2.9070033855052e-05
[...]

Issue #820, opened 2021-02-10
Delocalized orbitals after Foster-Boys localization


wangenau:
Hello,
I used the following script to calculate localized orbitals of an ethane
molecule using the Foster-Boys localization method. When looking at the
orbitals, numbers 4 and 5 came out delocalized (figures can be found at the
bottom) and differ from the expected form that can be obtained using, e.g.,
ERKALE. All other orbitals agree perfectly. 

When using [this PyCOM procedure](https://github.com/pyflosic/pyflosic/blob/master/src/pycom.py) with a stability analysis, the [ERKALE](https://github.com/susilehtola/erkale)
results can be reproduced.

```
from pyscf import gto, dft
from pyscf.lo.boys import Boys
atom = '''H      1.1851     -0.0039      0.9875
          C      0.7516     -0.0225     -0.0209
          H      1.1669      0.8330     -0.5693
          H      1.1155     -0.9329     -0.5145
          C     -0.7516      0.0225      0.0209
          H     -1.1669     -0.8334      0.5687
          H     -1.1157      0.9326      0.5151
          H     -1.1850      0.0044     -0.9875'''
mol = gto.M(atom=atom, basis='sto-3g')
mf = dft.RKS(mol)
mf.xc = 'pbe,pbe'
mf.kernel()
mo = Boys(mol, mf.mo_coeff[:, mf.mo_occ > 0])
mo.kernel()
print(mo.mo_coeff)
```

Also, one can show that the final Hessian has negative eigenvalues, i.e. with
verbose=9 you see eig= -4.64e-11. As far as I understand the details of [CIAH](https://arxiv.org/abs/1610.08423) 
the eigenvalues should all be positive definite at the end of the localization.

For debugging I tried to recalculate the Hessian and also found negative
eigenvalues.

```
import numpy
u = mo.get_init_guess()
g, h_op, h_diag = mo.gen_g_hop(u=u)
hessian = numpy.diag(h_diag)
hval, hvec = numpy.linalg.eigh(hessian)
hval_min = hval.min()
print(hval_min)
```

This is not a singularity. The problem exists for various other systems and
basis sets. I only used ethane as a simple, illustrative example.

Any hints or ideas on how one can overcome this issue are highly appreciated. 

These findings exist in various PySCF versions, but the results discussed here
are produced with the following package versions: 

PySCF version: 1.7.5.1
Numpy version: 1.19.2

Best, 
Wanja 

Delocalized orbitals:
[...]

PR #1990, opened 2023-12-05
Bug fix for Pipek Mezey + Stability check using Jacobi sweep

hongzhouye:
## Bug fix for Pipek Mezey

This PR fixes issue #1984. In particular, the missing factor of 2 in both gradient and Hessian is restored. The erroneous `h_diag` for exponent = 4 is fixed (as a result, some calculations with exponent = 4 require fewer cycles to converge with the new code). The new implementation supports arbitrary exponent (not limited to 2 and 4). Tests are added that examine if the new code generates gradient and Hessian that agree with finite difference.

A bug in the atomic initial guess (in `boys.py`) is also fixed: the molecular `s1e` was used for the pbc case, which is fixed now.

## Stability check of PM orbitals using Jacobi sweep

The second-order CIAH algorithm along with the atomic initial guess used by the PM code can sometimes result in convergence to a local minimum with orbitals that have high symmetry. This occurs for systems as simple as the occupied orbitals of an isolated water molecule. This PR also implements a Jacobi sweep algorithm that can break the said high symmetry in the localized orbitals and help escape from local minima. The cost is roughly two calls of `atomic_pops()`, which is very modest. The orbitals it generates serve as a better initial guess for PM optimization. A typical use case is
```Python
mlo = PM(mol, mo_coeff)
mlo.kernel()
isstable, mo_new = mlo.stability_jacobi()
Whie not isstable:
    mlo.kernel(mo_new)
    isstable, mo_new = mlo.stability_jacobi()
```
Usually one Jacobi check is enough for the subsequent PM optimization to converge to a "stable" minimum (stable w.r.t. Jacobi rotations). Two localized orbitals from water/cc-pVDZ are shown below.

Screenshot 2023-12-04 at 22 53 27


sunqm (approved):

sunqm:
I think the CIAH minimization converges to the local minimum probably due to the symmetry. The initial guess in your example are symmetry adapted orbitals. If the initial guess breaks the symmetry in molecular orbitals, the CIAH code should also converge the results the same to the Jacobi sweep.

hongzhouye:
@sunqm thanks! And completely agreed on the initial guess 👍

susilehtola:
Is  #820 also affected by this?

The issue with wrong symmetry is well-known and affects most solvers: orbitals of different symmetries can't mix in the localization. In this case the standard solution is to slightly perturb the initial guess with a random matrix. Should this be done by default?

hongzhouye:
@susilehtola Thanks for the note! I could reproduce #820 and verified that using converged PM orbitals from the this PR (which break the symmetry) as initial guess for Boys localization resolves the problem in #820. This is an indirect solution. A direct solution may use jacobi-sweep Boys to break the symmetry (PR #1648).

Screenshot 2023-12-09 at 11 31 07

Issue #1988, opened 2023-12-04
Potential bug in non-canonical UMP2


hungpham2017:
**Description**: I ran UMP2 and RMP2 on the same _close-shell_ carbon atom using arbitrary orbitals (non-converged SCF) and expected UMP2 and RMP2 return the same energy, but they were not.

**Example:**

```
from pyscf import gto, scf, mp

mol = gto.Mole(atom = [['C', 0,0,0]], basis = 'cc-pvdz')
mf = scf.RHF(mol)
mf.kernel()

# Running non-canonical MP2
mf.mo_coeff[:2:,2:] = 0
mf.converged = False
e_ump2 = mf.to_uhf().MP2().kernel()[0]
e_rmp2 = mf.MP2().kernel()[0]
print("E(UMP2) = ", e_ump2)
print("E(RMP2) = ", e_rmp2)
```

**Result**:
```
E(UMP2) =  0.6143361818889358
E(RMP2) =  0.6127504763660901
```
The energies are different. I found that the [update_amps](https://github.com/pyscf/pyscf/blob/148a6fc671e4ba17f64ef3ff74e5676c8de2cb8b/pyscf/mp/ump2.py#L129) function used in non-canonical UMP2 doesn't update the amplitudes when executed by the [_iterative_kernel](https://github.com/pyscf/pyscf/blob/148a6fc671e4ba17f64ef3ff74e5676c8de2cb8b/pyscf/mp/mp2.py#L80).
In line [157-159](https://github.com/pyscf/pyscf/blob/148a6fc671e4ba17f64ef3ff74e5676c8de2cb8b/pyscf/mp/ump2.py#L157), simply change:
```
    u2aa = eris_ovov.transpose(0,2,1,3) - eris_ovov.transpose(0,2,3,1)
    u2bb = eris_OVOV.transpose(0,2,1,3) - eris_OVOV.transpose(0,2,3,1)
    u2ab = eris_ovOV.transpose(0,2,1,3)
```
to 

```
    u2aa += eris_ovov.transpose(0,2,1,3) - eris_ovov.transpose(0,2,3,1)
    u2bb += eris_OVOV.transpose(0,2,1,3) - eris_OVOV.transpose(0,2,3,1)
    u2ab += eris_ovOV.transpose(0,2,1,3)
```
ran the same code again, the energies now agree:
```
E(UMP2) = 0.6127504763657433
E(RMP2) = 0.6127504763660907
```

[...]

PR #1991, opened 2023-12-05
Bug fix in interactive solver for non-canonical UMP2

hungpham2017:
This PR fixes a bug in the iterative non-canonical UMP2 calculation (described in  #1988), causing discrepancy in UMP2 and RMP2 for a close-shell carbon atom.


fishjojo:
Please fix the test

PR #2005, opened 2023-12-23
Enable numpy array form of mo_coeff and mo_occ in get_k_kpts in df_jk.py

zhcui:
Dear all,

I found that if the user gives a set of mo_occ or mo_coeff (in numpy array form) to the kmf with GDF, the current will raise an error.  It assumes the mo_coeff and mo_occ as list or tuple.

This PR fixes this.

Best,
Zhihao

### DIFF
--- a/pyscf/pbc/df/df_jk.py
+++ b/pyscf/pbc/df/df_jk.py
@@ -297,14 +297,14 @@ def get_k_kpts(mydf, dm_kpts, hermi=1, kpts=numpy.zeros((1,3)), kpts_band=None,
     if not mydf.force_dm_kbuild:
         if mo_coeff is not None:
             if isinstance(mo_coeff[0], (list, tuple)) or (isinstance(mo_coeff[0], numpy.ndarray)
-                                                          and mo_coeff[0] == 3):
+                                                          and mo_coeff[0].ndim == 3):
                 mo_coeff = [mo for mo1 in mo_coeff for mo in mo1]
             if len(mo_coeff) != nset*nkpts: # wrong shape
                 log.warn('mo_coeff from dm tag has wrong shape. '
                          'Calculating mo from dm instead.')
                 mo_coeff = None
             elif isinstance(mo_occ[0], (list, tuple)) or (isinstance(mo_occ[0], numpy.ndarray)
-                                                          and mo_occ[0] == 2):
+                                                          and mo_occ[0].ndim == 2):
                 mo_occ = [mo for mo1 in mo_occ for mo in mo1]
         if mo_coeff is not None:
             skmoR, skmoI = _format_mo(mo_coeff, mo_occ, shape=(nset,nkpts), order='F',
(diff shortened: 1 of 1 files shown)


verena-neufeld:
Hi @zhcui , thanks again for this PR. I wonder whether this PR will break other code in pyscf or break some of pyscf's pbc code conventions with k point sampling, e.g. a list for mo_coeff where each entry is the mo_coeff of a particular k point or (not 100% here) a tuple for unrestricted calculations. Does the `mo_coeff[0] == 3` etc prevent code from breaking? Should this be updated in other parts of pyscf, too, not just GDF?

sunqm:
The convention of pbc `mo_coeff` is [spin-index, k-index, ao-index, mo-index]. Testing the dimension of `mo_coeff[0]` looks no problem to me.

sunqm:
> Hi @zhcui , thanks again for this PR. I wonder whether this PR will break other code in pyscf or break some of pyscf's pbc code conventions with k point sampling, e.g. a list for mo_coeff where each entry is the mo_coeff of a particular k point or (not 100% here) a tuple for unrestricted calculations. Does the `mo_coeff[0] == 3` etc prevent code from breaking? Should this be updated in other parts of pyscf, too, not just GDF?

@zhcui do the aft_jk and fft_jk modules have the similar issue? If so, could you help to fix these modules too?

verena-neufeld:
Thanks @sunqm for clarifying!

pyscf/pbc/df/df_jk.py
```diff
@@ -296,13 +296,15 @@ def get_k_kpts(mydf, dm_kpts, hermi=1, kpts=numpy.zeros((1,3)), kpts_band=None,
     skmoR = skmo2R = None
     if not mydf.force_dm_kbuild:
         if mo_coeff is not None:
-            if isinstance(mo_coeff[0], (list, tuple)):
+            if isinstance(mo_coeff[0], (list, tuple)) or (isinstance(mo_coeff[0], numpy.ndarray)
+                                                          and mo_coeff[0].ndim == 3):
```
verena-neufeld, line 300: Is this check for unrestricted calculations or does this work for restricted ones, too, as ndim of  mo_coeff[0] changes then maybe?

verena-neufeld:
I've just checked and such a check isn't in the aft_jk and fft_jk files, so they might be ok, but I haven't tested. Sorry for not looking into this before, @zhcui !

pyscf/pbc/df/df_jk.py
zhcui: I think this is only related to the unrestricted calculation. i.e. for rhf / rks, it will not have ndim == 3

zhcui:
> I've just checked and such a check isn't in the aft_jk and fft_jk files, so they might be ok, but I haven't tested. I'm sorry for not looking into this before posting my comment here, @zhcui !

Thanks, Verena! I think you are correct. I also don't see the same issue in the fft_jk or aft_jk. It is related to the format_mo in the df_jk

pyscf/pbc/df/df_jk.py
verena-neufeld: Makes sense, thanks!

sunqm (approved):

Issue #2002, opened 2023-12-18
gccsd error when using an ERI with size (nmo, nmo, nmo, nmo)


zhcui:
Hi all,

I found a possible bug in the gccsd.ao2mo:

https://github.com/pyscf/pyscf/blob/e3ea7d1a095c6b7329bb64ddf9feb4cc567f87df/pyscf/cc/gccsd.py#L364

which missed a reshape:
` 
    if eri.size == nmo**4:  # if mycc._scf._eri is a complex array

        eri = eri.reshape(nmo**2, nmo**2)
    
        sym_forbid = (orbspin[:,None] != orbspin).ravel()`

Otherwise, the following `sym_forbid` mask would not work.

Best,
Zhihao

verena-neufeld:
Hi Zhihao, thanks for posting this. Just to check: the `shape` of `eri` is (nmo,nmo,nmo,nmo) not (nmo**2,nmo**2)? I just read some ao2mo kernel examples which seem to give out 2D arrays. Do you have a small example?
Thanks!

zhcui:
> Hi Zhihao, thanks for posting this. Just to check: the `shape` of `eri` is (nmo,nmo,nmo,nmo) not (nmo^2,nmo^2)? I just read some ao2mo kernel examples which seem to give out 2D arrays. Do you have a small example? Thanks!

https://github.com/pyscf/pyscf/blob/0bc590ffddec98eec5d2513a3d85274b7a791bc6/pyscf/ao2mo/incore.py#L126

This gives the shape of ijkl

verena-neufeld:
I think this is a larger issue than adding a `reshape` in `pyscf/pyscf/cc/gccsd.py`. The docstrings in the `ao2mo` functions indicate that a 2D array is array. So I think the return of https://github.com/pyscf/pyscf/blob/0bc590ffddec98eec5d2513a3d85274b7a791bc6/pyscf/ao2mo/incore.py#L126 should be changed or the docstrings have to be updated. I'll see what other returns of ao2mo there are.

For complex GCCSD, there is also the wrapper ao2mo in https://github.com/pyscf/pyscf/blob/0bc590ffddec98eec5d2513a3d85274b7a791bc6/pyscf/pbc/cc/ccsd.py#L107 which does have a reshape actually, but I think we should make sure the output of ao2mo is consistent.

verena-neufeld:
[...]

PR #2006, opened 2023-12-23
Fix the shape bug in make_eris_incore in gccsd.py

zhcui:
As described in issues #2002


verena-neufeld:
Hi @zhcui , thanks for this fix in response to the issue #2002 . I do think that ideally we shouldn't fix this in gccsd.py but rather in ao2mo/incore.py as that would make the output of ao2mo/incore.py consistent and also consistent with the docstring. I think that would keep the code cleaner. What do you think? @sunqm @fishjojo

verena-neufeld:
If ao2mo/incore.py does prove to be too tricky to change, I think this PR can be a temporary fix (but would be nice to get to the bottom of the underlying problem).

sunqm (approved):

Issue #1986, opened 2023-12-03
geometry optimization with dispersion correction


LKFEIYI:
Hi,
I tried to optimize a molecule with B3LYP-D3/D4 ([Simple DFT-D3](https://dftd3.readthedocs.io/en/latest/index.html)). But there are some errors. 

```
from pyscf import gto, dft, scf
import dftd3.pyscf as disp
from pyscf.geomopt.geometric_solver import optimize

#no spin RHF
try:
    mol = gto.M(atom='O 0 0 0; O 0 0 1.8', unit='ANG', basis='631g*')
    mf = scf.RHF(mol)
    #mf.xc='b3lyp'
    grad=disp.energy(mf).nuc_grad_method().as_scanner()
    grad.optimizer().kernel()
    print("no spin RHF: OK")
except Exception as e:
    print(e)


#spin=1 hydroxides RHF
try:
    mol = gto.M(atom='H 0 0 0; O 0 0 1.8', unit='ANG', basis='631g*', spin=1)
    mf = scf.RHF(mol)
    #mf.xc='b3lyp'
    grad=disp.energy(mf).nuc_grad_method().as_scanner()
    grad.optimizer().kernel()
    print("spin=1 RHF: OK")
except Exception as e:
    print(e)

#spin=0 RKS
try:
    mol = gto.M(atom='O 0 0 0; O 0 0 1.8', unit='ANG', basis='631g*')
    mf = dft.RKS(mol)
    mf.xc='b3lyp'
    grad=disp.energy(mf).nuc_grad_method().as_scanner()
    grad.optimizer().kernel()
    print("spin=0 RKS: OK")
except Exception as e:
    print(e)

```
The above optimization works.

```
#spin=1 hydroxides RKS
try:
    mol = gto.M(atom='H 0 0 0; O 0 0 1.8', unit='ANG', basis='631g*', spin=1)
    mf = dft.RKS(mol)
    mf.xc='b3lyp'
    grad=disp.energy(mf).nuc_grad_method().as_scanner()
    grad.optimizer().kernel()
    print("spin=1 RKS: OK")
except Exception as e:
    print(e)

#spin=2  RKS
try:
    mol = gto.M(atom='O 0 0 0; O 0 0 1.8', unit='ANG', basis='631g*', spin=2)
    mf = dft.RKS(mol)
    mf.xc='b3lyp'
    grad=disp.energy(mf).nuc_grad_method().as_scanner()
    grad.optimizer().kernel()
[...]

PR #2010, opened 2024-01-01
Chkfile initial guess problem in ROHF and ROKS (issue #1986)

sunqm:


fishjojo (approved):

pyscf/scf/hf.py
```diff
@@ -656,12 +656,28 @@ def init_guess_by_chkfile(mol, chkfile_name, project=None):
...
+            the same molecule with different ordering of atoms.
+
     Returns:
         Density matrix, 2D ndarray
     '''
     from pyscf.scf import uhf
     dm = uhf.init_guess_by_chkfile(mol, chkfile_name, project)
-    return dm[0] + dm[1]
+    mo_coeff = dm.mo_coeff[0]
+    mo_occ = dm.mo_occ[0] + dm.mo_occ[1]
```
susilehtola, line 679: Wouldn't e.g. UHF natural orbitals be a better way to initialize `mo_coeffs`?

pyscf/scf/rohf.py
```diff
@@ -49,7 +49,28 @@ def init_guess_by_atom(mol):
...
+
+            If project is set to None, the projection is only applied when the
+            basis sets of the chkfile's molecule are different to the basis
+            sets of the given molecule (regardless whether the geometry of
+            the two molecules are different).  Note the basis sets are
+            considered to be different if the two molecules are derived from
+            the same molecule with different ordering of atoms.
+    '''
+    dm = uhf.init_guess_by_chkfile(mol, chkfile_name, project)
+    mo_coeff = dm.mo_coeff[0]
```
susilehtola, line 71: same here

pyscf/scf/hf.py
sunqm: Updated to natural orbitals here

pyscf/scf/rohf.py
sunqm: Natural orbitals may be better. But I'm not sure how to properly assign the occupancy of spin-up and spin-down orbitals. I still use UHF spin-up orbitals here.

Issue #1886, opened 2023-10-02
cistring should use int64 everywhere


kevinsung:
Partly a question. Why does `fci.cistring` use `int32` in some places instead of `int64` everywhere? For example, `make_strings` returns a list of `int64`, but `gen_occslst` and `gen_linkstr_index` return arrays of `int32`. I think using `int64` everywhere would be more consistent and support more use cases.

sunqm:
Strictly speaking, strings should be a bit-wise data type, than `int64` type. Each bit in the bit-wise data represents one orbital, and 0/1 means unoccupied/occupied. Creating a custom bit-wise type for numpy arrays makes the code more complicated, thus we just use the `int64` to represent the bit-wise data type.  `int32` is used for the indices of orbitals only. It should not be used to represent any strings.

kevinsung:
Using `int32` for the indices causes issues when the range of indices exceeds the representable range of `int32`. For example, the following program crashes with `floating point exception (core dumped)`:
```python
from pyscf.fci import cistring

norb = 48
nelec = 24

string = 0b101100101000111111101011001001010001000100011011
address = cistring.str2addr(norb=norb, nelec=nelec, string=string)
string2 = cistring.addr2str(norb=norb, nelec=nelec, addr=address)
```
This is a bug that can be resolved by using `int64` instead of `int32`.

sunqm:
It has been fixed in the latest updates

verena-neufeld:
Hi @kevinsung , thanks for the comment! It seems that it has been resolved now and the issue can be closed?

kevinsung:
Yes, it looks like there was one change that was missed. I opened https://github.com/pyscf/pyscf/pull/2014, which closes this issue.

kevinsung:
Actually, I'm not sure #2014 is correct, so I closed it. The reason I opened it is that it seems strange that the example I provided above,
```python
from pyscf.fci import cistring

string = 0b101101111101101111001110111111110111111100
address = cistring.str2addr(norb=63, nelec=32, string=string)
string2 = cistring.addr2str(norb=63, nelec=32, addr=address)
assert string == string2
```
passes the call to `str2addr` but fails the call to `addr2str`:
```
Traceback (most recent call last):
  File "/home/ksung/projects/pyscf/scratch/scratch.py", line 5, in <module>
    string2 = cistring.addr2str(norb=63, nelec=32, addr=address)
[...]

PR #2014, opened 2024-01-04
make str2addr fail on large address

kevinsung:
Closes https://github.com/pyscf/pyscf/issues/1886.

Looks like this change was missed in https://github.com/pyscf/pyscf/pull/1971.


kevinsung:
Actually, it looks like this might be intentional.

PR #1893, opened 2023-10-06
Support using DF-CCSD and frozen for FNO-CCSD

hongzhouye:
Support using 1. DF-CCSD and 2. frozen for FNO-CCSD

PR #2021, opened 2024-01-11
fixing frozen for FNO

hongzhouye:
In a previous PR (#1893), the `frozen` variable in FNOCCSD was not used in the full-system MP2 calculation. As a result,
1. natural orbitals are selected based on non-frozen MP2 rdm1, leading to suboptimal active-space NOs
2. MP2 composite correction includes the core correlation energy, making the final FNOCCSD energy include a MP2-level core correlation and thus not approximate frozen-core CCSD.

Both problems are fixed in this PR. The `make_fno` function in MP2/UMP2 now supports general `frozen` pattern. The unit tests are updated for frozen-core tests. New tests for a UHF reference are added.

@sunqm The reason why previous tests not passing is likely because the reference values were generated using all-e potential while the tests themselves used a pseudopotential (likely a copy-paste error when I wrote the test file). Once I turned off the pseudopotential, all tests that do not use `frozen` (thus not affecting by this PR) can pass.

### DIFF
--- a/pyscf/cc/__init__.py
+++ b/pyscf/cc/__init__.py
@@ -195,17 +195,11 @@ def FNOCCSD(mf, thresh=1e-6, pct_occ=None, nvir_act=None, frozen=None):
             Number of virtual NOs to keep. Default is None. If present, overrides `thresh` and `pct_occ`.
     """
     import numpy
-    if frozen is None:
-        frozenocc = 0
-    else:
-        assert(isinstance(frozen, (int,numpy.int64)))
-        frozenocc = frozen
     from pyscf import mp
-    pt = mp.MP2(mf).set(verbose=0).run()
+    pt = mp.MP2(mf, frozen=frozen).set(verbose=2).run()
     frozen, no_coeff = pt.make_fno(thresh=thresh, pct_occ=pct_occ, nvir_act=nvir_act)
-    frozen = numpy.hstack([numpy.arange(frozenocc),frozen])
     if len(frozen) == 0: frozen = None
-    pt_no = mp.MP2(mf, frozen=frozen, mo_coeff=no_coeff).set(verbose=0).run()
+    pt_no = mp.MP2(mf, frozen=frozen, mo_coeff=no_coeff).set(verbose=2).run()
     mycc = CCSD(mf, frozen=frozen, mo_coeff=no_coeff)
     mycc.delta_emp2 = pt.e_corr - pt_no.e_corr
     from pyscf.lib import logger
--- a/pyscf/mp/mp2.py
+++ b/pyscf/mp/mp2.py
@@ -229,13 +229,18 @@ def make_fno(mp, thresh=1e-6, pct_occ=None, nvir_act=None, t2=None):
     n,v = numpy.linalg.eigh(dm[nocc:,nocc:])
     idx = numpy.argsort(n)[::-1]
     n,v = n[idx], v[:,idx]
+    logger.debug1(mp, 'make_fno: noon = %s', n)
 
     if nvir_act is None:
         if pct_occ is None:
             nvir_act = numpy.count_nonzero(n>thresh)
         else:
-            print(numpy.cumsum(n/numpy.sum(n)))
-            nvir_act = numpy.count_nonzero(numpy.cumsum(n/numpy.sum(n))<pct_occ)
+            pct_occ_sum = numpy.cumsum(n/numpy.sum(n))
+            logger.debug1(mp, 'make_fno: pctsum(noon) = %s', pct_occ_sum)
+            nvir_act = numpy.count_nonzero(pct_occ_sum<pct_occ)
+
+    if nvir_act == 0:
+        logger.warn(mp, 'make_fno: nvir_act = 0')
 
     fvv = numpy.diag(mf.mo_energy[nocc:])
     fvv_no = numpy.dot(v.T, numpy.dot(fvv, v))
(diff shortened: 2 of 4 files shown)


pyscf/cc/__init__.py
```diff
@@ -183,29 +183,27 @@ def RQCISD(mf, frozen=None, mo_coeff=None, mo_occ=None):
...
     from pyscf import mp
-    pt = mp.MP2(mf).set(verbose=0).run()
-    frozen, no_coeff = pt.make_fno(thresh=thresh, pct_occ=pct_occ, nvir_act=nvir_act)
-    frozen = numpy.hstack([numpy.arange(frozenocc),frozen])
+    pt = mp.MP2(mf, frozen=frozen).set(verbose=2).run()
+    frozen, no_coeff = pt.make_fno(thresh=thresh, pct_occ=pct_occ, pvir_act=pvir_act,
+                                   nvir_act=nvir_act)
     if len(frozen) == 0: frozen = None
-    pt_no = mp.MP2(mf, frozen=frozen, mo_coeff=no_coeff).set(verbose=0).run()
+    pt_no = mp.MP2(mf, frozen=frozen, mo_coeff=no_coeff).set(verbose=2).run()
```
verena-neufeld, line 206: Should verbosity level be taken from mol/cell instance?

verena-neufeld: (I haven't really read through enough to review beyond this high-level comment)

pyscf/cc/__init__.py
```diff
@@ -183,29 +183,27 @@ def RQCISD(mf, frozen=None, mo_coeff=None, mo_occ=None):
...
+def FNOCCSD(mf, thresh=1e-6, pct_occ=None, pvir_act=None, nvir_act=None, frozen=None):
     """Frozen natural orbital CCSD
 
     Attributes:
         thresh : float
             Threshold on NO occupation numbers. Default is 1e-6 (very conservative).
         pct_occ : float
             Percentage of total occupation number. Default is None. If present, overrides `thresh`.
+        pvir_act : float
+            Percentage of total number of virtuals (ceiling is taken to get an integer number).
```
verena-neufeld, line 195: Is this "Percentage of number of virtual NOs to keep" or something else? (I'm comparing this to the comment below "nvir_act").

pyscf/cc/__init__.py
```diff
@@ -183,29 +183,27 @@ def RQCISD(mf, frozen=None, mo_coeff=None, mo_occ=None):
...
+            Default is None. If present, overrides `thresh` and `pct_occ`.
         nvir_act : int
             Number of virtual NOs to keep. Default is None. If present, overrides `thresh` and `pct_occ`.
```
verena-neufeld, line 198: Does `nvir_act` override `pvir_act`, too? It starts getting a little bit tricky to distinguish the keywords here and what takes priority. Some sort of overview sentence comparing them might be helpful.

pyscf/cc/__init__.py
verena-neufeld: The "Percentage of total occupation number" is called `pct_occ`. `pvir_act` doesn't use that naming convention (no "pct") but follows `nvir_act` (which makes sense but might be a little bit inconsistent) <- minor comment.

pyscf/cc/__init__.py
```diff
@@ -183,29 +183,27 @@ def RQCISD(mf, frozen=None, mo_coeff=None, mo_occ=None):
...
     import numpy
-    if frozen is None:
-        frozenocc = 0
-    else:
-        assert(isinstance(frozen, (int,numpy.int64)))
-        frozenocc = frozen
     from pyscf import mp
-    pt = mp.MP2(mf).set(verbose=0).run()
-    frozen, no_coeff = pt.make_fno(thresh=thresh, pct_occ=pct_occ, nvir_act=nvir_act)
-    frozen = numpy.hstack([numpy.arange(frozenocc),frozen])
```
verena-neufeld, line 206: Should there be a comment in the docstring how `frozen` passed by the user is being used?

pyscf/mp/mp2.py
```diff
@@ -203,7 +203,7 @@ def _gamma1_intermediates(mp, t2=None, eris=None):
     return -dm1occ, dm1vir
 
 
-def make_fno(mp, thresh=1e-6, pct_occ=None, nvir_act=None, t2=None):
+def make_fno(mp, thresh=1e-6, pct_occ=None, pvir_act=None, nvir_act=None, t2=None):
     r'''
```
verena-neufeld, line 207: See previous review comments

pyscf/mp/mp2.py
```diff
@@ -212,40 +212,75 @@ def make_fno(mp, thresh=1e-6, pct_occ=None, nvir_act=None, t2=None):
...
 
     nmo = mp.nmo
     nocc = mp.nocc
+    frozen_mask = mp.get_frozen_mask()
+    dm = dm[numpy.ix_(frozen_mask,frozen_mask)]
     n,v = numpy.linalg.eigh(dm[nocc:,nocc:])
     idx = numpy.argsort(n)[::-1]
     n,v = n[idx], v[:,idx]
+    pct_occ_sum = numpy.cumsum(n/numpy.sum(n))
+    numpy.set_printoptions(threshold=nmo)
```
verena-neufeld, line 242: Should this be added as an option to `log` since this is about how the logging should be done and might apply to other parts of pyscf? Or is this a special case?

verena-neufeld (commented):
I have only added some high level comments. I have not tested this or more carefully read through.

hongzhouye:
@verena-neufeld Thanks for the comments! There are things in the MP2 module that I need to fix first (mostly handle user input mean-field through `mo_coeff` and `mo_occ`, but this affects the iterative solver and other places as well). I will let you know once it's ready for review.

PR #2039, opened 2024-01-25
add 6-31g basis set for 31-36 elements

wxj6000:
checked 6-31g only. Need to check if other basis set is incomplete.


susilehtola:
Note that PySCF already has an interface to the Basis Set Exchange. As long as `basis_set_exchange` is installed, basis sets don't need to be included in PySCF at all.

wxj6000:
> Note that PySCF already has an interface to the Basis Set Exchange. As long as `basis_set_exchange` is installed, basis sets don't need to be included in PySCF at all.

Thanks. I didn't know the package `basis_set_exchange` before. Nice work! Unfortunately it only supports python 3.6 - python 3.9.

sunqm:
basis-set-exchange is a noarch library. It supports any Python 3 versions. If the basis data added in this PR are available in basis set exchange database, I think it's not necessary to add them in the pyscf code base. We can close this PR.

wxj6000:
> basis-set-exchange is a noarch library. It supports any Python 3 versions. If the basis data added in this PR are available in basis set exchange database, I think it's not necessary to add them in the pyscf code base. We can close this PR.

You are right. I was mislead by PyPi webpage. The basis set is available in BSE. Should `basis_set_exchange` package be moved to 'install_requires' from 'extras_require'?

verena-neufeld:
Thanks for the pull request @wxj6000 even though it will probably be closed. I don't think `basis_set_exchange` should be moved to required installs as pyscf can run without that package, too, in many cases.

wxj6000:
Thank you for your opinions. I am closing the PR.

Issue #2043, opened 2024-01-30
pyscf.pbc.gto.parse bug


ZihangL:
Dear,
I encounter a small bug of `pyscf.pbc.gto.parse` when I am using pyscf2.4. In pyscf2.3, the code runs correctly.

My code
```python
import numpy as np
from pyscf.pbc import gto
cell = gto.Cell()
cell.unit = 'B'
cell.atom = '''H     4.1965594  4.227266   4.30915   
               H     1.2282612  2.5588777  4.30915   
               H     0.51177555 1.2794389  0.51177555
               H     2.5588777  0.         0.        '''
cell.spin = 0
cell.basis = {'H':gto.parse('''#BASIS SET
                               H SZV-GTH
                               1
                               1  0  0  4  1
                                       8.3744350009  -0.0283380461
                                       1.8058681460  -0.1333810052
                                       0.4852528328  -0.3995676063
                                       0.1658236932  -0.5531027541
                               #''')}
cell.a = np.eye(3) * 5.117755447140034
cell.build()
print(cell)
```

Run in pyscf 2.4
```bash
Traceback (most recent call last):
  File "/home/lizh/hydrogen/hydrogen/main/src/temp/pyscf24bug.py", line 10, in <module>
    cell.basis = {'H':gto.parse('''#BASIS SET
  File "/home/lizh/.local/lib/python3.9/site-packages/pyscf/gto/basis/__init__.py", line 496, in parse
    return parse_cp2k.parse(string, symb, optimize)
TypeError: parse() takes from 1 to 2 positional arguments but 3 were given
```

Run in pyscf 2.3
```bash
<pyscf.pbc.gto.cell.Cell object at 0x2aaab1fefe50>
```

verena-neufeld:
Thanks @ZihangL for reporting this bug! I haven't closely looked, but it does seem that in https://github.com/pyscf/pyscf/blob/35835bcb950b525aa89df6d8eb6e0f0366627549/pyscf/gto/basis/__init__.py#L494 , the line should be `return parse_cp2k_pp.parse(string)` and in https://github.com/pyscf/pyscf/blob/35835bcb950b525aa89df6d8eb6e0f0366627549/pyscf/gto/basis/__init__.py#L496 , the line should be `return parse_cp2k.parse(string, optimize)`.

ZihangL:
Thanks for your reply!

verena-neufeld:
I think this has been fixed now thanks to @fishjojo PR

PR #2045, opened 2024-01-30
fix cp2k basis parse (issue #2043)

fishjojo:
TODO: allow choosing symbols when parsing GTH PPs.

### DIFF
--- a/pyscf/gto/basis/parse_cp2k_pp.py
+++ b/pyscf/gto/basis/parse_cp2k_pp.py
@@ -24,7 +24,7 @@ import sys
 from pyscf.lib.exceptions import BasisNotFoundError
 import numpy as np
 
-def parse(string):
+def parse(string, symb=None):
     '''Parse the pseudo text *string* which is in CP2K format, return an internal
     basis format which can be assigned to :attr:`Cell.pseudo`
     Lines started with # are ignored.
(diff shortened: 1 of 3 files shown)


verena-neufeld:
Thanks @fishjojo . Since tests didn't pick this bug up, do you think we need new tests for this?

pyscf/gto/basis/parse_cp2k_pp.py
```diff
@@ -24,7 +24,7 @@
 from pyscf.lib.exceptions import BasisNotFoundError
 import numpy as np
 
-def parse(string):
+def parse(string, symb=None):
```
verena-neufeld, line 27: I don't think the `symb` keyword is used in that function. Is that there for consistency? Should individual element's PP be parsed?

verena-neufeld:
Thanks so much @fishjojo ! I think it looks great. I've only commented with one more small question (sorry, I should have done that yesterday with the other question).

pyscf/gto/basis/parse_cp2k_pp.py
sunqm: I guess this is just to make parse_cp2k_pp function signatures the same to that in parse_nwchem.

sunqm (approved):

pyscf/gto/basis/parse_cp2k_pp.py
verena-neufeld: Ok, thanks!

fishjojo: I can also add symbol choosing for PPs, maybe later this week.

PR #2046, opened 2024-01-30
SA-CASSCF Nonadiabatic Coupling Vectors

matthew-hennefarth:
Migration of the implementation of SA-CASSCF NACs from [mrh](https://github.com/MatthewRHermes/mrh). This code was written by @MatthewRHermes with some help from Paul Calio. I had added the tests which compare this implementation to that of OpenMolcas (v23.02). 

Briefly,
```python
from pyscf import gto, scf, mcscf
mol = gto.M(atom="H 0 0 0; H 1.3 0 0", basis="sto-3g", verbose=0, output="/dev/null")
mc = mcscf.CASSCF(scf.RHF(mol).run(), 2, 2).state_average([0.5, 0.5])
mc.fix_spin_(ss=0, shift=1)
mc.kernel()

# Similar to gradient method but requires a tuple of states
mc_nac = mc.nac_method()
de = mc_nac.kernel(state=(0,1))
print(de)

de_with_etf = mc_nac.kernel(state=(0,1), use_etfs=True)
print(de_with_etf)
```
Output:
```
[[-0.22461197  0.          0.        ]
 [-0.22461197  0.          0.        ]]
[[-1.76512088e-16  0.00000000e+00  0.00000000e+00]
 [ 1.49547359e-16  0.00000000e+00  0.00000000e+00]]
```

This code returns the NAC between state 0 and 1 (hence the `state=(0,1)`) of the calculation as 

$$
d_{01} = \langle\Psi_0 |\nabla_{R}\Psi_1 \rangle
$$

Additionally, the `use_etfs` kwarg is to control whether one adds in the so-called CSF contribution. Like in [OpenMolcas](https://molcas.gitlab.io/OpenMolcas/sphinx/users.guide/programs/alaska.html) once can decide to neglect this contribution since it is thought to break translational and rotational invariance [[1](https://doi.org/10.1021/jz3006173)].

1. Shervin Fatehi, Joseph E. Subotnik. Derivative couplings with built-in electron-translation factors: Application to benzene. J. Phys. Chem. Lett., 3[15] (2012) 2039–2043. [doi:10.1021/jz3006173](https://doi.org/10.1021/jz3006173)


verena-neufeld:
Hi @matthew-hennefarth , thank you so much for pull request and the new contribution! It looks very useful and a great augmentation to pyscf. In an effort to keep pyscf maintainable and since this seems to add a new module, I wonder what the criterion is for a new module to be in the main code or in an extension. What do you think?
Disclaimer: I haven't worked with nonadiabatic coupling vectors and this is a question I had wondered myself, not discussing this with others. @fishjojo @sunqm ?

matthew-hennefarth:
> In an effort to keep pyscf maintainable and since this seems to add a new module, I wonder what the criterion is for a new module to be in the main code or in an extension. What do you think?

I would think that new modules should be put in the main code if it is a well-defined method/property and should put into an extension (like [PySCF-forge](https://github.com/pyscf/pyscf-forge)) if it is more experimental/not ready to be integrated into the main code. SA-CASSCF NACs are well defined/used within the community for a variety of tasks including nonadiabatic molecular dynamics (using [SHARC](https://sharc-md.org/) for example) or finding minimum energy conical intersections. On the other hands, CMS-PDFT derivative couplings should be added to PySCF-forge] (in this [PR](https://github.com/pyscf/pyscf-forge/pull/44)) since MC-PDFT is not currently in the main repo. From what I know/could find quickly, NACs are implemented in the following programs for the following methods:

- OpenMolcas:
  - SA-CASSCF
  - [MS-CASPT2/RASPT2][10.1021/acs.jctc.2c00301]
  - [XMS-CASPT2/RASPT2][10.1021/acs.jctc.2c00301]
  - [XDW-CASPT2/RASPT2][10.1021/acs.jctc.2c00301]
  - [RMS-CASPT2/RASPT2][10.1021/acs.jctc.2c00301]
- BAGEL
  - SA-CASSCF
  - [MS-CASPT2][10.1021/acs.jctc.7b00018]
  - [XMS-CASPT2][10.1021/acs.jctc.7b00018]
- GAMESS
  - MCSCF
- Orca (v5.0)
  - [TD-DFT][10.1002/wcms.1606]
- QChem
  - [TD-DFT](https://manual.q-chem.com/5.3/subsec_NACMEs_and_CIs.html) + spin-flip
  - [CIS](https://manual.q-chem.com/5.3/subsec_NACMEs_and_CIs.html) + spin-flip
- Turbomole
  - [TD-DFT](https://pubs.acs.org/doi/10.1021/acs.jctc.3c00347)

Of course, adding SA-CASSCF NACs to the main repo is up to the discretion of the PySCF maintainers/owners.

[10.1021/acs.jctc.2c00301]: https://pubs.acs.org/doi/10.1021/acs.jctc.2c00301
[10.1021/acs.jctc.7b00018]: https://pubs.acs.org/doi/full/10.1021/acs.jctc.7b00018
[10.1002/wcms.1606]: https://wires.onlinelibrary.wiley.com/doi/full/10.1002/wcms.1606

pyscf/nac/sacasscf.py
```diff
@@ -0,0 +1,285 @@
...
+    use_etfs : logical
+        If True, use the ``electron translation factors'' of Fatehi and
+        Subotnik [JPCL 3, 2039 (2012)], which guarantee conservation of
+        total electron + nuclear momentum when the nuclei are moving
+        (i.e., in non-adiabatic molecular dynamics). This corresponds
+        to omitting the so-called ``CSF contribution'' [cf. JCTC 12,
+        3636 (2016)].
+    '''
+
+    def __init__(self, mc, state=None, mult_ediff=False, use_etfs=False):
```
sunqm, line 125: `assert len(state) == 2`

sunqm (approved):

pyscf/nac/sacasscf.py
matthew-hennefarth: if `state` is just an integer, the code is currently set-up to evaluate 

$$
  \langle\psi_I|\nabla_{R}\psi_I\rangle 
$$

which, since the derivative couplings are anti-symmetric with respect to index swapping

$$
\langle\psi_I|\nabla_{R}\psi_J\rangle = - \langle\psi_J|\nabla_{R}\psi_I\rangle
$$

means that we should just return 0. Currently it will return `+/-inf` since division by 0.

> There may need a check to ensure the specified states within the SACASSCF states

Currently, an `IndexError` will occur in `self.make_fcasscf_nacs(state)`. 

I think based off of this, adding the `assert len(state) == 2` as well as a check to make sure that `state[0]` and `state[1]` are some valid, distinct roots should address these problems.

matthew-hennefarth: Okay, added the following: 

- `assert len(state) == 2` in `__init__ and `kernel` (as well as whenever `_unpack_state` is called.
- when `state[0]==state[1]`, return the correctly shaped 0 object without doing any actual calculations.

PR #2053, opened 2024-02-01
DIIS with damping

hongzhouye:
## Highlight
Many of us have found SCF convergence issues for metallic systems with DIIS in PySCF even when smearing is used. By contrast, most other codes (QE, VASP, FHI-aims) do not have such a problem. The main reason is found to be the use of damping on top of DIIS in other codes. See e.g.,
- Eqn 92 in [10.1016/0927-0256(96)00008-0](https://doi.org/10.1016/0927-0256(96)00008-0) for VASP
- Eqn 43 in [10.1016/j.cpc.2009.06.022](https://doi.org/10.1016/j.cpc.2009.06.022) for FHI-aims

This PR modifies the current DIIS implementation in PySCF to enable using damping by simply
```
mf.diis_damp = 0.9
```
The SCF convergence for metallic systems is significantly improved as demonstrated in the plots below for two representative systems: ethylene carbonate adsorption on lithium and CO adsorption on copper. In both cases, the regular DIIS (black) struggles with converging SCF, while simple linear damping (blue) shows monotonic convergence at a very slow rate. By contrast, DIIS with damping (green) converges as fast as in many gapped systems.

scf_conv_compare

Note: the `damping` function is also updated to be the simple linear damping, i.e.,
```
F = Fnew * (1-damp) + Fold * damp
```

## Math

For a fixed point problem
```
f(x) = x
```
The regular DIIS as currently employed in PySCF mixes `h` previous output vectors
```
x_{m+1} = \sum_{i=m-h}^{m} c_i f(x_i)
```
while the damped DIIS also mixes the input vectors with a factor `damp`
```
x_{m+1} = \sum_{i=m-h}^{m} c_i [ f(x_i) * (1 - damp) + x_i * damp ]
```
---
Many thanks to @tberkel, @welltemperedpaprika, and @minyez for helpful discussion and @sohangkundu and @zhutianyu1991 for suggestions on test systems.

### DIFF
--- a/pyscf/pbc/scf/khf.py
+++ b/pyscf/pbc/scf/khf.py
@@ -125,7 +125,8 @@ def get_jk(mf, cell, dm_kpts, kpts, kpts_band=None, with_j=True, with_k=True,
                                  omega, exxdiv=mf.exxdiv)
 
 def get_fock(mf, h1e=None, s1e=None, vhf=None, dm=None, cycle=-1, diis=None,
-             diis_start_cycle=None, level_shift_factor=None, damp_factor=None):
+             diis_start_cycle=None, level_shift_factor=None, damp_factor=None,
+             fock_last=None):
     h1e_kpts, s_kpts, vhf_kpts, dm_kpts = h1e, s1e, vhf, dm
     if h1e_kpts is None: h1e_kpts = mf.get_hcore()
     if vhf_kpts is None: vhf_kpts = mf.get_veff(mf.cell, dm_kpts)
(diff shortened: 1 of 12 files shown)


verena-neufeld:
Hi @hongzhouye , thanks so much! For a coupled cluster case for example I know that here is `iterative damping` https://github.com/pyscf/pyscf/blob/0e0345b751a44023c56dcdaa4f4b69075a547443/pyscf/cc/ccsd.py#L77  On the very first glance it seems a little bit related? If not, your addition can also be used for cc's DIIS which would probably be helpful there, too! If it is related, maybe we could standardize how damping is called.

hongzhouye:
@verena-neufeld after reading the CC code I think it's the same as what I implemented here for SCF. Standardizing this function sounds a good idea, maybe in another PR.

sunqm (approved):

Issue #2072, opened 2024-02-07
add GPU related info to output file


wxj6000:
We will need to dump the following information to the output file  
- GPU4PySCF version
- CuPy version
- GPU type
- CUDA driver version, etc

The simplest implementation is inserting the corresponding script after
https://github.com/pyscf/pyscf/blob/master/pyscf/gto/mole.py#L2722

Any suggestions?

wxj6000:
Yes, we can close this issue now.

PR #2077, opened 2024-02-11
Add an API to dump system info

sunqm:
The `lib.misc.format_sys_info` function can be overwritten by external libraries, to add additional platform information @wxj6000


fishjojo (approved):

wxj6000:
https://github.com/pyscf/pyscf/issues/2072

Issue #2069, opened 2024-02-06
Unsure How Best To Employ ωB97X-D Functional


corinwagen:
Hey folks, I'm a bit perplexed about how to use ωΒ97X-D with PySCF. 

To select the functional, you can say [``xc=WB97X_D``](https://github.com/pyscf/pyscf/blob/master/pyscf/dft/libxc.py#L522) or [``xc=WB97X-D``](https://github.com/pyscf/pyscf/blob/master/pyscf/dft/libxc.py#L1251) - but the external dftd3 library doesn't parse either of these names. Instead, dftd3 seems to expect ``xc`` to be called [``wb97x``](https://github.com/dftd3/simple-dftd3/blob/main/src/dftd3/param.f90#L202), which [maps to a different functional in PySCF](https://github.com/pyscf/pyscf/blob/master/pyscf/dft/libxc.py#L521). 

Part of the issue here is that calling ``dftd3.pyscf.energy()`` [automatically builds the xc name from the PySCF object](https://github.com/dftd3/simple-dftd3/blob/main/python/dftd3/pyscf.py#L305C5-L311C6), with no way to override.

Here's an example script that dies:
```
import pyscf
import dftd3.pyscf as disp

xyz = """C                 -0.71618036    2.71883299    0.00000000
C                  0.67897964    2.71883299    0.00000000
C                  1.37651764    3.92658399    0.00000000
C                  0.67886364    5.13509299   -0.00119900
C                 -0.71596136    5.13501499   -0.00167800
C                 -1.41356236    3.92680899   -0.00068200
H                 -1.26593936    1.76651599    0.00045000
H                  1.22848764    1.76631999    0.00131500
H                  2.47619764    3.92666399    0.00063400
H                  1.22906364    6.08723599   -0.00125800
H                 -1.26608336    6.08729599   -0.00263100
H                 -2.51316636    3.92699199   -0.00086200"""

mol = pyscf.M(
    atom=xyz,
    basis="6-31G(d)",
    charge=0,
    spin=0,
    symmetry=True
)

mf = pyscf.scf.KS(mol)
mf.xc = "WB97X_D"

mf = disp.energy(mf)

energy = mf.kernel()
```

And here's the error message that I get:

```
Traceback (most recent call last):
  File "/Users/cwagen/business_code/pyscf/wb97xd.py", line 30, in <module>
    energy = mf.kernel()
             ^^^^^^^^^^^
  File "<string>", line 2, in kernel
[...]

PR #2084, opened 2024-02-17
DFT wrapper supporting conventional DFT methods

wxj6000:
This PR
- supports some conventional DFT methods combining XC and dispersion correction/nonlocal correction, e.g wb97m-d3bj = wb97m-v + d3bj, b3lyp-d3bj = b3lyp + d3bj
- is backward compatible. No API changes for users.
- works through `mf=rks.RKS(mol, xc='wb97m-d3bj')` only, not `mf.xc=wb97m-d3bj`. `mf.xc` is defined as the same as before. `mf.xc` and `mf.disp` are used for customizing DFT methods.
- replaces `ni.libxc.is_nlc` with `ni.is_nlc` such that `ni.is_nlc` can be customized.


wxj6000:
#2069

pyscf/dft/rks.py
```diff
@@ -260,6 +260,25 @@ def define_xc_(ks, description, xctype='LDA', hyb=0, rsh=(0,0,0)):
...
+    xc = dft_method
+    disp = None
+
+    # special cases
+    if method_lower == 'wb97m-d3bj':
+        return 'wb97m-v', False, 'd3bj'
+    if method_lower == 'b97m-d3bj':
+        return 'b97m-v', False, 'd3bj'
+    if method_lower == 'wb97x-d3bj':
+        return 'wb97x-v', False, 'd3bj'
```
hebrewsnabla, line 275: wb97x-d3 is also a special case.

hebrewsnabla:
> works through mf=rks.RKS(mol, xc='wb97m-d3bj') only, not mf.xc=wb97m-d3bj.

could you add an example for those functionals?

pyscf/dft/rks.py
```diff
@@ -260,6 +260,46 @@ def define_xc_(ks, description, xctype='LDA', hyb=0, rsh=(0,0,0)):
...
+DISP_VERSIONS = ['d3bj', 'd3zero', 'd3bjm', 'd3zerom', 'd3op', 'd4']
+
+def dft_method_parser(dft_method):
+    ''' conventional dft method ->
+    ( internal xc,
+      disable nlc,
+      dispersion version
+      with 3-body dispersion or not
+      '''
+    method_lower = dft_method.lower()
```
sunqm, line 273: Please check if xc is a string and return early. The input xc can be a tuple of strings or other data types.

pyscf/dft/rks.py
```diff
@@ -321,10 +361,13 @@ class KohnShamDFT:
     -76.415443079840458
     '''
 
-    _keys = {'xc', 'nlc', 'grids', 'nlcgrids', 'small_rho_cutoff'}
+    _keys = {'xc', 'nlc', 'grids', 'disp', 'nlcgrids', 'small_rho_cutoff'}
```
sunqm, line 364: Add the key 'disp_with_3body'

pyscf/dft/rks.py
```diff
@@ -339,6 +382,11 @@ def __init__(self, xc='LDA,VWN'):
 # don't modify the following attributes, they are not input options
         self._numint = numint.NumInt()
 
+        # disable nlc for certain xc such as wb97m-d3bj
+        if with_nlc is not None:
```
sunqm, line 386: if one sets `nlc`, or `xc` after creating RKS object, the `_numint.nlc_coeff` and `self._numint.is_nlc` can never parse the xc correctly. Are there any other ways to control this?

sunqm:
> > works through mf=rks.RKS(mol, xc='wb97m-d3bj') only, not mf.xc=wb97m-d3bj.
> 
> could you add an example for those functionals?

This is my suggestion to work so.

The additional parser in this PR can map these functionals to certain special combinations of other functionals when initializing DFT. If you manually set the `xc` attribute for a DFT object, that sounds you force the DFT to use the  specified functional. I think warnings can be issued in XC parser when these functionals are detected, and suggestions can be put in the warning message. But the DFT calculation should still use the specified xc functional.

pyscf/dft/rks.py
wxj6000: `nlc` is possibly disabled only when creating RSK object with certain functionals such as wb97m-d3bj. After the creation, setting nlc or xc does not change `self._numint.is_nlc`. The functionality of setting nlc or xc after creating RKS object is the same as before.

And one still can use `self._numint.libxc.is_nlc` and `self._numint.libxc.nlc_coeff` to parse xc code if `self._numint.is_nlc` is customized.

wxj6000: Yes, it is better to customize the parser in those two functions.
https://github.com/wxj6000/pyscf/blob/dft_wrapper/pyscf/dft/numint.py#L2647
and 
https://github.com/wxj6000/pyscf/blob/dft_wrapper/pyscf/dft/numint.py#L2644
I will work on it later.

pyscf/dft/dft_parser.py
```diff
@@ -0,0 +1,76 @@
...
+        return 'wb97x-v', False, ('wb97x', 'd3bj', False)
+
+    # J. Chem. Theory Comput. 2013, 9, 1, 263–272
+    if method_lower in ['wb97x-d3']:
+        raise NotImplementedError('wb97x-d3 is not supported yet.')
+
+    if method_lower.endswith('-3c'):
+        raise NotImplementedError('*-3c methods are not supported yet.')
+
+    for d in DISP_VERSIONS:
```
sunqm, line 63: May need to consider a corner case like `wb97m-d3bjm`

sunqm:
> > works through mf=rks.RKS(mol, xc='wb97m-d3bj') only, not mf.xc=wb97m-d3bj.
> 
> could you add an example for those functionals?

Should work in both cases with the latest updates.

pyscf/dft/dft_parser.py
wxj6000: `wb97m-d3bjm` is not a well-defined xc functional. In this parser, `wb97m-d3bjm` will be interpreted as `wb97m` and `d3bjm`. `wb97m` is not a well-defined xc functional. I am assuming xc parser or libxc will raise an error.

PR #2085, opened 2024-02-18
Add vDZP Basis Set

corinwagen:
The vDZP basis set is key to the wB97X-3c method from Muller and Grimme, published [last year](https://pubs.aip.org/aip/jcp/article-abstract/158/1/014103/2867476/B97X-3c-A-composite-range-separated-hybrid-DFT?redirectedFrom=fulltext), which is the best-performing of the 3c methods. While the other 3c methods depend on Grimme's gcp library, wB97X-3c is just wB97M-D4 with vDZP, a special molecule-optimized basis set that aggressively uses ECPs to cut costs:

> We show that the vDZP basis exhibits qualitative advantages over usual double-ζ basis sets as it is virtually free of BSSE and describes typical electron densities very similar to TZ or even large QZ basis sets. 

It seems like this might be of interest to PySCF users - I grabbed the files from MolSSI-BSE/basis_set_exchange#266, which hasn't seen activity for over a year, converted them into NWChem format, and added them to the internal basis set registry. 

I've included a demo script here in the root of the repo, which can easily be removed for the actual merge. This is my first contribution so I'm not sure if I've done anything incorrectly here; this is a very minor change but hopefully one with some utility for end users.

Notes:
- Setting vDZP as the basis set for a whole molecule results in messages like "ECP vDZP not found for  H" - every atom past Be has an ECP, but H–Be do not. I'm not sure if there's some way to suppress these messages, cosmetically. 
- The discussion in #2069 will also affect wB97X-3c, if this method is implemented in the future.


pyscf/gto/basis/vdzp.dat
```diff
@@ -0,0 +1,3442 @@
+# Generated via Basis Set Exchange CLI on Feb 17 2024
```
wxj6000, line 1: It seems that the basis set will be included in BSE.

susilehtola:
Like I wrote in https://github.com/MolSSI-BSE/basis_set_exchange/issues/266, there is no D-type projector on F, but there is an F-type projector. It is more than likely that there is an error in the basis set; also, many codes assume that projectors are reported in increasing angular momentum and a missing channel in-between can cause problems....

pyscf/gto/basis/vdzp.dat
```diff
@@ -0,0 +1,3442 @@
...
+2       8.16479800            -3.31212400
+F nelec 2
+F ul
+2       1.000000               0.000000
+F S
+2      22.350400             102.597952
+2      11.175200              19.049663
+F P
+2      26.476800             -15.143960
+2      13.238400               2.802921
```
susilehtola, line 2276: This will likely yield the wrong result

marcelmbn:
First, thank you very much for your joint efforts to implement the vDZP basis set into program packages beyond ORCA and TURBOMOLE!

Regarding the issue mentioned above:
> Like I wrote in [MolSSI-BSE/basis_set_exchange#266](https://github.com/MolSSI-BSE/basis_set_exchange/issues/266), there is no D-type projector on F, but there is an F-type projector. It is more than likely that there is an error in the basis set; also, many codes assume that projectors are reported in increasing angular momentum and a missing channel in-between can cause problems....

@corinwagen @susilehtola 
There are two ways to solve this problem (at least as far as I can see):
1. Add a "dummy" `d`-type projector with coefficient 0.0, so that the ascending order of projectors up to `f` can be retained.
2. Just remove the `f` projector as it has a 0.0 coefficient. IMHO, this would be the correct option, as a projector with a coefficient of 0.0 is only a technical remnant from the development process.

I quickly tested both approaches with both TURBOMOLEv7.7.1 and ORCAv5.0.4 and in all cases, the same single-point energy and properties resulted for a simple F_2 molecule.
Please let me know what you think about the proposed solution!

pyscf/gto/basis/vdzp.dat
corinwagen: TBD - unsure what the timeline on that will be, see https://github.com/MolSSI-BSE/basis_set_exchange/issues/266. I used the ``bse convert-basis`` command-line utility ([ref](https://molssi-bse.github.io/basis_set_exchange/conversion.html)) to convert the Turbomole-formatted basis set into NWChem format, which is why this note is there.

corinwagen:
Going to defer to the likes of @susilehtola @wxj6000 @sunqm on how to proceed here - I think it would be great to have a wB97X-3c implementation in PySCF, perhaps building off of the nice recent work on #2084, but unsure what the heuristic is for having basis sets within the repo vs in the basis set exchange, or what the typical timeline for incorporation into the BSE repo is. Happy to do any grunt work necessary to make this happen.

susilehtola:
Once there is a suitable basis set implemented in a working pull request in the BSE, upon merge it will appear in the master branch. The web interface at http://www.basissetexchange.org only gets updated when a new stable release is made. I believe @bennybp is working on a next release.

The discussion on the structure and correct implementation of the basis set should be kept at https://github.com/MolSSI-BSE/basis_set_exchange/issues/266 so that all information is there.

sunqm:
> Going to defer to the likes of @susilehtola @wxj6000 @sunqm on how to proceed here - I think it would be great to have a wB97X-3c implementation in PySCF, perhaps building off of the nice recent work on #2084, but unsure what the heuristic is for having basis sets within the repo vs in the basis set exchange, or what the typical timeline for incorporation into the BSE repo is. Happy to do any grunt work necessary to make this happen.

#2084 may only solve the issue of functional parser. Handling basis set needs to overwrite the existing attributes of mol object. It is against the modular principle of pyscf. This will be solved in another PR. Anyway, the support to wB97X-3c and other wB97-D3 type calculations are recognized and we are working on it.

I'd like to drop the duplicated basis data in pyscf that are available in the BSE repo. But some basis sets are not completely the same to BSE data. Before cleaning up the basis data, it requires some checks throughout these data in pyscf. I'll open a new issue for this problem.

pyscf/gto/basis/vdzp.dat
marcelmbn: @corinwagen: Thanks to @susilehtola and @bennybp, the full vDZP basis set for ωB97X-3c (https://pubs.aip.org/aip/jcp/article/158/1/014103/2867476/B97X-3c-A-composite-range-separated-hybrid-DFT) has been merged in https://github.com/MolSSI-BSE/basis_set_exchange/pull/316 into the basis set exchange data base. This should (in the long term) also resolve this issue.

marcelmbn:
Sorry for the duplicated answers on some issues and pull requests on GitHub, but I want to make sure that this is documented correctly everywhere:
Thanks to @susilehtola and @bennybp, the full vDZP basis set for ωB97X-3c (https://pubs.aip.org/aip/jcp/article/158/1/014103/2867476/B97X-3c-A-composite-range-separated-hybrid-DFT) has been merged in https://github.com/MolSSI-BSE/basis_set_exchange/pull/316 into the basis set exchange data base. This should (in the long term) also resolve this issue.

PR #2098, opened 2024-02-24
Fixed small bugs in scf/addons.py

sunchong137:
Hi there,

Here are the changes I made:

1. Fixed a small bug regarding given mu0 in the `get_occ()` function of the `_SmearingSCF` class. 
2. Still in the `get_occ()` function of the `_SmearingSCF` class: In model systems, the `mol.tot_electrons()` might not be defined, so I added a line to get electron number from `mol.nelec` if the original approach fails.
3. Added an example of running UHF for the Hubbard model.

Please let me know if this works!
Chong

### DIFF
--- a/pyscf/pbc/scf/addons.py
+++ b/pyscf/pbc/scf/addons.py
@@ -121,12 +121,17 @@ class _SmearingKSCF(mol_addons._SmearingSCF):
             if self.mu0 is None:
                 mu_a, occa = mol_addons._smearing_optimize(f_occ, mo_es[0], nocc[0], sigma)
                 mu_b, occb = mol_addons._smearing_optimize(f_occ, mo_es[1], nocc[1], sigma)
-                mu = [mu_a, mu_b]
-                mo_occs = [occa, occb]
             else:
-                mu = self.mu0
-                mo_occs = f_occ(mu[0], mo_es[0], sigma)
-                mo_occs = f_occ(mu[1], mo_es[1], sigma)
+                if numpy.isscalar(self.mu0):
+                    mu_a = mu_b = self.mu0
+                elif len(self.mu0) == 2:
+                    mu_a, mu_b = self.mu0
+                else:
+                    raise TypeError('Unsupported mu0: {self.mu0}')
+                occa = f_occ(mu_a, mo_es[0], sigma)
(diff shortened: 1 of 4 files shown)


fishjojo:
Do you have an example for using `mu0`?

pyscf/pbc/scf/addons.py
```diff
@@ -121,12 +121,17 @@ def get_occ(self, mo_energy_kpts=None, mo_coeff_kpts=None):
...
             else:
-                mu = self.mu0
-                mo_occs = f_occ(mu[0], mo_es[0], sigma)
-                mo_occs = f_occ(mu[1], mo_es[1], sigma)
+                if numpy.isscalar(self.mu0):
+                    mu_a = mu_b = self.mu0
+                elif len(self.mu0) == 2:
+                    mu_a, mu_b = self.mu0
+                else:
+                    raise TypeError('Unsupported mu0: {self.mu0}')
```
sunqm, line 130: `f'Unsupported mu0: {self.mu0}'`

pyscf/scf/addons.py
```diff
@@ -140,12 +140,17 @@ def get_occ(self, mo_energy=None, mo_coeff=None):
...
             else:
-                mu = self.mu0
-                mo_occs = f_occ(mu[0], mo_es[0], sigma)
-                mo_occs = f_occ(mu[1], mo_es[1], sigma)
+                if numpy.isscalar(self.mu0):
+                    mu_a = mu_b = self.mu0
+                elif len(self.mu0) == 2:
+                    mu_a, mu_b = self.mu0
+                else:
+                    raise TypeError('Unsupported mu0: {self.mu0}')
```
sunqm, line 149: `f'Unsupported mu0: {self.mu0}'`

sunqm (approved):

sunchong137:
> Do you have an example for using `mu0`?

Occasions of giving a specified `mu0` are rare. In most cases, we usually require the average electron number to be some value, thus mu is optimized. However, in some thermodynamic calculations, people want to fix mu for comparison or other purposes.

However, if you are talking about a script using `mu0`, I can add one or two lines to the finite temperature example!

sunchong137:
Thank you for the timely feedback and for approving the pull request!

PR #2101, opened 2024-02-25
Add tests for ndarray_pointer_2d function

sunqm:


pyscf/lib/numpy_helper.py
```diff
@@ -1117,15 +1117,11 @@ def expm(a):
     return y
 
 def ndarray_pointer_2d(array):
-    '''Get the C pointer of a 2D array
+    '''Get the memory addresses for each element within the given 2D array
```
fishjojo, line 1120: The function returns the pointer of an array that contains the address of the first element of each row of the input 2d array.

fishjojo (approved):

Issue #2105, opened 2024-02-28
floating point exception in fci calculation over 22 orbitals


xubwa:
Hi all,
I tried a relativistic fci calculation on B2 molecule and encountered a floating point exception thrown by the FCIlinkstr_index function. After digging into the problem, it is due to the improper implementation of binomial coefficients, and for orbital space less than 22, it is avoided due to the hard coded binomial coefficients.
When orbital exceeds 21 or 22, the binomial coefficients is computed by the loop starting from this line https://github.com/pyscf/pyscf/blob/e2cc8c136ed0e57f8597ce3a06e66f97630916ac/pyscf/lib/mcscf/fci_string.c#L135, it is divided by 0 at the first iteration leading to the exception. However, even this divided by 0 is fixed, I'm afraid this loop will not give correct binomial since it impose a round off at every iteration leading the roundoff error to increase.
Here are two scripts that help reproduce the issue, one for nonrelativistic Hamiltonian and one for relativistic Hamiltonian. 
```
from functools import reduce

import numpy

from pyscf import ao2mo
from pyscf import gto
from pyscf import scf
from pyscf import fci

mol = gto.Mole()
mol.verbose = 0
mol.output = None
mol.atom = [
    ['Li', ( 1., 0.    , 1.   )],
    ['Li', (-1., 0.    , 1.   )],
]
mol.basis = 'ccpvtz'
mol.verbose=5
mol.build()
print(mol.nao)
m = scf.RHF(mol).run(conv_tol=1e-14)
ci = fci.FCI(mol, m.mo_coeff)
ci.kernel()
```
```
from functools import reduce

import numpy

from pyscf import ao2mo
from pyscf import gto
from pyscf import scf
from pyscf import fci

mol = gto.Mole()
mol.verbose = 0
mol.output = None
mol.atom = [
    ['B', ( 1., 0.    , 1.   )],
    ['B', (-1., 0.    , 1.   )],
]
mol.basis = '6-31g'
mol.verbose=5
mol.build()

m = scf.X2C(mol).run(conv_tol=1e-14)

norb = m.mo_coeff.shape[1]
h1e = reduce(numpy.dot, (m.mo_coeff.conj().T, m.get_hcore(), m.mo_coeff))
mo_a, mo_b = m.mo_coeff[:mol.nao], m.mo_coeff[mol.nao:]
print('integral transformation')
norb = 24
[...]

PR #2106, opened 2024-02-28
fix binomial function in fci_string

xubwa:
Fix issue #2105

### DIFF
--- a/pyscf/lib/mcscf/fci_string.c
+++ b/pyscf/lib/mcscf/fci_string.c
@@ -130,12 +130,7 @@ static int binomial(int n, int m)
                 if (m*2 <= n) {
                         m = n - m;
                 }
-                uint64_t i;
-                uint64_t val = 1;
-                for (i = m; i <= n; i++) {
-                        val *= i;
-                        val /= i - m;
-                }
+		int val = binomial(n-1,m-1) + binomial(n-1,m);
                 return val;
         }
 }


pyscf/lib/mcscf/fci_string.c
```diff
@@ -130,12 +130,7 @@ static int binomial(int n, int m)
                 if (m*2 <= n) {
                         m = n - m;
                 }
```
matthew-hennefarth, line 132: If you are using the recursion relation below, shouldn't these lines also be removed?

matthew-hennefarth:
Generally speaking, recursion should be avoided as it is typically slow. Though I am not sure if the binomial function will be used in cases much beyond the current cache size due to CPU limitations to perform FCI calculations with such large number of orbitals/electrons, especially given the memoization used here.

To implement this more efficiently people often use a multiplicative version instead of a recursion. Of course, then one has to be careful of not producing over/underflowed values, and it is even more tricky to make sure we are not getting rounding errors due to some intermediate division. I had written an [integer only version in rust](https://github.com/matthew-hennefarth/sci-rs/blob/5b1c1a6ffcb4219fef67b3c62ce47006af589036/src/special/combinatorics.rs#L111-L115) which does not suffer from floating point precision. In c, and using the same input args as given in the binomial signature it might look something like

```c
static int binomial(int n, int m) {
    if (m >= n) {
        return 1;
    } 
    if (n <= 20) {
        return _binomial_cache[n*(n-1)/2+m];
    } 
    int k = n + 1;
    int n_terms = min(m, n-m) + 1; // supposing a min function is defined somewhere...

    int result = 1;
    for (int i = 1; i < n_terms; i++) {
        result = (result * (k-i)) /i;
    }
    return result;
}
```

Not saying you should change the implementation to this version, but just something to keep in mind...

xubwa:
I used a recursion here because in practice, a fci calculation with many electrons and many orbitals is impossible, it either exceeds the cached binomial array up to 21/22 orbitals by a few orbitals or the system only have very few electrons, yielding few recursion layers and can also make full use of the cahced array to get some savings. The code you provide seems able to guarantee the result to be integer at every step, so it also looks good to me. I have no strong preference on this as long as the problem can be fixed.

xubwa:
I used a recursion here because in practice, a fci calculation with many electrons and many orbitals is impossible, it either exceeds the cached binomial array up to 21/22 orbitals by a few orbitals or the system only have very few electrons, yielding few recursion layers and can also make full use of the cahced array to get some savings. The code you provide seems able to gaurantee

matthew-hennefarth:
> I used a recursion here because in practice, a fci calculation with many electrons and many orbitals is impossible, it either exceeds the cached binomial array up to 21/22 orbitals by a few orbitals or the system only have very few electrons, yielding few recursion layers and can also make full use of the cahced array to get some savings.

I agree, this implementation should be fine for PySCF.

sunqm (approved):

PR #2108, opened 2024-02-29
ruff auto-fix for UP027

fchapoton:
this is about
`
UP027 [*] Replace unpacked list comprehension with a generator expression
`
and was scripted using
`
ruff --fix --select=UP027 pyscf/
`

### DIFF
--- a/pyscf/agf2/uagf2.py
+++ b/pyscf/agf2/uagf2.py
@@ -875,8 +875,8 @@ def _make_qmo_eris_incore(agf2, eri, coeffs_a, coeffs_b, spin=None):
     # npaira, npairb = nmoa*(nmoa+1)//2, nmob*(nmob+1)//2
     cia, cja, caa = coeffs_a
     cib, cjb, cab = coeffs_b
-    nia, nja, naa = [x.shape[1] for x in coeffs_a]
-    nib, njb, nab = [x.shape[1] for x in coeffs_b]
+    nia, nja, naa = (x.shape[1] for x in coeffs_a)
+    nib, njb, nab = (x.shape[1] for x in coeffs_b)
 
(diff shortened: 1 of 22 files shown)


sunqm:
I don't see particular advantages of such changes involving coding style. Unless there are reasons that I did not realize, I tend to reject this update.

verena-neufeld:
Separate comment: I wonder whether we could remove some of the to be ignored rules in .flake8. That is I think maybe we could follow more of the flake8 rules.

fchapoton:
this avoids creating the list, so it is slightly more efficient

fchapoton:
and by the way, the linter `ruff` only suggests code improvements, and not cosmetic code formatting changes.

https://docs.astral.sh/ruff/rules/

sunqm:
I don't think it is a good practice to follow ruff's suggestion in this case. It basically replaces a small list with an iterator. Technically, this optimization depends on CPython versions and it does not always improve performance. In certain CPython versions, there are specialized OP code to unpack a list of two or three elements. The iterator has no performance advantage in these cases. Even the iterator is faster than list comprehension for a few nano seconds, the performance gain here is negligible since the relevant code are only executed a few times. 

I'd suggest to focus on the optimization for the algorithm or architecture. Unless a significant performance improvement, optimization related to coding style should be discouraged.

sunqm:
> Separate comment: I wonder whether we could remove some of the to be ignored rules in .flake8. That is I think maybe we could follow more of the flake8 rules.

As long as the code is clear and easy to read, I think it is acceptable to write pyscf code in any style, not necessarily constrained by flake8 or other style cops. Actually, I wish to add more ignore rules in flake8 and pylint. But I guess that may lead to more arguments.

PR #2115, opened 2024-03-05
Improve memory usage for RPA

yangdatou:
- outcore direct RPA
- try to raise something when loading libpbc.get_Gv

### DIFF
--- a/pyscf/gw/rpa.py
+++ b/pyscf/gw/rpa.py
@@ -26,7 +26,8 @@ Method:
     X. Ren et al., New J. Phys. 14, 053020 (2012)
 """
 
-import numpy as np
+import numpy as np, scipy
+
 from pyscf import lib
 from pyscf.lib import logger
 from pyscf.ao2mo import _ao2mo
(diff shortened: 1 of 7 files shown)


sunqm:
Please put the DF-RPA in a separated implementation. I'm not clear if DF-RPA is a more common choice. If that's true, you can make the DF version as the default one. But the old program without DF should not be removed.

yangdatou:
> Please put the DF-RPA in a separated implementation. I'm not clear if DF-RPA is a more common choice. If that's true, you can make the DF version as the default one. But the old program without DF should not be removed.

Hi Qiming, thank you for the comments. The current implementation in pyscf is in fact DF-RPA. (https://github.com/pyscf/pyscf/blob/master/pyscf/gw/rpa.py) The main difference in my code is the outcore version of AO2MO and forming dielectric matrix in aux basis. If you think it is necessary, I would be glad to add RPA without DF.

sunqm:
I see. You don't have to implement more. I will read the code later.

tberkel:
Just confirming what @yangdatou is saying: DF-RPA is really the "standard" implementation of RPA. Without DF, I don't think there's a way to do RPA with less than N^6 scaling.

pyscf/gw/rpa.py
```diff
@@ -36,85 +37,104 @@
...
 
     # Compute total energy
     e_tot = e_hf + e_corr
-
     logger.debug(rpa, '  RPA total energy = %s', e_tot)
     logger.debug(rpa, '  EXX energy = %s, RPA corr energy = %s', e_hf, e_corr)
 
     return e_tot, e_hf, e_corr
 
-def get_rpa_ecorr(rpa, Lpq, freqs, wts):
```
sunqm, line 85: Why these methods are removed? Are they problematic?

pyscf/pbc/gto/cell.py
```diff
@@ -589,13 +589,21 @@ def get_Gv_weights(cell, mesh=None, **kwargs):
...
+
+    try:
+        fn(Gv.ctypes.data_as(ctypes.c_void_p),
+        rx.ctypes.data_as(ctypes.c_void_p),
+        ry.ctypes.data_as(ctypes.c_void_p),
+        rz.ctypes.data_as(ctypes.c_void_p),
+        mesh.ctypes.data_as(ctypes.c_void_p),
+        b.ctypes.data_as(ctypes.c_void_p))
+
+    except Exception:
```
sunqm, line 604: Why calling the C function can fail? Can this error be more specific?

sunqm (approved):

pyscf/pbc/gto/cell.py
yangdatou: I updated pyscf to master and didn't compile the new binary file. The code breaks in some weird place. Adding this assert might help people (like me) to locate the problem. It shall also be the common practice in pyscf when call C functions.

sunqm: If the symbol not available from the C extension, the code should be failed at the line `fn = getattr(libpbc, 'get_Gv', None)` . You can change it to `fn libpbc.get_Gv` , which will give an accurate error info. Catching errors for the C caller is misleading.

PR #2117, opened 2024-03-06
activate flake8 check for W391 (no empty final lines)

fchapoton:
as another easy check to be done by the flake8 linter

The changes here were mostly scripted using autopep8, plus a few ones hand-made.


sunqm:
In next pyscf dev meeting (I don't know when), I'll raise a discussion about the coding style. If people can get some kinds of agreement on coding style, we can systematically clean up the coding style of the package.

I'll close this pure coding style update for now.

verena-neufeld:
I also think we should decide on a code style to follow, pep8 being the default if no other style is in place. Once we have decided, we should clean up the code following that agreed upon style to improve code readability. As you know, pep8 has the advantage that automatic tools make it easy to clean up code, but of course it can be restrictive.

PR #2132, opened 2024-03-14
Build with scikit-build-core backend rather than setuptools

chillenb:
I've gotten PySCF to build with scikit-build-core, using a pyproject.toml file and no setup.py.

The benefits of this are:
- setup.py now seems to be disfavored as a way to build python packages
- scikit-build-core integrates well with CMake and is recommended by the NumPy developers
- you can [pass arguments to CMake](https://scikit-build-core.readthedocs.io/en/latest/configuration.html) much more easily, without editing any files.
- scikit-build-core has a convenient experimental capability to [recompile on initial import for editable installations](https://scikit-build-core.readthedocs.io/en/latest/configuration.html#editable-installs).

I was also able to get libcint to build as a submodule using `FetchContent`, further simplifying the management of compile options. (although I had to change libcint's CMakeLists to make it work.) If there are downsides to `FetchContent`, it can easily be switched back to `ExternalProject_Add`.

The traditional build method (clone and cmake) still works fine.
Give it a spin if you get bored!

GitHub won't let me aim this pull request at a new branch. Probably a bad idea to merge it into the master branch right now


sunqm:
This update will break the wheel release pipeline.

The current setup.py is not just a simple call to cmake. It also include platform specific customization for the wheel build process. These functions should be properly migrated before abandoning setup.py.  I think the static configurations of setup.py can be moved to pyproject.toml . The remaining functions can be left in setup.py. And the installation can let pyproject.toml invoke setup.py to compile the package.

chillenb:
Thanks for this valuable feedback. Definitely would not be good to break the wheel release pipeline!
I will restore those platform-specific customizations and check for other potential problems.

chillenb:
@sunqm It took a while, but I was able to make the CI and wheel release pipelines work. I have also checked that libxsmm and fftw3 build properly. As a bonus, libxc, libcint, and xcfun now all compile simultaneously with correct `rpath` and the compile options get passed in from the top level build.

The modified wheel release workflow is in the file `.github/workflows/publish-skbuildcore.yml`. It just builds the wheels; the actual publishing part is commented out. Have not yet checked the Docker part.

chillenb:
[Successful CI run](https://github.com/chillenb/pyscf/actions/runs/8352350267)

[Wheel release test](https://github.com/chillenb/pyscf/actions/runs/8352320022/job/22862225005)

.github/workflows/ci_linux/python_deps.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 python -m pip install --upgrade pip
-pip install "numpy!=1.16,!=1.17" "scipy!=1.5" h5py pytest pytest-cov pytest-timer
+pip install "numpy!=1.16,!=1.17" "scipy!=1.5" h5py pytest pytest-cov pytest-timer build
```
sunqm, line 3: Why do you need to introduce the package build?

.github/workflows/publish-skbuildcore.yml
sunqm, line 1: Please modify the existing workflow file instead of adding a new workflow

.github/workflows/release_info.sh
sunqm, line 1: You can add a new filter, for `_version.py` , the existing rules for `__init__.py` should not removed

conda/build.sh
```diff
@@ -13,8 +13,10 @@ set -x -e
 
 # C extensions must be installed with sequential BLAS library
 # https://pyscf.org/install.html#using-optimized-blas
-export CMAKE_CONFIGURE_ARGS="-DWITH_F12=OFF -DBLA_VENDOR=Intel10_64lp_seq"
+# Finding sequential BLAS is is now handled in pyscf/lib/CMakeLists.txt
+
+export CMAKE_ARGS="-DWITH_F12=OFF"
```
sunqm, line 18: Please do not change the existing compiling configurations. On conda,  mkl is configured as default dependency. The setting for BLA_VENDOR is necessary to ensure to link to correct mkl library.

setup.py
sunqm, line 1: The setup.py customize the output format of wheel. This should not be deleted. Please run the setup.py in pyproject

pyscf/lib/agf2/CMakeLists.txt
```diff
@@ -17,6 +17,6 @@ add_library(agf2 SHARED
 add_dependencies(agf2 cvhf)
 
 set_target_properties(agf2 PROPERTIES
-  LIBRARY_OUTPUT_DIRECTORY ${PROJECT_SOURCE_DIR})
+  LIBRARY_OUTPUT_DIRECTORY ${PYSCF_LIBOUTDIR_ROOT})
```
sunqm, line 20: This change will break the search paths  in load_library function. Please do not change any of the cmake configurations.

pyproject.toml
```diff
@@ -0,0 +1,78 @@
...
+shciscf = ["pyscf-shciscf"] # broken
+
+[tool.scikit-build]
+cmake.source-dir = "pyscf/lib"
+cmake.verbose = true
+logging.level = "DEBUG"
+wheel.install-dir = "pyscf"
+wheel.py-api = "py3"
+
+[tool.scikit-build.metadata.version]
```
sunqm, line 76: Please use the old way to find version. There are other triggers rely on the version updates in `__init__.py`

chillenb:
Thanks for reviewing this and providing feedback! This PR should definitely not be merged as-is and I will take a conservative approach for any future pull requests. Even so, it was instructive for me to experiment with the build system.

conda/build.sh
chillenb: It probably would have been better to stick to the old rule "if it ain't broke, don't fix it." But in my defense, I did make sure the behavior would stay the same and it would link MKL:
https://github.com/chillenb/pyscf/blob/bc6dab5ed7ed0c2d9bd909603a371f52e14d42e3/pyscf/lib/CMakeLists.txt#L88

pyscf/lib/agf2/CMakeLists.txt
chillenb: I really did try not to break anything, I swear! See [these lines](https://github.com/chillenb/pyscf/blob/bc6dab5ed7ed0c2d9bd909603a371f52e14d42e3/pyscf/lib/CMakeLists.txt#L22C1-L26C8):

```
if(DEFINED SKBUILD)
  set(PYSCF_LIBOUTDIR_ROOT ${PROJECT_BINARY_DIR})
else()
  set(PYSCF_LIBOUTDIR_ROOT ${PROJECT_SOURCE_DIR})
endif()
```

on the other hand, I get it. "if it ain't broke, don't fix it."

PR #2164, opened 2024-04-09
reset enuc in _build_supcell_

hongzhouye:
In PR #2078, the nuclear energy in periodic HF was changed from `cell.energy_nuc()` to `cell.enuc`. The latter allows reusing previously calculated nuclear energy saved in `cell._enuc`. This causes a bug when making a supercell from a unit cell where `_enuc` is already set because the supercell inherits `_enuc` from the unit cell. This PR fixes the bug.


fishjojo (approved):

fishjojo:
Maybe also add a check in https://github.com/pyscf/pyscf/blob/d57f1d6c89c723e11a7f0933380a6139ba372554/pyscf/pbc/tools/test/test_pbc.py#L141

Issue #2178, opened 2024-04-17
Modify DFT grid schemes


KarimAED:
The DFT grid schemes currently exhibit a few behaviours that seem odd / could be implemented with improvements:

1. The Treutler radial grids are implemented with different parametrization for $\xi$ to the one suggested in Treutler's original paper.
2. Different radial shells [are shuffled](https://github.com/pyscf/pyscf/blob/master/pyscf/dft/gen_grid.py#L230) in `gen_atomic_grids`. This has to then be compensated for with the `sort_grids` kwarg.
3. This kwarg in turn **cannot** be overridden in calls to the top-level API, making this routine inflexible.

Point 1 is simply a divergence from the described optimal behaviour in the paper, which I believe should be fixed.

Points 2 & 3 constitute an issue for our work, as we are trying to validate our own code (DQC) against PySCF. However, there is no way of aligning the grids without outright sorting them for both codes or changing the way PySCF approaches this. As the former is expensive to do repeatedly, and I believe that the PySCF routine is currently suboptimal, I suggest the latter.

I am happy to contribute a PR addressing these points.

susilehtola:
Per 1, do you mean that PySCF uses different values for the atomic radii?

KarimAED:
No, I don't think that is the issue. If you have a look at [the Treutler paper](https://pubs.aip.org/aip/jcp/article-abstract/102/1/346/482489/Efficient-molecular-numerical-integration-schemes?redirectedFrom=fulltext), it parametrizes the M4 radial transform (eq. 19) in terms of $\xi$, a value that is optimized on a per-element basis in a table further down in the paper. In DQC, these values are implemented as seen in [the code](https://github.com/OxfordHED/dqc/blob/master/dqc/grid/factory.py#L175). However, to obtain the same number of grid points as I get with PySCF, this needs to be overwritten with $\xi = 1$ in DQC. The resulting grids then contain the same points (up to padding and sorting).

The $\xi$ factor changes the spacing relative to the atomic radii values, thereby changing how many radial points correspond to a given Lebedev order in the NWChem truncation.

KarimAED:
[...]

PR #2186, opened 2024-04-25
The missing quadrature adjustment in treutler_ahlrichs radial grids

sunqm:
Related to issue #2178

### DIFF
--- a/pyscf/df/test/test_df.py
+++ b/pyscf/df/test/test_df.py
@@ -135,12 +135,12 @@ class KnownValues(unittest.TestCase):
         with tempfile.NamedTemporaryFile() as ftmp:
             mf.with_df._cderi_to_save = ftmp.name
             mf.run()
-        self.assertAlmostEqual(mf.e_tot, -103.4965622991, 7)
+        self.assertAlmostEqual(mf.e_tot, -103.4965622991, 6)
 
         mol.max_memory = 4000
         mf = mol.RKS(xc='lda+0.5*SR_HF(0.3)').density_fit()
         mf.run()
-        self.assertAlmostEqual(mf.e_tot, -103.4965622991, 7)
+        self.assertAlmostEqual(mf.e_tot, -103.4965622991, 6)
 
 if __name__ == "__main__":
     print("Full Tests for df")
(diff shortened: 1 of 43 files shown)


sunqm:
Does this change look reasonable to you @KarimAED ?

KarimAED:
Yes, this makes sense to me

susilehtola:
IMHO the atomic radius adjustment should be factored out of the radial rule, since the equations are the same for all rules: scale radius by $r_i \to Rr_i$ and weight by $w_i \to R^3 w_i$.

sunqm:
> IMHO the atomic radius adjustment should be factored out of the radial rule, since the equations are the same for all rules: scale radius by ri→Rri and weight by wi→R3wi.

I don't understand your suggestion. Do you suggest to apply the adjustment parameters for all radius grids schemes, not just the Treutler-Ahlrichs grids?

susilehtola:
> > IMHO the atomic radius adjustment should be factored out of the radial rule, since the equations are the same for all rules: scale radius by ri→Rri and weight by wi→R3wi.
> 
> I don't understand your suggestion. Do you suggest to apply the adjustment parameters for all radius grids schemes, not just the Treutler-Ahlrichs grids?

Yes, using the same atomic radii for all stages of the DFT quadrature: both radial grid generation and pruning. The scaling relations are the same regardless of the way the radial grid is generated, so using a common implementation saves effort and removes code duplication.

KarimAED:
Hmm, I'm not sure if I agree 100% @susilehtola. I think the code duplication is quite minimal, and I think there is some added value in the code following the original sources more closely.

Note, however, that I am speaking not as a regular maintainer, but as someone who recently tried to get their head wrapped around the code. In that sense, I am biased towards a more verbose code structure, so feel free to ignore my opinions.

susilehtola:
The problem I see is that if you use inconsistent atomic radii in different parts of the code, you are bound to get some weird results...

sunqm:
@KarimAED Treutler-Ahlrichs grids are fixed. To ensure compatibility to old versions, the old grids scheme is still used as the default . You can add a config line in `pyscf_conf.py` to explicitly enable the fixed grid scheme.

In the next board meeting, we should discuss whether to switch to the fixed grids @ssh2 @gkc1000 @sanshar @tberkel

fishjojo (approved):

PR #2192, opened 2024-04-29
move auxbasis_response out of __init__

wxj6000:


sunqm (approved):

wxj6000:
@sunqm Please note `from pyscf import hessian` in `__init__.py`. Without importing hessian, the example `hessian/01-scf_hessian.py` does not work.

pyscf/__init__.py
```diff
@@ -98,6 +98,7 @@ def _get_void_ptr(arr):
 from pyscf import gto
 from pyscf import scf
 from pyscf import ao2mo
+from pyscf import hessian
```
sunqm, line 101: why do you need to import hessian in the global `__init__.py` file?

wxj6000: `.Hessian()` is not defined in SCF. It was injected when import hessian. When we call mf.Hessian(), we have to `from pyscf import hessian` first for now.

sunqm: hessian module will lead an import chain that loads many irrelevant modules. Please remove this line in this PR. It can be discussed in another issue/PR whether these modules should be imported in the global `__init__.py`

wxj6000: OK. Let me create an issue for this.

sunqm (approved):

Issue #1915, opened 2023-10-22
Potential efficiency improvement of density-fitting (cderi, get_jk)


ajz34:
Hi devs!

While performing RI-JK (SCF) computations, I found two possible efficiency bottlenecks. These problems may be alleviated by minor code modifications.

For some remarks below, I think I think I'm not ready to propose a pull request for this issue.

## Coulomb integral not paralleled using `numpy.einsum` without `optimize=True`

For evaluation of coulomb integrals, codes like https://github.com/pyscf/pyscf/blob/231da24f746d3a7988a4c48c9237c4788bda6a36/pyscf/df/df_jk.py#L260-L261 and https://github.com/pyscf/pyscf/blob/231da24f746d3a7988a4c48c9237c4788bda6a36/pyscf/df/df_jk.py#L290-L291 envokes `numpy.einsum`, which have optional parameter `optimize=False`.

For `optimize=False`, `numpy.einsum` just contract tensors or matrices without parallel. That could be great efficiency loss when using more CPU threads.

Simply add `optimize=True` to `numpy.einsum` will resolve this problem.

Since many codes in PySCF uses `numpy.einsum` without `optimize=True`, I think that this is not only a problem of RI-JK SCF, but also other tasks.

## Cholesky decomposed ERI with `scipy.linalg.solve_triangular`

I found that, probably, `scipy.linalg.solve_triangular` just copies or transposes the large 3c-2e integrals `ints` $J_{P, \mu \nu}$ to $J_{\mu \nu, P}$. Since NumPy/SciPy often performs copy or transposition with only one thread, this can be extremely slow.

For evaluation of cholesky decomposed ERI $Y_{Q, \mu \nu}$ (`mf.with_df._cderi`, discuss incore case for simplicity)
$$\sum_{Q} L_{PQ} Y_{Q, \mu \nu} = J_{P, \mu \nu} \text{ or } \mathbf{L} \mathbf{Y} = \mathbf{J}$$
where $L_{PQ}$ is the lower triangular matrix of cholesky-decomposed 2c-2e ERI $J_{PQ}$, as well as $J_{P, \mu \nu}$ the 3c-2e ERI
$$\sum_{R} L_{PR} L_{QR} = J_{PQ} \text{ or } \mathbf{L} \mathbf{L}^\dagger = \mathbf{J}$$

[...]

PR #2205, opened 2024-05-05
Density fitting performance optimization (re issue #1915)

chillenb:
This pull request incorporates enhancements to density fitting proposed in #1915 by @ajz34 .

- Replace unoptimized `np.einsum` with `gemm` via `numpy.matmul`
- Replace `scipy.linalg.triangular_solve` with `?trsm` in order to avoid a single-threaded array transpose and related issues surrounding Fortran vs C order
- Since with trsm, setting `overwrite_b=True` actually _does_ cause `b` to get overwritten, we have to use `map` rather than `lib.map_with_prefetch` when writing the cderi chunks to disk
- Use of `read_direct` in `_load_from_h5g` avoids an unnecessary copy and reads the data right where they need to go

The test case is the same as in the issue: [here](https://github.com/pyscf/pyscf/files/13063667/benchmark_code.py.txt)

Benchmark, 48 cores cascade lake (CPU: 2x 24-core Intel Xeon Gold 6252N)
PySCF compiled with gcc. NumPy installed from conda-forge along with Intel MKL.

1087 cGTOs
| Task | Before | After |
| --- | --- | --- |
| density fitted RHF | 87.5568 s |  31.4202 s |
| CPU time for cholesky_eri | 25.96 s | 8.41 s |
| CPU time for df vj and vk  | 31.67 s |  11.24 s |
| `df.outcore.cholesky_eri_b` | 11.8159 s |  4.2692 s |
| `df.outcore.cholesky_eri` | 22.12 s |  10.08 s |

### DIFF
--- a/pyscf/df/df_jk.py
+++ b/pyscf/df/df_jk.py
@@ -289,7 +289,7 @@ def get_jk(dfobj, dm, hermi=1, with_j=True, with_k=True, direct_scf_tol=1e-13):
         for eri1 in dfobj.loop(blksize):
             naux, nao_pair = eri1.shape
             assert (nao_pair == nao*(nao+1)//2)
-            if with_j:               
+            if with_j:
                 vj += numpy.matmul(numpy.matmul(dmtril, eri1.T), eri1)
 
             for k in range(nset):
--- a/pyscf/df/df_jk.py
+++ b/pyscf/df/df_jk.py
@@ -260,8 +260,8 @@ def get_jk(dfobj, dm, hermi=1, with_j=True, with_k=True, direct_scf_tol=1e-13):
 
     if not with_k:
         for eri1 in dfobj.loop():
-            rho = numpy.einsum('ix,px->ip', dmtril, eri1)
-            vj += numpy.einsum('ip,px->ix', rho, eri1)
+            vj += numpy.matmul(numpy.matmul(dmtril, eri1.T), eri1)
+
 
     elif getattr(dm, 'mo_coeff', None) is not None:
         #TODO: test whether dm.mo_coeff matching dm
@@ -289,9 +289,8 @@ def get_jk(dfobj, dm, hermi=1, with_j=True, with_k=True, direct_scf_tol=1e-13):
         for eri1 in dfobj.loop(blksize):
             naux, nao_pair = eri1.shape
             assert (nao_pair == nao*(nao+1)//2)
-            if with_j:
-                rho = numpy.einsum('ix,px->ip', dmtril, eri1)
-                vj += numpy.einsum('ip,px->ix', rho, eri1)
+            if with_j:               
+                vj += numpy.matmul(numpy.matmul(dmtril, eri1.T), eri1)
 
             for k in range(nset):
                 nocc = orbo[k].shape[1]
@@ -304,7 +303,7 @@ def get_jk(dfobj, dm, hermi=1, with_j=True, with_k=True, direct_scf_tol=1e-13):
                          ctypes.c_int(naux), ctypes.c_int(nao),
                          (ctypes.c_int*4)(0, nocc, 0, nao),
                          null, ctypes.c_int(0))
-                    vk[k] += lib.dot(buf1.T, buf1)
+                    vk[k] += numpy.matmul(buf1.T, buf1)
             t1 = log.timer_debug1('jk', *t1)
     else:
         #:vk = numpy.einsum('pij,jk->pki', cderi, dm)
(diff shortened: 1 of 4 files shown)


pyscf/df/df_jk.py
```diff
@@ -260,8 +260,8 @@ def get_jk(dfobj, dm, hermi=1, with_j=True, with_k=True, direct_scf_tol=1e-13):
 
     if not with_k:
         for eri1 in dfobj.loop():
-            rho = numpy.einsum('ix,px->ip', dmtril, eri1)
-            vj += numpy.einsum('ip,px->ix', rho, eri1)
+            vj += numpy.matmul(numpy.matmul(dmtril, eri1.T), eri1)
```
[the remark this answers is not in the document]
chillenb: Thanks for the suggestion! I personally love the `@` notation but didn't use it here because
- I couldn't find examples of `@` in pyscf and didn't want to be the first
- There are a lot of matrix multiply functions in use (`lib.dot`, `np.dot`, `lib.ddot`) so I thought it would be useful to indicate explicitly which one is being called

To make it more readable, I could split that line or clean it up with functools.reduce

pyscf/df/outcore.py
```diff
@@ -189,7 +197,11 @@ def process(sh_range):
         return dat
 
     feri = _create_h5file(erifile, dataname)
-    for istep, dat in enumerate(lib.map_with_prefetch(process, shranges)):
+
+    # Cannot use lib.map_with_prefetch with trsm because
```
sunqm, line 201: Create two buffers and swap the two buffers, as the incore code does.

pyscf/df/df_jk.py
```diff
@@ -330,7 +331,7 @@ def get_jk(dfobj, dm, hermi=1, with_j=True, with_k=True, direct_scf_tol=1e-13):
                      ctypes.c_int(naux), *rargs)
 
                 buf2 = lib.unpack_tril(eri1, out=buf[1])
-                vk[k] += lib.dot(buf1.reshape(-1,nao).T, buf2.reshape(-1,nao))
+                vk[k] += numpy.matmul(buf1.reshape(-1,nao).T, buf2.reshape(-1,nao))
```
sunqm, line 334: use lib.dot here. dsyrk does not match this pattern.

pyscf/df/df_jk.py
```diff
@@ -304,7 +304,7 @@ def get_jk(dfobj, dm, hermi=1, with_j=True, with_k=True, direct_scf_tol=1e-13):
                          ctypes.c_int(naux), ctypes.c_int(nao),
                          (ctypes.c_int*4)(0, nocc, 0, nao),
                          null, ctypes.c_int(0))
-                    vk[k] += lib.dot(buf1.T, buf1)
+                    vk[k] += numpy.matmul(buf1.T, buf1)
```
sunqm, line 307: Please use `lib.dot` . due to threading parallel efficiency issue in openblas and mkl, the dsyrk in np.dot is often slower than lib.dot for larger number of threads. If you wish to optimize this line using dsyrk, please implement the function in C with explicit openmp parallelization as the lib.dot does.

pyscf/df/df_jk.py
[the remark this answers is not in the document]
sunqm: Many `np.dot`s were implemented before python2.7 was retired. Any np.dot can be replaced by `@`. But as long as they work, it's okay to leave np.dot there.

pyscf/df/df_jk.py
chillenb: Good point about dsyrk. I did not think of that and will fix this.

pyscf/df/outcore.py
chillenb: OK!

pyscf/df/df_jk.py
chillenb: done

chillenb:
@sunqm I've tried to implement the buffer-swapping technique from incore as you suggested, but I am unsure about the proper buffer size. I left buflen alone and the memory usage seems to stay well under max_memory, but I have not done extensive tests. Your advice would be appreciated!

pyscf/df/outcore.py
```diff
@@ -163,6 +168,7 @@ def cholesky_eri_b(mol, erifile, auxbasis='weigend+etb', dataname='j3c',
     #    cintopt = gto.moleintor.make_cintopt(atm, bas, env, int3c)
     cintopt = gto.moleintor.make_cintopt(atm, bas, env, int3c)
     bufs1 = numpy.empty((comp*max([x[2] for x in shranges]),naoaux))
+    bufs2 = numpy.empty_like(bufs1)
```
wxj6000, line 171: It seems that the double buffers are not effective as expected. The second buffer can be used in the `transpose` operation.

chillenb: OK, I did a quick test on a ~15G example and observed that there isn't much difference because the kernel write buffer is already buffering writes.

The double-buffering `does` help if you make HDF5 use the 'direct' IO driver.

sunqm (approved):

sunqm:
I'll leave this PR one more week for more comments before merging. If no more comments, this optimization will be available in next release (pyscf 2.6)

PR #2208, opened 2024-05-08
M3SOSCF Implementation

LinusBDittmer:
This is an implementation of the Markovian Multiagent Monte-Carlo SOSCF based on CIAH. It is a powerful, albeit slow SCF algorithm that enables near-guaranteed convergence even in extremely difficult systems.

Publication: [https://doi.org/10.1063/5.0159737](https://doi.org/10.1063/5.0159737)

The added commits have been rebased and match the latest commit on pyscf:master

### DIFF
--- a/examples/scf/73-m3soscf.py
+++ b/examples/scf/73-m3soscf.py
@@ -1,0 +1,75 @@
+'''
+
+This file details the basic usage of the M3SOSCF algorithm.
+
+'''
+
+from pyscf import gto, scf, dft
+
+
+
+# Building the molecule, here CO is used as a simple example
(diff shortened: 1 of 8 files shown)


sunqm:
The reference should be cited in the code.

I suggest to add this feature in pyscf-forge (https://github.com/pyscf/pyscf-forge) first. There will be not many arguments about whether a new feature is needed. If you feel confusion about how to use pyscf-forge, please contact me or @fishjojo . We are preparing a more detailed contributing guideline for pyscf-forge

matthew-hennefarth:
I believe this PR should be closed for now as it has been merged into [PySCF-forge](https://github.com/pyscf/pyscf-forge/pull/47).

PR #2216, opened 2024-05-16
Population analysis for KSCF

hongzhouye:
The current population analysis code for KSCF object analyzes only the wavefunction at Gamma point. This PR completes the implementation for analyzing the wavefunction from all k-points. The results are equivalent to the corresponding supercell calculation. Old tests are updated.


verena-neufeld:
Thanks @hongzhouye ! Do you have some insights why some numbers in the tests have changed with this PR?

hongzhouye:
@verena-neufeld I believe it's because the old references were generated using the old code, which only uses a Gamma point.

sunqm (approved):

PR #2217, opened 2024-05-17
Fix frozen for FNO

hongzhouye:
In the previous implementation of `FNOCCSD`, the input `frozen` argument was not passed to the MP2 calculation used to generate FNOs. The FNOs generated this way are okay but the ∆MP2 composite correction is wrong. This PR fixes the bug. 

See the figure below for the convergence of the FNOCCSD correlation energy for benzene/cc-pVDZ with carbon 1s electrons being frozen. The uncorrected FNOCCSD energy from the old code is okay compared to the new one, but only the new PT2-corrected energy converges to the correct limit.

fno_cmp

The `make_fno` function was rewritten to handle general `frozen` (integer or list) for both RMP2 and UMP2. Tests are added for both.

NOTE: for UMP2, the natural orbital occupation number for each spin is multiplied by 2 (see [this line](https://github.com/pyscf/pyscf/pull/2217/files#diff-2413d646fb7c38c96523d14de45c6e41aae20d8f32021f21b52af8121fa8810bR318)) so that when the same `thresh` is used, RHF and UHF based FNO calculations generate the same results. I believe that this is a nice behavior but suggestions are appreciated.

### DIFF
--- a/pyscf/cc/test/test_fno.py
+++ b/pyscf/cc/test/test_fno.py
@@ -51,16 +51,16 @@ class Water(unittest.TestCase):
         et = CCSD_T(mcc, eris=eris)
         return mcc.e_corr, et
 
-    @unittest.skip('fail due to updates of pp_int?')
     def test_fno_by_thresh(self):
         threshs = [1e-2,1e-3,1e-4]
         refs = [
-            [-0.1180210293, -0.0001530894],
-            [-0.1903456906, -0.0006322229],
-            [-0.2130000748, -0.0030061317],
+            [-0.1200179470,-0.0001506723],
+            [-0.1925851371,-0.0006428496],
+            [-0.2149635715,-0.0030270215],
         ]
         for thresh,ref in zip(threshs,refs):
             eccsd, et = self.kernel(cc.FNOCCSD, thresh=thresh)
+            # print('[%s],' % (','.join([f'{x:.10f}' for x in [eccsd,et]])))
             self.assertAlmostEqual(eccsd, ref[0], 6)
             self.assertAlmostEqual(et, ref[1], 6)
 
@@ -69,16 +69,16 @@ class Water(unittest.TestCase):
         self.assertAlmostEqual(eccsd, eccsd0, 6)
         self.assertAlmostEqual(et, et0, 6)
 
-    @unittest.skip('fail due to updates of pp_int?')
     def test_fno_by_thresh_frozen(self):
         threshs = [1e-2,1e-3,1e-4]
         refs = [
-            [-0.1173030018, -0.0001459448],
-            [-0.1889586685, -0.0006157799],
-            [-0.2109365568, -0.0029841323],
+            [-0.0777291294,-0.0000381574],
+            [-0.1409238993,-0.0004851219],
+            [-0.1500184791,-0.0021645217],
         ]
         for thresh,ref in zip(threshs,refs):
             eccsd, et = self.kernel(cc.FNOCCSD, thresh=thresh, frozen=1)
+            # print('[%s],' % (','.join([f'{x:.10f}' for x in [eccsd,et]])))
             self.assertAlmostEqual(eccsd, ref[0], 6)
             self.assertAlmostEqual(et, ref[1], 6)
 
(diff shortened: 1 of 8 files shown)


sunqm (approved):

pyscf/cc/test/test_fno.py
```diff
@@ -51,16 +51,16 @@ def kernel(self, CC, **kwargs):
...
-            [-0.1180210293, -0.0001530894],
-            [-0.1903456906, -0.0006322229],
-            [-0.2130000748, -0.0030061317],
+            [-0.1200179470,-0.0001506723],
+            [-0.1925851371,-0.0006428496],
+            [-0.2149635715,-0.0030270215],
         ]
         for thresh,ref in zip(threshs,refs):
             eccsd, et = self.kernel(cc.FNOCCSD, thresh=thresh)
+            # print('[%s],' % (','.join([f'{x:.10f}' for x in [eccsd,et]])))
```
verena-neufeld, line 63: Should this be deleted before merging?

pyscf/cc/test/test_fno.py
```diff
@@ -69,16 +69,16 @@ def test_fno_by_thresh(self):
...
-            [-0.1173030018, -0.0001459448],
-            [-0.1889586685, -0.0006157799],
-            [-0.2109365568, -0.0029841323],
+            [-0.0777291294,-0.0000381574],
+            [-0.1409238993,-0.0004851219],
+            [-0.1500184791,-0.0021645217],
         ]
         for thresh,ref in zip(threshs,refs):
             eccsd, et = self.kernel(cc.FNOCCSD, thresh=thresh, frozen=1)
+            # print('[%s],' % (','.join([f'{x:.10f}' for x in [eccsd,et]])))
```
verena-neufeld, line 81: Delete?

pyscf/cc/test/test_fno.py
```diff
@@ -118,16 +118,16 @@ def kernel(self, CC, **kwargs):
...
-            [-0.1180086745, -0.0001528637],
-            [-0.1903723005, -0.0006323478],
-            [-0.2130687280, -0.0030086849],
+            [-0.1200039667,-0.0001504459],
+            [-0.1926097163,-0.0006429844],
+            [-0.2150297885,-0.0030295410],
         ]
         for thresh,ref in zip(threshs,refs):
             eccsd, et = self.kernel(cc.FNOCCSD, thresh=thresh)
+            # print('[%s],' % (','.join([f'{x:.10f}' for x in [eccsd,et]])))
```
verena-neufeld, line 130: Delete?

pyscf/cc/test/test_fno.py
```diff
@@ -136,16 +136,16 @@ def test_fno_by_thresh(self):
...
-            [-0.1172906846, -0.0001457165],
-            [-0.1889855085, -0.0006158811],
-            [-0.2110052374, -0.0029866440],
+            [-0.0777156383,-0.0000380971],
+            [-0.1409426400,-0.0004850786],
+            [-0.1500564421,-0.0021661544],
         ]
         for thresh,ref in zip(threshs,refs):
             eccsd, et = self.kernel(cc.FNOCCSD, thresh=thresh, frozen=1)
+            # print('[%s],' % (','.join([f'{x:.10f}' for x in [eccsd,et]])))
```
verena-neufeld, line 148: Delete?

pyscf/cc/test/test_fno_uhf.py
```diff
@@ -0,0 +1,93 @@
...
+    def test_fno_by_thresh(self):
+        threshs = [1e-2,1e-3,1e-4]
+        refs = [
+            [-0.1200179470,-0.0001506723],
+            [-0.1925851371,-0.0006428496],
+            [-0.2149635715,-0.0030270215],
+        ]
+        for thresh,ref in zip(threshs,refs):
+            eccsd, et = self.kernel(cc.FNOCCSD, thresh=thresh)
+            # print('[%s],' % (','.join([f'{x:.10f}' for x in [eccsd,et]])))
```
verena-neufeld, line 63: Delete?

verena-neufeld (commented):
I have not fully read it but I am suggesting maybe removing commented out print statements?

fishjojo:
> > I have not fully read it but I am suggesting maybe removing commented out print statements?
> 
> Hi Verena, thanks for the suggestions! These print statements print the values of the correlation energies being checked in these tests. This can ease the debugging of the code in the future by simply uncommenting them. Please let me know if you think we should remove them.

can use something like `logger.debug1`

fishjojo (approved):

PR #2218, opened 2024-05-17
allow passing in options when creating HDF5 files

chillenb:
I am trying to add a way to specify extra options when creating HDF5 files, and this is a first attempt. It hasn't been tested. Are there cleaner / more consistent ways to do it?

### DIFF
--- a/pyscf/lib/misc.py
+++ b/pyscf/lib/misc.py
@@ -1319,7 +1319,7 @@ class h5filewrite_options(temporary_env):
     Examples:
 
     >>> with h5filewrite_options(alignment_interval=4096, alignment_threshold=4096):
-    ...     f = h5py.File('mydata.h5', 'w')
+    ...     f = lib.H5FileWrap('mydata.h5', 'w')
     >>> print(h5py.h5p.PropFAID.get_alignment(f.id.get_access_plist()))
     (4096, 4096)
     '''
--- a/pyscf/lib/misc.py
+++ b/pyscf/lib/misc.py
@@ -23,6 +23,7 @@ Some helper functions
 import os
 import sys
 import time
+import random
 import platform
 import warnings
 import tempfile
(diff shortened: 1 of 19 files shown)


matthew-hennefarth:
Another option for this would be to create a wrapper function to `hdf5.File()` in `pyscf.lib` which essentially always adds the kwargs `**param.H5F_WRITE_KWARGS`. That is

in pyscf/lib somewhere:
```python
def h5py_write(filename, mode, **kwargs):
    if kwargs:
        write_options = param.H5F_WRITE_KWARGS.copy()
        write_options.update(kwargs)

    else:
        write_options = param.H5F_WRITE_KWARGS

    h5py.File(filename, mode, **write_options)
```
This way the rest of the code base just calls `h5py_write` without having to pass the `**param.H5F_WRITE_KWARGS` each time. But the `if kwargs` statement above would still allow someone to pass some additional kwargs/update the default kwargs if they wanted.

matthew-hennefarth:
The downside to your approach is that if someone adds a new h5py.File() call, they may not include the `**param.H5F_WRITE_KWARGS` option. Whereas if one uses a wrapper function in the library, all `h5py.File` calls always have the same `**kwargs`. Of course, this requires that all usages of `h5py.File` are replaced, so it is a bit of a trade-off.

chillenb:
Thanks for the suggestion! I like your wrapper idea and will switch to it soon.

chillenb:
OK, I've implemented Matthew's wrapper idea! This is now ready for review.

Also, as part of a program to fight a bug where outcore (especially PBC) DF randomly hangs on NFS, I reimplemented H5TmpFile so that the temporary file is not opened in both the HDF5 library and Python, with two separate file descriptors. The procedure for generating unique filenames is unchanged.

chillenb:
@sunqm
If you have the time to look at this before the 2.6 release, I'd be very grateful! My group is running into the aforementioned HDF5+NFS bug fairly often these days, and it has been difficult to track down the cause or reliably reproduce the bug.

Hopefully this PR is the end of that bug -- if not, the ability to pass in extra HDF5 options would really help with troubleshooting!

sunqm:
Could you open an issue to describe the issue you encountered in NFS? Is it only caused by the H5TmpFile instances, or all HDF5 files?

This PR introduces an abstract to manage the h5py files, which is generally not bad IMO. However, it's not clear what the HDF5+NFS issues exactly are. How would one reproduce it or avoid it by using this PR. Simply applying this PR does not seem to solving the problem. If others meet the similar problem, they might not get information from this PR.

This update will not be applied in 2.6.0 release, as 2.6.0 will be packaged soon. We can target to include this PR in the patch release of 2.6, if problems can be analyzed in more details in a separated github issue.

chillenb:
@sunqm

Thanks for looking at this for me! I've just opened a separate issue describing the bugs my lab has encountered (#2248).

Sadly, the first of those bugs will probably be difficult to diagnose properly. I will do my best and let you know if I find any useful information.

sunqm (approved):

pyscf/lib/misc.py
```diff
@@ -1249,6 +1312,20 @@ def __enter__(self):
...
         return self.c
 
+class h5filewrite_options(temporary_env):
+    '''Within the context of this macro, extra keyword arguments are
+    passed to h5py.File() whenever an HDF5 file is opened for writing.
+
+    Examples:
+
+    >>> with h5filewrite_options(alignment_interval=4096, alignment_threshold=4096):
+    ...     f = h5py.File('mydata.h5', 'w')
```
sunqm, line 1322: Isn't it `f = lib.H5FileWrap('mydata.h5', 'w')` ?

chillenb: You're right. My bad!

chillenb: Fixed now.

PR #2228, opened 2024-05-24
change default basis linear dependence threshold

wxj6000:
1e-9 -> 1e-7


sunqm:
Threshold settings in https://github.com/pyscf/pyscf/blob/master/pyscf/lib/linalg_helper.py#L34 can also be updated accordingly

sunqm:
This PR impacts
* The linear dependency threshold for the auxiliary basis. For range-separated functionals, it may slightly change the results when using density fitting (~ 1e-7 on energy).
* The Krylov subspace solver. It may slightly impact the convergence of CPHF solver. This impact should be negligible as long as the CPHF is converged.

I believe this update is necessary as various numerical stability issues were observed in range-separated functional due to the threshold 1e-9 for density fitting methods. Slightly loosening it seems helpful. The question is which default value should be set. This PR proposes 1e-7, following 10.1021/acs.jctc.3c00670 . @susilehtola do you have comments on this setting?

I tend to update the linear dependency threshold in pbc.df modules as well. However, it may affect PBC density fitting in all calculations (not just the RSH functionals as that in molecule) since PBC auxbasis is more likely getting linear dependency issue. Do you have any comments on this @hongzhouye @tberkel

pyscf/lib/linalg_helper.py
```diff
@@ -31,9 +31,9 @@
 from pyscf.lib import misc
 from pyscf import __config__
 
-SAFE_EIGH_LINDEP = getattr(__config__, 'lib_linalg_helper_safe_eigh_lindep', 1e-15)
+SAFE_EIGH_LINDEP = getattr(__config__, 'lib_linalg_helper_safe_eigh_lindep', 1e-13)
```
sunqm, line 34: This setting affects the mcscf convergence. Please change it back to 1e-15 for backward reproduciblity .

susilehtola:
@sunqm in the above-mentioned paper [J. Chem. Theory Comput. 2023, 19, 6242−6254](https://doi.org/10.1021/acs.jctc.3c00670), I only used the value `1e-7` in the pivoted Cholesky decomposition of the atomic electron repulsion integral matrix to produce the primitive auxiliary basis sets; I found the value `1e-6` in the Cholesky decomposition to lead to more noticeable errors. After this, the auxiliary basis was contracted with a singular value decomposition.

The actual molecular calculations employed `1e-7` as the basis set linear dependency threshold in both Psi4 and Gaussian. However, I did not change the Psi4 default RI threshold, which is [`1e-10` by default](https://psicode.org/psi4manual/master/autodoc_glossary_options_c.html#term-DF_FITTING_CONDITION-SCF).

sunqm (approved):

sunqm:
Since there is no more comment, I'll merge this PR in one week. Let's target the problem of PBC auxiliary basis in another issue.

pyscf/df/incore.py
```diff
@@ -28,7 +28,9 @@
 
 
 MAX_MEMORY = getattr(__config__, 'df_outcore_max_memory', 2000)  # 2GB
-LINEAR_DEP_THR = getattr(__config__, 'df_df_DF_lindep', 1e-9)
+# LINEAR_DEP_THR cannot be below 1e-7,
+# see qchem default setting in https://manual.q-chem.com/5.4/sec_Basis_Customization.html
+LINEAR_DEP_THR = getattr(__config__, 'df_df_DF_lindep', 1e-7)
```
susilehtola, line 34: The comment is misleading. This is the threshold for density fitting, while the Q-Chem manual refers to the linear dependence threshold for the orbital basis set.

susilehtola: Note that Psi4 uses the value `1e-10` by default for the auxiliary basis.

I believe Q-Chem does not employ a linear dependency threshold, instead opting to solve the RI equations directly.

wxj6000: > Note that Psi4 uses the value `1e-10` by default for the auxiliary basis.
> 
> I believe Q-Chem does not employ a linear dependency threshold, instead opting to solve the RI equations directly.

`1e-10` is probably still OK for SCF. But the numerical issues are observed in gradient and Hessian calculations where multiple inversions/solves are involved. To be consistent, I propose we use a larger threshold for the calculations. I was not sure how it affects the accuracy.

For Q-Chem, it was my best guess. I could be wrong. But solving RI equations also requires some thresholds, right? Otherwise, the equations can be ill-conditioned. I didn't find the setting anywhere else.

susilehtola: There's no threshold when you do a full solve for the linear equation ${\bf S} {\bf C} = \boldsymbol{\gamma}$ for the expansion coefficients ${\bf C}$ with the auxiliary basis Coulomb overlap ${\bf S}$ from the source terms $\gamma_A = \sum_{rs} (A|rs) P_{rs}$.

sunqm: The linear dependency issue is mostly found in the RSH functional, where pyscf computes the density fitting for the LR part. This is highly singular and numerical issues are noticed with 1e-10 threshold. I think the threshold 1e-10 or 1e-7 do not impact the density fitting for SR part, the full-range Coulomb, since there are probably very few eigenvalues between 1e-10 and 1e-7. In the case of density fitting of LR, there are many eigenvectors near 1e-10. The linearly independent sets can be quite different for geometries with small differences.

PR #2241, opened 2024-06-01
Optimize nuclear hessian

sunqm:
- [x] Remove chkfile and I/O in hessian
- [x] Improve CPHF solver convergence


pyscf/hessian/rhf.py
```diff
@@ -488,7 +457,7 @@ def __init__(self, scf_method):
         self.verbose = scf_method.verbose
         self.stdout = scf_method.stdout
         self.mol = scf_method.mol
-        self.chkfile = scf_method.chkfile
+        self.base = scf_method
```
wxj6000, line 460: Is this defined twice?

pyscf/lib/linalg_helper.py
```diff
@@ -1384,72 +1386,37 @@ def krylov(aop, b, x0=None, tol=1e-10, max_cycle=30, dot=numpy.dot,
...
     if x0 is not None:
         x += x0
     return x
 
 
-def solve(aop, b, precond, tol=1e-12, max_cycle=30, dot=numpy.dot,
+def solve(aop, b, precond=None, tol=1e-12, max_cycle=60, dot=numpy.dot,
           lindep=DSOLVE_LINDEP, verbose=0, tol_residual=None):
-    '''Davidson iteration to solve linear equation.
+    '''Solve linear equation using the scipy.sparse module
```
fishjojo, line 1410: maybe specify that GMRES is used, or we could rename the function as something like `gmres_wrapper`

fishjojo (approved):

PR #2253, opened 2024-06-06
micro-optimisation about len in tests

fchapoton:
mainly changing some if condition to avoid using len

also a few other minor code details, to simplify things


fchapoton:
ok, numpy does not like that.. let me close

PR #2255, opened 2024-06-09
traceless quadrupole moment implementation for SCF

chillenb:
Here's a method to get the traceless quadrupole moment from an SCF calculation. It's very similar to dip_moment. I've added simple tests for RHF and ROHF.

Does anyone have suggestions for further tests or comparison to literature values?


susilehtola:
For consistency, the quadrupole moment should be computable with the same origin as the dipole moment, yet the dipole moment lacks the possibility to specify the used origin. For a charge neutral system, the dipole is of course origin independent, but it does depend on the origin for a charged system!

chillenb:
@susilehtola How about this---both dip_moment and quad_moment use the center of nuclear charge by default and accept the `with_origin` keyword argument. In addition, `dip_moment` prints a warning if the total charge is nonzero.

chillenb:
For backward compatibility and to pass tests, I decided to use (0,0,0) as the default origin for both dipole and quadrupole moments. The tests should now pass.

sunqm (approved):

Issue #2260, opened 2024-06-17
PySCF is incompatible with NumPy 2.0


wangenau:
[NumPy 2.0.0](https://github.com/numpy/numpy/releases/tag/v2.0.0) got released yesterday with a lot of changes and currently PySCF does not work with it. This is especially unpleasant since the installation has no bounds (e.g. `numpy<2`) and new installations will pull the newest numpy version, resulting in a broken environment.

One (incomplete) way to check the compatibility is to run [ruff](https://github.com/astral-sh/ruff) to find deprecated functions:
`ruff check --select NPY`

These issues can also be automatically fixed (except for `NPY002`, but this one should work even with numpy 2.0) with
`ruff check --select NPY --fix`

But there are more issues that are not catched by Ruff, e.g,
https://github.com/pyscf/pyscf/blob/master/pyscf/dft/numint.py#L1144
https://github.com/pyscf/pyscf/blob/master/pyscf/dft/numint.py#L1269
where both lines will fail with
`ValueError: setting an array element with a sequence.`

On a side note, ruff also reveals some broken examples that can not be parsed, namely:
* examples/1-advanced/020-dmrg_casscf_nevpt2_for_FeS.py
* examples/1-advanced/019-state_average_dmrg_casscf.py
* examples/mcscf/33-make_init_guess/43-dmet_cas.py
* examples/mrpt/03-dmrg_nevpt2.py
* examples/local_orb/nlocal.py
* examples/mcscf/33-make_init_guess/32-dmrg_casscf_nevpt2_for_FeS.py
* examples/mcscf/43-dmet_cas.py
* examples/local_orb/pmloc.py

wangenau:
The Ruff check option was sadly not enough to fix all NumPy 2.0 compatibility issues. PySCF 2.6.1 still fails for simple examples like https://github.com/pyscf/pyscf/blob/master/examples/dft/00-simple_dft.py .

chillenb:
A lot of it is due to the fact that `np.array(..., order='C', copy=False)` now throws an error if a copy needs to be made...
Here's a conda-forge test log showing some of the failing tests: [link](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=957905&view=logs&jobId=4f922444-fdfe-5dcf-b824-02f86439ef14&j=4f922444-fdfe-5dcf-b824-02f86439ef14&t=937c195f-508d-5135-dc9f-d4b5730df0f7)

wangenau:
[...]

PR #2262, opened 2024-06-17
Add ruff check

sunqm:
Also updates the legacy code due to numpy 2.0 release and cleans up tests (issue #2260)


fishjojo (approved):

.ruff.toml
wangenau, line 1: The Ruff configuration file can probably be trimmed down with `respect-gitignore = true`.
Most of the values in `exclude` are excluded [by default](https://docs.astral.sh/ruff/settings/#exclude), most of the other additions can probably savely added to the gitignore file, like `.ipynb_checkpoints`.

And `indent-width`, `dummy-variable-rgx`, and the `format` section only use default values which *could* be removed as well.

`quote-style = "single"` could be added to the format section, since most modules use single quotes over double quotes from what I have seen.

PR #2268, opened 2024-06-19
fix conda package CI

chillenb:
Most of the packages used in the build process are installed from conda-forge, especially gcc and mkl. I think removing the anaconda channel might fix the 2.6.1 build error, but I'm not 100% sure.

The conda-forge channel supplies `anaconda-client`, so it should work; if not, then `conda install -y -c anaconda anaconda-client`, then `conda install -y conda-build` should install Anaconda's version without bringing in Anaconda g++ and gfortran.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -168,7 +168,7 @@ jobs:
         uses: s-weigand/setup-conda@v1
         with:
           update-conda: true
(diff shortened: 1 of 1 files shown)


chillenb:
Testing the workflow here
https://github.com/chillenb/pyscf/actions/runs/9574191987/job/26397008419

sunqm (approved):

PR #2276, opened 2024-06-19
Optimize fft performance

sunqm:


fishjojo (approved):

chillenb:
@sunqm by any chance, have you tried the mkl_fft package? I wonder if it could be useful here. If can recommend a benchmark to run, I can try to compare pyscf's performance using mkl_fft and scipy_fft!

sunqm:
I actually have benchmarked various settings, various problem size for various FFT libraries in my new book "Python for quantum chemistry" (in progress). For example, this is results of 2D FFT using 8 threads. 
[chap04-fft2-8t.pdf](https://github.com/user-attachments/files/16121728/chap04-fft2-8t.pdf)

There are some observations:
1. The MKL fft does not show particular advantages for high-dimensional FT, compared to scipy.fft or numpy.fft
2. MKL fft can be easily enabled as the backend for scipy.fft.
```
with scipy.fft.set_backend(mkl_fft.interfaces.scipy_fft):
        import pyscf
        ...
```

That's why I updated the implementation to scipy.fft in this PR .

chillenb:
@sunqm

The following benchmark might interest you:

time_48thread_fft3d

Sadly, mkl_fft is only fast for these 3D transforms if `axes=(0,1,2)` and `overwrite_x=True`. Most 3D FFTs in PySCF aren't of this type.

mkl_fft is slow even if the input array is shaped like (1,N,N,N) and `axes=(1,2,3)`. The wrapper probably doesn't recognize this special case and takes a slower code path.

PR #2277, opened 2024-06-19
NumPy 2 compatibility

chillenb:
These small tweaks fix many of the tests with NumPy 2.0.

- np.array and np.asarray: the meaning of the `copy` kwarg is different; see https://numpy.org/doc/stable/release/2.0.0-notes.html#copy-keyword-changes-2-0

- np.unique behaves differently: https://github.com/numpy/numpy/issues/26738


chillenb:
Oops, they're already fixed :S

PR #2278, opened 2024-06-20
Fix wheel build for setuptools 70

sunqm:


setup.py
```diff
@@ -83,11 +83,22 @@ def run(self):
...
+initialize_options_1 = bdist_wheel.initialize_options
 def initialize_with_default_plat_name(self):
-    initialize_options(self)
+    initialize_options_1(self)
     self.plat_name = get_platform()
+    self.plat_name_supplied = True
+bdist_wheel.initialize_options = initialize_with_default_plat_name
+
+# For setuptools>=70
+from setuptools.command.bdist_wheel import bdist_wheel
```
chillenb, line 96: Will this be okay if you install from source (eg `pip install --no-binary pyscf pyscf`) but have setuptools <70.1.0?

Not sure whether it is better to require setuptools >=70.1, check the version number of setuptools, or check whether setuptools.command.bdist_wheel exists...

sunqm: Actually, I'm not sure which version of setuptools causes the problem. Just a general fix so that all setuptools>61 work

fishjojo (approved):

setup.py
chillenb: According to setuptools' webpage, bdist_wheel was [added](https://setuptools.pypa.io/en/latest/history.html#v70-1-0) in the latest version, 70.1.0. Given the timing of the Github Actions build failure, this version seems to be the culprit!

PR #2288, opened 2024-06-24
Fixed molden.py by shifting coordinates in header to center of mass

### LINKED ISSUE #2287: PySCF produces wrong Molden input file
Hi everyone,

I have found an issue with the Molden file produced by PySCF. For the following minimal example, molden is unable to visualize the orbitals:

```
from pyscf import gto, scf, tools

mol = gto.Mole()
mol.build(
    atom="""
        C
        H  1  1.1015
        H  1  1.1015  2  109.5
        H  1  1.1015  2  109.5  3  120.0
        H  1  1.1015  2  109.5  3  240.0
    """,
    basis="cc-pVDZ",
)

hf = scf.RHF(mol).run()

tools.molden.from_mo(mol, "mo_coeff.molden", hf.mo_coeff)
```

This also appears to be the case for some other symmetric molecules. I was able to fix this issue by shifting the coordinates in the header to the center of mass and will add a corresponding PR.

jonas-greiner:
Fixes #2287.

### DIFF
--- a/pyscf/tools/molden.py
+++ b/pyscf/tools/molden.py
@@ -417,11 +417,14 @@ def header(mol, fout, ignore_h=IGNORE_H):
     fout.write('[Molden Format]\n')
     fout.write('made by pyscf v[%s]\n' % pyscf.__version__)
     fout.write('[Atoms] (AU)\n')
+    mass = mol.atom_mass_list()
+    coords = mol.atom_coords()
+    com = numpy.einsum('i,ij->j', mass, coords) / mass.sum()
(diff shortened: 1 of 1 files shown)


sunqm:
The patch itself doesn't have any issues. However, I don't understand why the Molden format requires a shift in the molecule geometry. There is no such requirement in the Molden format specification, like https://www.theochem.ru.nl/molden/molden_format.html . Do you have more insight why shifting the center of the molecule would solve the visualization problem?

PR #2300, opened 2024-07-01
activate more checks by ruff

fchapoton:
This is adding a few checks in the lint workflow, using `ruff`.

Some fixes are done so that the linter should be happy.


sunqm:
I'll close this PR for the moment as the style updates do not look particularly necessary to me. If you think the style checks is important, please keep posting suggestions and reasons in this issue. We can wait for more feedbacks and reopen the PR after more careful consideration.

PR #2301, opened 2024-07-04
Superposition of Atomic Potentials (SAP) initial guess implementation for HF-SCF calculations

Jondesan:
Added VSAP (superposition of atomic potentials) initial guess implementation for Hartree-Fock SCF calculations (described by S. Lehtola here: https://doi.org/10.1021/acs.jctc.8b01089)
Modified fakemol_for_charges to be usable for SAP 3C2E (three-center two-electron) integrals.
Supplied sap_grasp coefficients computed by S. Lehtola to the basis sets.


pyscf/gto/basis/__init__.py
```diff
@@ -421,6 +426,7 @@ def _is_pople_basis(basis):
 _BASIS_DIR = os.path.dirname(__file__)
 _GTH_BASIS_DIR = os.path.abspath(f'{pyscf.__file__}/../pbc/gto/basis')
 _GTH_PP_DIR = os.path.abspath(f'{_GTH_BASIS_DIR}/../pseudo')
+_SAP_BASIS_DIR = os.path.abspath(f'{_BASIS_DIR}/sap')
```
susilehtola, line 429: The files could be placed in the normal `_BASIS_DIR`

pyscf/gto/mole.py
```diff
@@ -4047,22 +4048,22 @@ def fakemol_for_charges(coords, expnt=1e16):
...
-    fakebas[:,NPRIM_OF] = 1
+    fakebas[:,NPRIM_OF] = contr_coeff.size
     fakebas[:,NCTR_OF] = 1
     if expnt.size == 1:
         expnt = expnt[0]
         # approximate point charge with gaussian distribution exp(-1e16*r^2)
         fakebas[:,PTR_EXP] = ptr
         fakebas[:,PTR_COEFF] = ptr+1
-        fakeenv.append([expnt, 1/(2*numpy.sqrt(numpy.pi)*gaussian_int(2,expnt))])
+        fakeenv.append([expnt, 1 / (2*numpy.sqrt(np.pi)*gaussian_int(2,expnt))])
```
susilehtola, line 4058: `numpy` is used elsewhere in the code

pyscf/gto/mole.py
```diff
@@ -4047,22 +4048,22 @@ def fakemol_for_charges(coords, expnt=1e16):
...
     else:
-        assert expnt.size == nbas
+        assert expnt.size == nbas or expnt.size == contr_coeff.size
         # approximate point charge with gaussian distribution exp(-expnt*r^2)
         fakebas[:,PTR_EXP] = ptr + numpy.arange(nbas) * 2
-        fakebas[:,PTR_COEFF] = ptr + numpy.arange(nbas) * 2 + 1
-        coeff = 1 / (2 * numpy.sqrt(numpy.pi) * gaussian_int(2, expnt))
-        fakeenv.append(numpy.vstack((expnt, coeff)).T.ravel())
+        fakebas[:,PTR_COEFF] = ptr + numpy.arange(nbas) * 2 + contr_coeff.size
+        coeff = contr_coeff / (2 * numpy.sqrt(numpy.pi) * gaussian_int(2, expnt))
```
susilehtola, line 4065: this only works when `expnt.size == contr_coeff.size`

pyscf/scf/hf.py
```diff
@@ -80,6 +80,8 @@ def kernel(mf, conv_tol=1e-10, conv_tol_grad=None,
             generated by the builtin function :func:`locals`, so that the
             callback function can access all local variables in the current
             environment.
+        sapgrasp : str (small | large)
+            SAP initial guess coefficient set size.
```
susilehtola, line 84: This is a bad name for the parameter controlling the potential basis, since many parametrizations have been suggested in the literature, and it can also be customized by the user.

pyscf/scf/hf.py
```diff
@@ -719,13 +721,38 @@ def fproj(mo):
         dm = lib.tag_array(dm, mo_coeff=mo_coeff[:,::-1], mo_occ=mo_occ)
     return dm
 
+def init_guess_by_vsap(mol, sapgrasp='small'):
```
susilehtola, line 724: rename `sapgrasp` to `sap_basis`

pyscf/scf/hf.py
```diff
@@ -719,13 +721,38 @@ def fproj(mo):
...
+    '''Generate initial guess density matrix using the superposition of
+    atomic potentials (SAP) guess, doi:10.1021/acs.jctc.8b01089
```
susilehtola, line 726: This implementation is based on the Gaussian fit presented in [doi:10.1063/5.0004046](https://doi.org/10.1063/5.0004046)

pyscf/scf/hf.py
```diff
@@ -719,13 +721,38 @@ def fproj(mo):
...
+    dm = make_rdm1(coeff, occ)
+    return dm
 
 def get_init_guess(mol, key='minao', **kwargs):
     '''Generate density matrix for initial guess
 
     Kwargs:
         key : str
-            One of 'minao', 'atom', 'huckel', 'hcore', '1e', 'chkfile'.
+            One of 'minao', 'atom', 'huckel', 'hcore', '1e', 'vsap', 'chkfile'.
```
susilehtola, line 755: Isn't `vsap` already used in the DFT code for the original variant?

pyscf/scf/hf.py
```diff
@@ -757,6 +784,56 @@ def level_shift(s, d, f, factor):
...
     return f*(1-factor) + f_prev*factor
 
+def make_vsap(mol, sapgrasp='small'):
+    '''Superposition of atomic potentials (SAP) potential matrix
+
+    Args:
+        mol : MoleBase object
+            molecule for which SAP is computed
+        sapgrasp : str (small | large)
+            which coefficient basis to use
```
susilehtola, line 794: same comment here

pyscf/scf/hf.py
```diff
@@ -757,6 +784,56 @@ def level_shift(s, d, f, factor):
...
+                load(('sap_grasp_' + sapgrasp), atom)[0][1:]
+            ) for atom in set(atoms)
+    }
+
+    # charge sumcheck
+    Z_eff = sum([numpy.sum(sapbas[a][:,1]) for a in atoms])
+    if numpy.abs(Z_eff + mol.nelectron) > 1e-3:
+        raise RuntimeError(
+            '\n'.join([f'SAP basis coefficients must be equal or close to total electronic charge: {Z_eff} !≃ {mol.nelectron}',
+            f'Check fails with value {numpy.abs(Z_eff + mol.nelectron)})']))
```
susilehtola, line 820: this should be a warning to `logger.warn`

pyscf/scf/hf.py
```diff
@@ -1710,6 +1787,13 @@ def init_guess_by_1e(self, mol=None):
         mo_occ = self.get_occ(mo_energy, mo_coeff)
         return self.make_rdm1(mo_coeff, mo_occ)
 
+    @lib.with_doc(init_guess_by_vsap.__doc__)
+    def init_guess_by_vsap(self, mol=None, sapgrasp='small'):
+        if mol is None: mol = self.mol
+        logger.info(self, '''Initial guess from superposition of atomic potentials,
+using procedure laid out by S. Lehtola at doi:10.1021/acs.jctc.8b01089.''')
```
susilehtola, line 1794: other reference missing

pyscf/scf/hf.py
```diff
@@ -1737,8 +1821,11 @@ def get_init_guess(self, mol=None, key='minao', **kwargs):
         elif key == 'atom':
             dm = self.init_guess_by_atom(mol)
         elif key == 'vsap' and hasattr(self, 'init_guess_by_vsap'):
-            # Only available for DFT objects
-            dm = self.init_guess_by_vsap(mol)
+            if 'sapgrasp' not in kwargs:
+                sapgrasp = 'small'
+            else:
+                sapgrasp = kwargs['sapgrasp']
```
susilehtola, line 1827: change logic to allow other potentials to be used

pyscf/scf/hf.py
Jondesan: Well, yes, but the PySCF logic seems to call the corresponding vsap function depending on whether the SCF object is of type DFT or HF. (well, it used to only call the DFT one, but I have now added a check which should take care of calling the corresponding vsap function depending on whether the object is DFT or HF).

sunqm:
@susilehtola do you think this PR good to merge after your comment is addressed?

pyscf/gto/basis/__init__.py
```diff
@@ -421,6 +426,7 @@ def _is_pople_basis(basis):
 _BASIS_DIR = os.path.dirname(__file__)
 _GTH_BASIS_DIR = os.path.abspath(f'{pyscf.__file__}/../pbc/gto/basis')
 _GTH_PP_DIR = os.path.abspath(f'{_GTH_BASIS_DIR}/../pseudo')
+_SAP_BASIS_DIR = os.path.abspath(f'{_BASIS_DIR}/')
```
susilehtola, line 429:
```suggestion
_SAP_BASIS_DIR = _BASIS_DIR
```

pyscf/gto/mole.py
```diff
@@ -4071,6 +4071,48 @@ def fakemol_for_charges(coords, expnt=1e16):
     fakemol._built = True
     return fakemol
 
+def fakemol_for_cgtf_charge(coord, expnt=1e16, contr_coeff=1):
+    '''Construct a fake Mole object that holds the charge on the given
+    coordinate (coord).  The shape of the charge can be a linear
+    combination of gaussian type functions with exponents expnt and
+    contraction coefficients contr_coeff.
```
susilehtola, line 4078:
```suggestion
    '''Constructs a "fake" Mole object that has a Gaussian charge
    distribution at the specified coordinate (coord).  The charge 
    can be given as a linear combination of Gaussians with 
    exponents expnt and contraction coefficients contr_coeff.
```

pyscf/gto/mole.py
```diff
@@ -4071,6 +4071,48 @@ def fakemol_for_charges(coords, expnt=1e16):
...
+    if expnt.size == 1:
+        expnt = expnt[0]
+        # approximate point charge with gaussian distribution
+        # contr_coeff*exp(-expnt*r^2)
+        fakebas[:,PTR_EXP] = ptr
+        fakebas[:,PTR_COEFF] = ptr+1
+        fakeenv.append([expnt, 1 / (2*numpy.sqrt(numpy.pi)*gaussian_int(2,expnt))])
+        ptr += 2
+    else:
+        assert expnt.size == nbas or expnt.size == contr_coeff.size
```
susilehtola, line 4103:
```suggestion
        assert expnt.size == contr_coeff.size
```

The `expnt.size == nbas` comes from `fakemol_for_charges` where it has a different effect: it looks like you can feed that routine a set of coordinates with a corresponding set of exponents.

susilehtola: Does this routine work correctly if you feed many coordinates? The function should place the same exponents and contraction coefficients on every one of those coordinates. In practice, one could do this to calculate the potential of H2O, for instance: place hydrogen atom potentials on the two hydrogens.

pyscf/scf/hf.py
```diff
@@ -719,13 +721,41 @@ def fproj(mo):
         dm = lib.tag_array(dm, mo_coeff=mo_coeff[:,::-1], mo_occ=mo_occ)
     return dm
 
+def init_guess_by_sap(mol, sap_basis='sapgraspsmall', **kwargs):
+    '''Generate initial guess density matrix using the superposition of
+    atomic potentials (SAP) guess, doi:10.1063/5.0004046
```
susilehtola, line 726:
```suggestion
    '''Generate initial guess density matrix from a superposition of
    atomic potentials (SAP), doi:10.1021/acs.jctc.8b01089.
    This is the Gaussian fit implementation, see doi:10.1063/5.0004046.

File: pyscf/scf/hf.py [hunk 16]
```diff
@@ -757,6 +787,60 @@ def level_shift(s, d, f, factor):
...
+    atom_coords = numpy.asarray([coord[1] for coord in mol._atom])
+    atoms = numpy.asarray([coord[0] for coord in mol._atom])
+
+    sapbas = {
+        atom: numpy.asarray(
+                load((sap_basis), atom)[0][1:]
+            ) for atom in set(atoms)
+    }
+
+    logger.note(mol, f'Found SAP basis! Using {sap_basis.split("/")[-1]}')
```
susilehtola, line 815: As I said in person, this routine should take in the SAP basis already as a dict. There are other potentials in the Basis Set Exchange, but this routine is limited to the internal basis set library in NWChem format.

pyscf/scf/hf.py
```diff
@@ -757,6 +787,60 @@ def level_shift(s, d, f, factor):
...
+
+    # charge sumcheck
+    Z_eff = sum([numpy.sum(sapbas[a][:,1]) for a in atoms])
+    if numpy.abs(Z_eff + mol.nelectron) > 1e-3:
```
susilehtola, line 819: This check could be tighter, e.g. `1e-6`

pyscf/scf/hf.py
```diff
@@ -1444,7 +1528,7 @@ class SCF(lib.StreamObject):
             be skipped and the kernel function will compute only the total
             energy based on the initial guess. Default value is 50.
         init_guess : str
-            initial guess method.  It can be one of 'minao', 'atom', 'huckel', 'hcore', '1e', 'chkfile'.
+            initial guess method.  It can be one of 'minao', 'atom', 'huckel', 'hcore', '1e', 'vsap', 'chkfile'.
```
susilehtola, line 1531: sap not vsap

pyscf/scf/hf.py
```diff
@@ -1710,6 +1794,15 @@ def init_guess_by_1e(self, mol=None):
         mo_occ = self.get_occ(mo_energy, mo_coeff)
         return self.make_rdm1(mo_coeff, mo_occ)
 
+    @lib.with_doc(init_guess_by_sap.__doc__)
+    def init_guess_by_sap(self, mol=None, sap_basis='sapgraspsmall', **kwargs):
+        if mol is None: mol = self.mol
+        logger.info(self, '''Initial guess from superposition of atomic potentials,
+using procedure laid out by S. Lehtola at 
+doi:10.1021/acs.jctc.8b01089 and
+doi:10.1063/5.0004046.''')
```
susilehtola, line 1803:
```suggestion
        logger.info(self, '''Initial guess from superposition of atomic potentials (doi:10.1021/acs.jctc.8b01089)
This is the Gaussian fit version as described in doi:10.1063/5.0004046.''')
```

pyscf/scf/hf.py
```diff
@@ -1710,6 +1794,15 @@ def init_guess_by_1e(self, mol=None):
...
+        return init_guess_by_sap(mol, sap_basis=sap_basis, **kwargs)
```
susilehtola, line 1804: This is where you should build the `sap_basis` if it is not already a PySCF dictionary

pyscf/scf/hf.py
```diff
@@ -1737,8 +1830,9 @@ def get_init_guess(self, mol=None, key='minao', **kwargs):
         elif key == 'atom':
             dm = self.init_guess_by_atom(mol)
         elif key == 'vsap' and hasattr(self, 'init_guess_by_vsap'):
-            # Only available for DFT objects
             dm = self.init_guess_by_vsap(mol)
+        elif key == 'sap' and hasattr(self, 'init_guess_by_sap'):
```
susilehtola, line 1834:
```suggestion
        elif key == 'sap':
```

susilehtola:
You also need to 
- [ ] add the guess to `scf/dhf.py`
- [ ] add the guess to `scf/ghf.py`
- [ ] add the guess to `scf/rohf.py`
- [ ] add the guess to `scf/uhf.py`
- [ ] check the guess works in all of the above
- [ ] add an example to `examples/scf/15-initial_guess.py`

pyscf/scf/hf.py
Jondesan: The load actually does look through the bse catalog if it doesn't find the basis locally.
But I see your point regarding the input argument sap_basis being a dict.

Jondesan: But there does seem to be a slight annoyance with that: if pyscf doesn't find the data locally and ends up using the bse api, load returns a dict with one key-value pair: the selected atom symbol as key and the basis of that element as value. If the data is found locally load returns only the basis array without the element symbol as key. So the load function is kinda bad in that regard but oh well, I wrote a workaround.

examples/scf/15-initial_guess.py
```diff
@@ -93,6 +93,23 @@
...
+#
+# The SAP fit basis can be changed using the kwarg sap_basis.
+# The argument accepts either python dictionary (basis set in internal
+# format) or filename/pathname. If BSE API is installed with pip, 
+# the implementation will also look through the BSE catalog for basis sets.
+#
+mf.kernel(sap_basis='sapgraspsmall') # Will use PySCF SAP_ALIAS and find it
+# in the local files
+mf.kernel(sap_basis='sap_helfem_large') # Will be found from BSE if installed
+
```
susilehtola, line 112: Add an example here for the use of a PySCF dict basis

pyscf/gto/basis/__init__.py
```diff
@@ -603,6 +609,10 @@ def load(filename_or_basisname, symb, optimize=OPTIMIZE_CONTRACTION):
         basis_dir = _GTH_BASIS_DIR
     elif _is_pople_basis(name):
         basmod = _parse_pople_basis(name, symb)
+    elif name in SAP_ALIAS:
+        basmod = SAP_ALIAS[name]
+        fload = parse_nwchem.load
+        basis_dir = _SAP_BASIS_DIR
```
susilehtola, line 615:
```suggestion
    elif name in SAP_ALIAS:
        basmod = SAP_ALIAS[name]
```

pyscf/gto/basis/__init__.py
```diff
@@ -421,6 +426,7 @@ def _is_pople_basis(basis):
 _BASIS_DIR = os.path.dirname(__file__)
 _GTH_BASIS_DIR = os.path.abspath(f'{pyscf.__file__}/../pbc/gto/basis')
 _GTH_PP_DIR = os.path.abspath(f'{_GTH_BASIS_DIR}/../pseudo')
+_SAP_BASIS_DIR = _BASIS_DIR
```
susilehtola, line 429:
```suggestion
```

examples/scf/15-initial_guess.py
```diff
@@ -86,13 +86,35 @@
...
+# sap_basis accepts either python dictionary (basis set in internal
+# format) or filename/pathname. If BSE API is installed with pip, 
+# the implementation will also look through the BSE catalog for basis sets.
+#
+mf.sap_basis = 'sapgraspsmall'
+mf.kernel() # Will use PySCF SAP_ALIAS and find it
+# in the local files
+mf.sap_basis = 'sap_helfem_large'
+mf.kernel() # Will be found from BSE if installed
+mf.sap_basis = gto.basis.load('sap_helfem_large', 'Ne')
```
susilehtola, line 114: This doesn't make sense? This should be a dict containing basis sets for every element! Also, the molecule has C not Ne...

pyscf/gto/mole.py
```diff
@@ -4071,6 +4071,48 @@ def fakemol_for_charges(coords, expnt=1e16):
...
+    fakeatm[:,PTR_COORD] = numpy.arange(ptr, ptr+nbas*3, 3)
+    fakeenv.append(coord.ravel())
+    ptr += nbas*3
+    fakebas[:,ATOM_OF] = numpy.arange(nbas)
+    fakebas[:,NPRIM_OF] = contr_coeff.size
+    fakebas[:,NCTR_OF] = 1
+    if expnt.size == 1:
+        expnt = expnt[0]
+        # approximate point charge with gaussian distribution
+        # contr_coeff*exp(-expnt*r^2)
```
susilehtola, line 4097:
```suggestion
```

pyscf/gto/mole.py
```diff
@@ -4071,6 +4071,48 @@ def fakemol_for_charges(coords, expnt=1e16):
...
+        fakebas[:,PTR_EXP] = ptr
+        fakebas[:,PTR_COEFF] = ptr+1
+        fakeenv.append([expnt, 1 / (2*numpy.sqrt(numpy.pi)*gaussian_int(2,expnt))])
+        ptr += 2
+    else:
+        assert expnt.size == contr_coeff.size
+        fakebas[:,PTR_EXP] = ptr + numpy.arange(nbas) * 2
+        fakebas[:,PTR_COEFF] = ptr + numpy.arange(nbas) * 2 + contr_coeff.size
```
susilehtola, line 4105: I am still wondering this works with `nbas>1`

pyscf/gto/mole.py
```diff
@@ -4071,6 +4071,48 @@ def fakemol_for_charges(coords, expnt=1e16):
...
+        coeff = contr_coeff / (2 * numpy.sqrt(numpy.pi) * gaussian_int(2, expnt))
+        fakeenv.append(numpy.vstack((expnt, coeff)).ravel())
```
susilehtola, line 4107: ... since this does not have dimension `nbas`

pyscf/scf/hf.py
```diff
@@ -1710,6 +1786,31 @@ def init_guess_by_1e(self, mol=None):
...
+        if isinstance(sap_basis, str):
+            atoms = [coord[0] for coord in mol._atom]
+            sapbas = dict()
+            for atom in set(atoms):
+                single_element_bs = load(sap_basis, atom)
+                if isinstance(single_element_bs, dict):
+                    sapbas[atom] = numpy.asarray(single_element_bs[atom][0][1:], dtype=float)
+                else:
+                    sapbas[atom] = numpy.asarray(single_element_bs[0][1:], dtype=float)
+            logger.note(self, f'Found SAP basis!\nUsing {sap_basis.split("/")[-1]}')
```
susilehtola, line 1805:
```suggestion
            logger.note(self, f'Found SAP basis {sap_basis.split("/")[-1]}')
```

pyscf/scf/hf.py
```diff
@@ -1710,6 +1786,31 @@ def init_guess_by_1e(self, mol=None):
...
+        elif isinstance(sap_basis, dict):
+            sapbas = dict()
+            for key in sap_basis:
+                sapbas[key] = numpy.asarray(sap_basis[key][0][1:], dtype=float)
+        else:
+            logger.error(self, 'sap_basis is of wrong datatype.')
```
susilehtola, line 1811:
```suggestion
            logger.error(self, 'sap_basis is of an unexpected datatype.')
```

pyscf/scf/hf.py
```diff
@@ -1737,8 +1838,9 @@ def get_init_guess(self, mol=None, key='minao', **kwargs):
         elif key == 'atom':
             dm = self.init_guess_by_atom(mol)
         elif key == 'vsap' and hasattr(self, 'init_guess_by_vsap'):
-            # Only available for DFT objects
```
susilehtola, line 1740:
```suggestion
        elif key == 'vsap' and hasattr(self, 'init_guess_by_vsap'):
```

susilehtola: The comment `# Only available for DFT objects` should not be removed

susilehtola:
> @susilehtola do you think this PR good to merge after your comment is addressed?

@sunqm I am happy with it now, the only question is the role of `nbas` in the integrals routines. Another question is on how the integrals should be computed.

The $N_{\rm bf}^2$  repulsive corrections to the Fock matrix over the core guess are obtained by tracing out the nuclear index in the $N_{\rm bf}^2 N_{\rm nuc}$ integrals, which are computed all at once; one could also accumulate on the fly, computing the contributions one nucleus at a time. Which is better, in your opinion? I think @Jondesan's code is easy to change either way, since his function already creates the auxiliary basis for an atom at a time; these basis sets are then just concatenated before doing the integrals.

susilehtola (approved):
LGTM , ~~see question on the use of `nbas` though~~

sunqm:
> > @susilehtola do you think this PR good to merge after your comment is addressed?
> 
> @sunqm I am happy with it now ~, the only question is the role of `nbas` in the integrals routines~.
> 
> The only question is how should the integrals be computed. The Nbf2 repulsive corrections to the Fock matrix over the core guess are obtained by tracing out the nuclear index in the Nbf2Nnuc integrals, which are computed all at once; one could also accumulate on the fly, computing the contributions one nucleus at a time. Which is better, in your opinion? I think @Jondesan's code is easy to change either way, since his function already creates the auxiliary basis for an atom at a time; these basis sets are then just concatenated before doing the integrals.

@susilehtola Thank you for the careful review of the PR! I think either way works. There is a small difference in memory consumption, but I think this should not be a problem unless the system has more than 1000 atoms. That is an unusual case that can be considered later.

sunqm (approved):

sunqm (changes requested):
Please add unit tests for the new initial guess

Jondesan:
> Please add unit tests for the new initial guess

Added unit tests for RHF, ROHF, UHF, DHF and GHF.

sunqm (approved):

PR #2314, opened 2024-07-11
(local) ISDF module in PySCF 

NingZhang1:
1. Local Interpolative separable density fitting module for PBC in PySCF together with an experimental module for THC-post-HF methods.
2. Supporting robust fitting, k-point, MPI.
3. Add new files in pyscf/lib/pbc to support ISDF
4. Changes on pyscf/lib/numpy_helper.py to remove unnecessary copy when doing dgemm. 
5. Changes on pyscf/scf/hf.py to support occ-RI-K.


chillenb:
Exciting!!!!

I wonder what can be done about the fact that the Apache 2.0 and GPL 3.0 licenses are [incompatible](https://www.apache.org/licenses/GPL-compatibility.html). That may lead to a license problem when making FFTW a required dependency, because then pyscf binaries cannot be distributed. So maybe an FFT wrapper is needed?

Some options:

- Intel MKL has an [FFTW3 interface](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2024-2/fftw3-interface-to-onemkl.html).
- Use [pocketfft](https://github.com/scipy/scipy/tree/main/scipy/fft/_pocketfft) like SciPy. PocketFFT has [both C and C++](https://github.com/mreineck/pocketfft) versions. Do they have the same performance?

tberkel:
Hi @NingZhang1 , these PRs should probably be staying in a local fork for now while you're still developing ISDF. If there are features that are finished and ready to merge, they can be separated into a standalone PR so that we don't have to review 450 commits and 150+ files changed. And if the features are very new, we may ask that the PR first go to pyscf-forge for testing before being incorporated into the main pyscf distribution.

NingZhang1:
Hi Timothy,

Sorry for updating such a large PR. Thanks for suggestions. I am preparing
a more clean branch for update, which will contain the most stable features
that have been fully tested (by myself) and ready to merge. I agree that it
is better to first merge it into pyscf-forge beforeing incorportated into
the main pyscf distribution.

Best wishes,

Ning

Timothy Berkelbach ***@***.***> 于2024年7月25日周四 10:19写道：

PR #2316, opened 2024-07-12
Use the logger instead of stderr for warnings in the pbc modules

william-dawson:
Inside the pbc modules, many times `sys.stderr.write` is used to write warnings, instead of the `warn` member function of `pyscf.lib.logger`. This makes it difficult to suppress warnings. I modified each of those spots to instead call the logger. 

In all but one place, I could pass `self` to `warn` to ensure the verbosity parameter was communicated. However, in `pyscf/pbc/dft/numint.py`, the function `eval_ao` takes instead an integer directly for the verbosity. This is why I use the `SimpleNamespace` trick there, but if you prefer another approach let me know.

### DIFF
--- a/pyscf/pbc/dft/numint.py
+++ b/pyscf/pbc/dft/numint.py
@@ -17,7 +17,6 @@
 #         Qiming Sun <osirpt.sun@gmail.com>
 #
 
-import sys
 import numpy
 from pyscf import lib
 from pyscf.dft import numint
@@ -28,6 +27,7 @@ from pyscf.dft.gen_grid import NBINS, CUTOFF, ALIGNMENT_UNIT
 from pyscf.pbc.dft.gen_grid import make_mask, BLKSIZE
 from pyscf.pbc.lib.kpts_helper import is_zero, KPT_DIFF_TOL
 from pyscf.pbc.lib.kpts import KPoints
+from pyscf.lib import logger
 
 
 def eval_ao(cell, coords, kpt=numpy.zeros(3), deriv=0, relativity=0, shls_slice=None,
@@ -75,10 +75,12 @@ def eval_ao_kpts(cell, coords, kpts=None, deriv=0, relativity=0,
         ao_kpts: (nkpts, [comp], ngrids, nao) ndarray
             AO values at each k-point
     '''
+    from types import SimpleNamespace
     if kpts is None:
         if 'kpt' in kwargs:
-            sys.stderr.write('WARN: KNumInt.eval_ao function finds keyword '
-                             'argument "kpt" and converts it to "kpts"\n')
+            logger.warn(SimpleNamespace(verbose=verbose),
+                        'KNumInt.eval_ao function finds keyword '
+                        'argument "kpt" and converts it to "kpts"')
             kpts = kwargs['kpt']
         else:
             kpts = numpy.zeros((1,3))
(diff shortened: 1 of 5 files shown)


fishjojo:
I think most of these warnings are important and are not meant to be suppressed.

william-dawson:
@fishjojo I agree, and I don't want to give the impression that I'm ignoring these warnings for production. With this modification, the warnings are still printed out by default. The change just has them be printed out in the same way as as equivalent warnings in other parts of the code.

pyscf/pbc/dft/numint.py
```diff
@@ -75,10 +75,12 @@ def eval_ao_kpts(cell, coords, kpts=None, deriv=0, relativity=0,
         ao_kpts: (nkpts, [comp], ngrids, nao) ndarray
             AO values at each k-point
     '''
+    from types import SimpleNamespace
     if kpts is None:
         if 'kpt' in kwargs:
-            sys.stderr.write('WARN: KNumInt.eval_ao function finds keyword '
-                             'argument "kpt" and converts it to "kpts"\n')
+            logger.warn(SimpleNamespace(verbose=verbose),
+                        'KNumInt.eval_ao function finds keyword '
```
sunqm, line 82: `SimpleNamespace(verbose=verbose)` fails with `NameError: name 'verbose' is not defined` . Change this to `cell`.

sunqm:
> I think most of these warnings are important and are not meant to be suppressed.

For the critical issues, how about changing them to errors instead of just warnings? Wrapping other warnings should be okay IMO.

fishjojo:
> > I think most of these warnings are important and are not meant to be suppressed.
> 
> For the critical issues, how about changing them to errors instead of just warnings? Wrapping other warnings should be okay IMO.

okay. I think we can also add an option to always print, since sometimes we don't want the calculation to fail but let the program to choose the correct setting.

pyscf/pbc/dft/numint.py
```diff
@@ -1159,8 +1160,9 @@ def nr_rks(self, cell, grids, xc_code, dms, relativity=0, hermi=1,
                kpts=None, kpts_band=None, max_memory=2000, verbose=None, **kwargs):
         if kpts is None:
             if 'kpt' in kwargs:
-                sys.stderr.write('WARN: KNumInt.nr_rks function finds keyword '
-                                 'argument "kpt" and converts it to "kpts"\n')
+                logger.warn(self, 'KNumInt.nr_rks function finds '
```
sunqm, line 1163: Please raise an error for this incorrect kwarg.

pyscf/pbc/dft/numint.py
```diff
@@ -1174,8 +1176,9 @@ def nr_uks(self, cell, grids, xc_code, dms, relativity=0, hermi=1,
                kpts=None, kpts_band=None, max_memory=2000, verbose=None, **kwargs):
         if kpts is None:
             if 'kpt' in kwargs:
-                sys.stderr.write('WARN: KNumInt.nr_uks function finds keyword '
-                                 'argument "kpt" and converts it to "kpts"\n')
+                logger.warn(self, 'KNumInt.nr_uks function finds '
```
sunqm, line 1179: Please raise an error for this incorrect kwarg.

sunqm (changes requested):

sunqm:
Other sys.stderr can be replace IMO. @fishjojo please mark any critical warnings that you believe should not be suppressed.

fishjojo:
> Other sys.stderr can be replace IMO. @fishjojo please mark any critical warnings that you believe should not be suppressed.

I agree. I don't have more comments.

PR #2325, opened 2024-07-16
clean up tmpfiles from _ERIS class

cjknight:
I/We noticed some tmp files being left behind after runs. Small change to help ensure the files are removed. Improvements welcomed.


cjknight:
Yes! #2320 works perfect. This PR can be closed in favor of that. Thanks for quick response!

PR #2342, opened 2024-07-29
Use `self.ci` if set for mcscf

matthew-hennefarth:
Suppose there is an mcscf solver as such 

```python
mol = gto.M(atom="h 0 0 0; h 1 0 0", basis="sto-3g")
mf = mol.RHF().run()
mc = mcscf.CASSCF(2,2)

mc.kernel()
```

If we call `mc.kernel` again, one would expect that the calculation would reuse both the converged MO coefficients and CI. Unfortunately this is not the case, it will only reuse the MO coefficients. This PR fixes this by reusing the `mc.ci` if `ci0` is not passed into the kernel function. 

Note, that this is also important if one uses a checkpoint file and saves the ci vector to it using the `chk_ci` flag. If one uses the provided `.update()` function as 

```python
mc.update("old_chk")
mc.kernel()
```
again, the old CI coefficients stored in `"old_chk"` would not be used.

### DIFF
--- a/pyscf/fci/direct_spin0_symm.py
+++ b/pyscf/fci/direct_spin0_symm.py
@@ -97,7 +97,8 @@ def get_init_guess(norb, nelec, nroots, hdiag, orbsym, wfnsym=0):
         ci0.append(x.ravel().view(direct_spin1.FCIvector))
 
     if len(ci0) == 0:
-        raise RuntimeError(f'Initial guess for symmetry {wfnsym} not found')
+        raise lib.exceptions.WfnSymmetryError(
+            f'Initial guess for symmetry {wfnsym} not found')
     return ci0
 
 def get_init_guess_cyl_sym(norb, nelec, nroots, hdiag, orbsym, wfnsym=0):
@@ -146,7 +147,8 @@ def get_init_guess_cyl_sym(norb, nelec, nroots, hdiag, orbsym, wfnsym=0):
             break
 
     if len(ci0) == 0:
-        raise RuntimeError(f'Initial guess for symmetry {wfnsym} not found')
+        raise lib.exceptions.WfnSymmetryError(
+            f'Initial guess for symmetry {wfnsym} not found')
     return ci0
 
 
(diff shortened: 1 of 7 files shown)


matthew-hennefarth:
Seems that the `state_average_mix` test failed since it tried to redo the energy in the natural orbital picture, but was not passing in the updated CI coefficients as well.

matthew-hennefarth:
The test fails on the `test_n2/test_wfn_sym` since the kernel is called twice, but the wfn symmetry is changed. This causes an assertion to be thrown later down the line when it is trying to determine the symmetry and finds that the CI vector no longer agrees with the target symmetry.

```python3
def test_wfnsym(self):
  mc = mcscf.CASSCF(msym, 4, (3,1))
  mc.fcisolver.wfnsym = 14
  emc = mc.mc1step()[0]
  self.assertAlmostEqual(emc, -108.74508322877787, 7)
  
  mc.wfnsym = 'A2u'
  emc = mc.mc1step()[0]
  self.assertAlmostEqual(emc, -108.69019443475308, 7)
```

A fix would be to amend the test to reset the ci vector between kernel calls:

```python3
def test_wfnsym(self):
  mc = mcscf.CASSCF(msym, 4, (3,1))
  mc.fcisolver.wfnsym = 14
  emc = mc.mc1step()[0]
  self.assertAlmostEqual(emc, -108.74508322877787, 7)
  
  mc.ci = None
  mc.wfnsym = 'A2u'
  emc = mc.mc1step()[0]
  self.assertAlmostEqual(emc, -108.69019443475308, 7)
```

However, I am not sure if the way that this test is constructed makes the most sense with regard to real-world usage. 

I think this now becomes a question of what the expected behavior should be if one sets a wfn symmetry, runs a calculation, then changes the wfn symmetry, and re-runs the calculation. Should the program crash? Should the program warn and then unset the ci vector? Should the user have to manually unset the ci vector?

matthew-hennefarth:
Since `casci` already has the code to reuse the ci vector if it is stored in the solver already, I can see that their is a test [`test_with_ci_init_guess`](https://github.com/pyscf/pyscf/blob/ca8c7c1680defdfee2380eda3af3a28d9fb375cb/pyscf/mcscf/test/test_casci.py#L155-L164) which actually recreates the CASCI solver before changing the wave function symmetry. Playing around with it, if I remove the `mc2 = mcscf.CASCI(msym, 4, (3,1))` line, the calculation raises an exception similar to the CASSCF test currently. 

I think then, this would be the most logical way to move forward and the CASSCF test which is failing should be re-written in this way.

sunqm (approved):

Issue #2289, opened 2024-06-24
Smearing electron number bug


chillenb:
Smeared RHF with odd nelec gives incorrect electron number.

```
from pyscf import gto, scf
from pyscf.scf import addons

mol = gto.Mole()
mol.atom = '''
    7      0.   0  -0.7
    7      0.   0   0.7'''
mol.basis = 'cc-pvdz'
mol.charge = -1
mol.spin = 1
mol.build()
mf = scf.RHF(mol)
mf = addons.smearing(mf, sigma=0.1)
mf.kernel()
print(mf.mol.nelectron)
print(mf.mo_occ.sum())
```

```
converged SCF energy = -42.2391465363651
15
16.0
```

PR #2356, opened 2024-08-09
Improve smearning-ROHF

sunqm:
This update is related to issue #2289 . The smearing for ROHF `get_fock` seems not providing reasonable total energy. Changing the ROHF to smearing-RHF can give reasonable results. @chillenb

### DIFF
--- a/pyscf/scf/addons.py
+++ b/pyscf/scf/addons.py
@@ -44,6 +44,18 @@ def smearing(mf, sigma=None, method=SMEARING_METHOD, mu0=None, fix_spin=False):
         return mf
 
     assert not mf.istype('KSCF')
+    if mf.istype('ROHF'):
+        # Roothaan Fock matrix does not make much sense for smearing.
+        # Restore the conventional RHF treatment.
+        from pyscf import scf
+        from pyscf import dft
+        known_class = {
+            dft.rks_symm.ROKS: dft.rks_symm.RKS,
+            dft.roks.ROKS    : dft.rks.RKS     ,
+            scf.hf_symm.ROHF : scf.hf_symm.RHF ,
+            scf.rohf.ROHF    : scf.hf.RHF      ,
+        }
+        mf = _object_without_soscf(mf, known_class)
     return lib.set_class(_SmearingSCF(mf, sigma, method, mu0, fix_spin),
                          (_SmearingSCF, mf.__class__))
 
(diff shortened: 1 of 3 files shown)


chillenb:
Thanks for looking into this! Tianyu will appreciate it

This may be a silly question, but why can't ROHF be treated the same way as UHF when it comes to smearing? (With the extra constraint that mo_coeff is the same for both spins)

chillenb:
Kinda like https://pubs.aip.org/aip/jcp/article/133/14/141102/350937/Communication-ROHF-theory-made-simple

sunqm:
> Thanks for looking into this! Tianyu will appreciate it
> 
> This may be a silly question, but why can't ROHF be treated the same way as UHF when it comes to smearing? (With the extra constraint that mo_coeff is the same for both spins)

To apply the eigenvalue solver and DIIS extrapolation, the Fock matrix for ROHF is constructed in a different way. It's not the canonical Fock matrix utilized by RHF or UHF. It's a hybrid of the Fock matrices from the two spins, as shown in the linked paper. However, in the case of fractional occupancy, the separation of core-open-virtual space cannot be clearly defined. In this case, the RHF Fock can be constructed for the fractional occupancy. It is sufficient to perform SCF iteration.

chillenb:
Thanks for the clarification!

I wonder what would happen if one treated all orbitals as "open" for ROHF smearing. I'll play around with this in order to learn more about it.

pyscf/scf/addons.py
```diff
@@ -134,9 +149,14 @@ def get_occ(self, mo_energy=None, mo_coeff=None):
         if self.fix_spin and (is_uhf or is_rohf): # spin separated fermi level
             if is_rohf: # treat rohf as uhf
                 mo_es = (mo_energy, mo_energy)
+                nocc = self.mol.nelec
```
fishjojo, line 152: maybe remove these two lines

fishjojo (approved):

sunqm:
> Thanks for the clarification!
> 
> I wonder what would happen if one treated all orbitals as "open" for ROHF smearing. I'll play around with this in order to learn more about it in order to understand why it is problematic.

The problem arises from the Roothaan effective Fock in ROHF methods.
https://github.com/pyscf/pyscf/blob/bc36e37b0471a7e4950b51682cd5e9092b8b78b1/pyscf/scf/rohf.py#L120

When ROHF with integer occupancy reaches the stationary condition, this effective Fock matrix can produce the same orbitals. However, it's not clear what's the corresponding effective Fock in smearing. The Roothaan one may not necessary meet the same stationary condition.

PR #2369, opened 2024-08-15
Fix numpy.unique inverse indices due to NumPy 2.0 API changes

sunqm:


pyscf/pbc/lib/kpts_helper.py
```diff
@@ -91,6 +91,7 @@ def unique(kpts):
             kpts.round(digits), return_index=True, return_inverse=True, axis=0)[1:3]
         idx = uniq_index.argsort()
         rank = idx.argsort()
+        uniq_inverse = uniq_inverse.ravel()
```
fishjojo, line 94: seems unnecessary to me

fishjojo (approved):

pyscf/pbc/lib/kpts_helper.py
sunqm: The inverse indices is a (n, 1) shaped array in numpy 2.0, while this is a (n,) shaped vector in numpy 1.* . This difference leads to certain `pbc.df` modules fail when using numpy-2.0. For some reason, these issues were not covered in unittests .

fishjojo: FYI, this is reverted in numpy 2.0.1 except for `axis=None`.
https://github.com/numpy/numpy/pull/26961

sunqm: I see. Let's keep this change specifically for the exceptional case of NumPy 2.0.0. Then it works for all versions.

PR #2376, opened 2024-08-21
Fix `examples/dft/31-xc_value_on_grid.py`

### LINKED ISSUE #2375: Question about HFX in examples/dft/31-xc_value_on_grid.py
Hello! Thanks for great effort for great program.

I'm trying to reproduce some examples and I found something that confuse me.

In examples/dft/31-xc_value_on_grid.py

```
mf = mol.RHF().run()
dm = mf.make_rdm1()
vj, vk = mf.get_jk(dm)
print('Analytical HFX:', .5 * np.einsum('pq,qp->', vk, dm))

>>> Analytical HFX: 17.904382108498805
```

and I also run code below to check value of HFX

```
mf = mol.RKS()
mf.xc = 'HF'
mf.run()
mf.dump_scf_summary()

>>>**** SCF Summaries ****
Total Energy =                         -75.983948498105377
Nuclear Repulsion Energy =               9.188258417746113
One-electron Energy =                 -122.967473332490258
Two-electron Coulomb Energy =           46.747457470888172
DFT Exchange-Correlation Energy =       -8.952191054249402
```

The value of HFX was 8.952191054249402, which was half of the value of 17.904382108498805 in above example codes.

I'm confused which one is right.

matthew-hennefarth:
Fix #2375


fishjojo:
This example is not runnable.
The magnetic moments are no longer returned
```
# mL, mS are the spin-magentic moment at each point
rhoL, mL = r_numint.eval_rho(mol, aoL_value, dmLL)
rhoS, mS = r_numint.eval_rho(mol, aoS_value, dmSS)
```
Will need more fixes. @sunqm

examples/dft/31-xc_value_on_grid.py
```diff
@@ -98,4 +98,4 @@
 print('Semi-numerical HFX:', exx.dot(weights))
 dm = mf.make_rdm1()
 vj, vk = mf.get_jk(dm)
-print('Analytical HFX:', .5 * np.einsum('pq,qp->', vk, dm))
+print('Analytical HFX:', .25 * np.einsum('pq,qp->', vk, dm))
```
fishjojo, line 101: please also add a minus sign

examples/dft/31-xc_value_on_grid.py
```diff
@@ -98,4 +98,4 @@
 print('Semi-numerical HFX:', exx.dot(weights))
```
fishjojo, line 98: should be
```
exx = - np.einsum('rpq,ri,rj,pj,qi->r', rpq, mo, mo, occ_orb, occ_orb)
```

matthew-hennefarth: Fixed

examples/dft/31-xc_value_on_grid.py
matthew-hennefarth: Fixed.

matthew-hennefarth:
> This example is not runnable. The magnetic moments are no longer returned

They are returned, but it seems the unpacking is done incorrectly.

> ```
> # mL, mS are the spin-magentic moment at each point
> rhoL, mL = r_numint.eval_rho(mol, aoL_value, dmLL)
> rhoS, mS = r_numint.eval_rho(mol, aoS_value, dmSS)
> ```
 The first element returned by `r_numint.eval_rho` should be the density and the last 3 are the magnetic moments. The issue is that python tries to unpack a list of length 4 into 2 elements and this fails. The following code should work.

```python
# mL, mS are the spin-magentic moment at each point
rhoL, *mL = r_numint.eval_rho(mol, aoL_value, dmLL)
rhoS, *mS = r_numint.eval_rho(mol, aoS_value, dmSS)
mL = np.array(mL)
mS = np.array(mS)
rho = rhoL + rhoS
mx, my, mz = mL + mS
```

Note, the `np.array()` is necessary since `mL` and `mS` will become python lists and the `mL + mS` will default to appending rather than vector addition.

sunqm (approved):

PR #2392, opened 2024-08-29
More efficient PBC-CDERI loading

chillenb:
More efficient method for loading k-point CDERI. The idea is really simple: skipping numpy.hstack by letting h5py load the CDERI directly into the output buffer.

### DIFF
--- a/pyscf/pbc/df/df.py
+++ b/pyscf/pbc/df/df.py
@@ -845,35 +845,33 @@ def _hstack_datasets(data_to_stack, slices=None):
     numpy.ndarray
         The stacked data, equal to numpy.hstack([dset[slices] for dset in data_to_stack])
     """
-    sliced_dims = []
-    shapes = [x.shape for x in data_to_stack]
+    # Step 1. Calculate the shape of the output array, and store it
+    # in res_shape.
+    res_shape = list(data_to_stack[0].shape)
+    dset_shapes = [x.shape for x in data_to_stack]
 
-    # Step 1. Calculate the shapes of dset[slices] for all dset in data_to_stack.
-    # These shapes are stored in sliced_dims.
-    if isinstance(slices, tuple) or isinstance(slices, list):
-        for i, t in enumerate(slices):
-            ts = numpy.s_[t]
-            if i == 1:
-                sliced_dims.append([
-                    len(range(sh[i])[ts]) for sh in shapes
-                ])
-            else:
-                # Except along axis 1, the dimensions need to be the same.
-                # Don't worry, if they aren't, an error gets raised later.
-                sliced_dims.append([
-                    len(range(shapes[0])[ts])
-                ])
-        res_shape = sliced_dims
-        ax1widths_sliced = sliced_dims[1]
-        res_shape[1] = sum(ax1widths_sliced)
-    else:
-        # If slices is not a tuple, we assume it acts on axis 0 only.
-        res_shape = list(shapes[0])
-        ax1widths_sliced = [sh[1] for sh in shapes]
-        res_shape[0] = len(range(shapes[0][0])[slices])
+    if not (isinstance(slices, tuple) or isinstance(slices, list)):
+        # If slices is not a tuple, we assume it is a single slice acting on axis 0 only.
+        slices = (slices,)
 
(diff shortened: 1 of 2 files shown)


chillenb:
I've made a basic measurement of the impact of this PR on the performance of the following KRHF calculation:

```
import numpy as np
from pyscf.pbc import gto, scf, dft
import os

cell = gto.M(
    a = np.eye(3)*3.5668,
    atom = '''C     0.      0.      0.    
              C     0.8917  0.8917  0.8917
              C     1.7834  1.7834  0.    
              C     2.6751  2.6751  0.8917
              C     1.7834  0.      1.7834
              C     2.6751  0.8917  2.6751
              C     0.      1.7834  1.7834
              C     0.8917  2.6751  2.6751''',
    basis = 'cc-pvdz',
    verbose = 5,
)

kpts = cell.make_kpts([2,2,2])
mf = cell.KRHF(kpts=kpts).density_fit()
mf.with_df._cderi_to_save = 'diamond222.h5'
if not os.path.isfile('diamond222.h5'):
    mf.with_df.build()
else:
    mf.with_df._cderi = 'diamond222.h5'
mf.kernel()
```
On a machine with an i9-9900k (8-core) processor, this calculation takes 50s (SCF wall time) normally and 40s with the changes proposed in this PR. The benefits may be smaller for other methods, but I can't think of a reason they would get slower.

Suggestions are appreciated!

pyscf/pbc/df/df.py
```diff
@@ -831,6 +831,67 @@ def shape(self):
...
+    dset_shapes = [x.shape for x in data_to_stack]
+
+    if not (isinstance(slices, tuple) or isinstance(slices, list)):
+        # If slices is not a tuple, we assume it is a single slice acting on axis 0 only.
+        slices = (slices,)
+
+    for i, t in enumerate(slices):
+        ts = numpy.s_[t]
+        if i == 1:
+            ax1widths_sliced = [len(range(sh[1])[ts]) for sh in dset_shapes]
```
sunqm, line 860: What if `ts` is a list, or a np.array?

It is not easy to understand the purpose of `range()[ts]`. Please consider to change to a more straightforward implementation.

pyscf/pbc/df/df.py
```diff
@@ -831,6 +831,67 @@ def shape(self):
...
+            # If they aren't, an error gets raised later.
+            res_shape[i] = len(range(res_shape[i])[ts])
+    if len(slices) <= 1:
+        ax1widths_sliced = [sh[1] for sh in dset_shapes]
+
+    # Final dim along axis 1 is the sum of the post-slice axis 1 widths.
+    res_shape[1] = sum(ax1widths_sliced)
+
+    # Step 2. Allocate the output buffer
+    out = numpy.empty(res_shape, dtype=numpy.result_type(*data_to_stack))
```
sunqm, line 872: Here is an implicit type conversion.  `np.result_type` requires the arguments to be arrays or dtypes. The h5dset may or may not be compatible with arrays. It's better to first extract data types for each element in `data_to_stack` then pass to `np.result_type()` .

sunqm (changes requested):

pyscf/pbc/df/df.py
chillenb: Good point!

The expression `len(range(N)[ts]` is supposed to compute the length of a length-N array after slicing it by the slice `ts`.  Calculating this manually is a little cumbersome and error-prone because of the many possible cases, and this was the shortest way I could think of. But, it should be pretty simple to write.

If `t` is a list or numpy array, then it's not going to work with `read_direct`. It might therefore be reasonable to fall back to numpy.hstack if `slices` isn't one of the following:
- `()`
- a slice
- a tuple of slices
- a list of slices

chillenb: OK, I made an attempt to address your comment. Hopefully it's a step in the right direction

pyscf/pbc/df/df.py
chillenb: Done!

sunqm (approved):

chillenb (commented):
:+1:

PR #2418, opened 2024-09-16
refactor dfmp2 & rpa

hongzhouye:
The molecular DFMP2 code is refactored as follows:
1. fixing memory error and reducing memory cost
2. auto-switching to outcore mode when Lov cannot be cached in memory
3. evaluating Lov-type integrals directly without caching AO DF integrals if the latter is not available
4. reducing CPU time for kernel function by two by using symmetry of ERIs
5. improving multithreading efficiency with C code
6. support for UHF reference
7. Old Python-implementation is revised and available in `dfmp2_slow.py` and `dfump2_slow.py`

[Update] The RPA and URPA codes in `pyscf.gw` are also updated with the new implementation of DFMP2 ao2mo.

Some benchmarks:
1. **Multithreading**. The efficiency of multithreading is benchmarked [here](https://github.com/user-attachments/files/17008378/multithreading.pdf) using a diamond cluster C51H52 in cc-pVTZ basis sets with 256 electrons and 2200 MOs. The new code achieves 16x speedup when using 24 CPU cores (8x for old code).
3. **Scaling**. Wall time using 24 CPU cores for a series of diamond clusters with increasing size is shown [here](https://github.com/user-attachments/files/17417034/twall_cmp.pdf). The new code is in general 4x faster than the old code (left panel). Also, using outcore mode results in negligible overhead (right panel). Cost of DFUMP2 is 4x that of DFRMP2 (right panel) as expected.


hongzhouye:
The CI test on `linux-build (ubuntu-20.04, 3.12)` failed with the following error message:
```
E   OSError: /home/runner/work/pyscf/pyscf/pyscf/lib/libmp.so: undefined symbol: dtrtrs_
```
The symbol `dtrtrs_` is a Lapack function, and other environments appear to locate it without issues. Any thoughts on what might be causing this or how to fix it?

pyscf/gw/rpa.py
```diff
@@ -59,126 +59,100 @@ def kernel(rpa, mo_energy, mo_coeff, cderi_ov=None, nw=40, x0=0.5, verbose=logge
...
     freqs, wts = np.polynomial.legendre.leggauss(nw)
     freqs_new = x0 * (1.0 + freqs) / (1.0 - freqs)
     wts = wts * 2.0 * x0 / (1.0 - freqs)**2
     return freqs_new, wts
 
 def _get_clenshaw_curtis_roots(nw):
-    """
-    Clenshaw-Curtis quadrature on [0,inf)
+    '''
+    Clenshaw-Curtis qaudrature on [0,inf)
```
fishjojo, line 149: typo

chillenb:
> The CI test on `linux-build (ubuntu-20.04, 3.12)` failed with the following error message:
> 
> ```
> E   OSError: /home/runner/work/pyscf/pyscf/pyscf/lib/libmp.so: undefined symbol: dtrtrs_
> ```
> 
> The symbol `dtrtrs_` is a Lapack function, and other environments appear to locate it without issues. Any thoughts on what might be causing this or how to fix it?

LAPACK isn't linked at all by PySCF. Often LAPACK comes bundled with BLAS (eg MKL and OpenBLAS) and the code will work as is, but this is not always true.

A good solution might be to substitute dtrtrs with the BLAS function dtrsm. The reference LAPACK implementation of dtrtrs is essentially a [wrapper](https://netlib.org/lapack/explore-html/d4/dc1/group__trtrs_gab0b6a7438a7eb98fe2ab28e6c4d84b21.html) for this function.

chillenb:
test probably failed due to lack of underscore in `dtrsm_`, because fortran subroutines commonly (depending on fortran compiler) have underscores appended to their symbol names

hongzhouye:
@chillenb Thank you so much for your help! I replaced it with the underscored `dtrsm_`, and it seems to be working now.

chillenb:
No problem--happy to help!

pyscf/gw/rpa.py
hongzhouye: fixed 👍

sunqm (approved):

PR #2449, opened 2024-10-08
Fix compiling issues. Remove the non-standard ffsll function

sunqm:


chillenb:
I think the problem may be due to the fact that `ffsll` is defined in `<strings.h>` which is not included. CMake thinks `ffsll` exists, but I think this header needs to be included for the code to compile

fishjojo:
When compiled with the C99 standard, the [Feature Test Macro](https://man7.org/linux/man-pages/man7/feature_test_macros.7.html) definition may be disabled. But ffsll requires to define that [macro](https://man7.org/linux/man-pages/man3/ffs.3.html). So the following may work
```
#ifdef HAVE_FFS
  #define _DEFAULT_SOURCE //may use _GNU_SOURCE for glibc<2.27
  #include <string.h>
#endif
```

sunqm:
> When compiled with the C99 standard, the [Feature Test Macro](https://man7.org/linux/man-pages/man7/feature_test_macros.7.html) definition may be disabled. But ffsll requires to define that [macro](https://man7.org/linux/man-pages/man3/ffs.3.html). So the following may work
> 
> ```
> #ifdef HAVE_FFS
>   #define _DEFAULT_SOURCE //may use _GNU_SOURCE for glibc<2.27
>   #include <string.h>
> #endif
> ```

@fishjojo I updated the code using the solution provided by #2452 . I'm not sure whether your suggested change is needed on top of that PR. If so, please change the relevant files.

sunqm:
> When compiled with the C99 standard, the [Feature Test Macro](https://man7.org/linux/man-pages/man7/feature_test_macros.7.html) definition may be disabled. But ffsll requires to define that [macro](https://man7.org/linux/man-pages/man3/ffs.3.html). So the following may work
> 
> ```
> #ifdef HAVE_FFS
>   #define _DEFAULT_SOURCE //may use _GNU_SOURCE for glibc<2.27
>   #include <string.h>
> #endif
> ```

@fishjojo I updated the code using the solution provided by #2452 . I'm not sure whether your suggested change is needed on top of that PR. If so, please change the relevant files.

fishjojo (approved):

PR #2452, opened 2024-10-08
fix check for ffsll and use builtin if available

chillenb:
This is a comment on PR #2449.

It seems that, in the conda build environment, `ffsll` is defined in glibc but not included in any headers. Replacing `check_function_exists` with `check_symbol_exists` allows this situation to be detected.

A Github Actions workflow for testing the conda-build process is also included. It's running on this PR here:
https://github.com/chillenb/pyscf/actions/runs/11243087685/job/31258188365

### DIFF
--- a/.github/workflows/ci_conda.yml
+++ b/.github/workflows/ci_conda.yml
@@ -1,0 +1,25 @@
+name: CI_conda
+
+on:
+  workflow_dispatch:
+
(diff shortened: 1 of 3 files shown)


sunqm:
Thanks for the solution for #2449. This change has been merged into that PR.

Issue #2456, opened 2024-10-10
Calculating exchange integrals fails when only_dfj=True


jonas-greiner:
Hi everyone,
the following minimal example fails due to an assertion error:
```
from pyscf import gto, dft

mol = gto.M(
    atom="""
        O  0.00000000  0.00000000  0.00000000
        H -0.75390364  0.00000000 -0.58783729
        H  0.75390364  0.00000000 -0.58783729
    """
)

mf = dft.RKS(mol).density_fit(only_dfj=True).run()
vk = mf.get_k(mol=mol)
```
It should still be possible to get the regular exchange integrals which are also returned for `get_jk` if `with_j=True` and `with_k=True`. The issue appears to be in the following function:
https://github.com/pyscf/pyscf/blob/410d9608a73ad3a507b3008a6f68549288ebfc33/pyscf/df/df_jk.py#L133-L159

I would propose the following solution:
 ```
   def get_jk(self, mol=None, dm=None, hermi=1, with_j=True, with_k=True,
               omega=None):
        assert (with_j or with_k)
        if dm is None: dm = self.make_rdm1()
        if not self.with_df:
            return super().get_jk(mol, dm, hermi, with_j, with_k, omega)

        with_dfk = with_k and not self.only_dfj
        if with_j:
            if isinstance(self, scf.ghf.GHF):
                def jkbuild(mol, dm, hermi, with_j, with_k, omega=None):
                    vj, vk = self.with_df.get_jk(dm.real, hermi, with_j, with_k,
                                                self.direct_scf_tol, omega)
                    if dm.dtype == numpy.complex128:
                        vjI, vkI = self.with_df.get_jk(dm.imag, hermi, with_j, with_k,
                                                    self.direct_scf_tol, omega)
                        if with_j:
                            vj = vj + vjI * 1j
                        if with_k:
                            vk = vk + vkI * 1j
                    return vj, vk
                vj, vk = scf.ghf.get_jk(mol, dm, hermi, with_j, with_dfk,
                                        jkbuild, omega)
            else:
                vj, vk = self.with_df.get_jk(dm, hermi, with_j, with_dfk,
                                            self.direct_scf_tol, omega)
        else:
            vj = None
        if with_k and not with_dfk:
            vk = super().get_jk(mol, dm, hermi, False, True, omega)[1]
        return vj, vk
```
[...]

PR #2457, opened 2024-10-10
Fixes calculating exchange integrals when only_dfj=True, issue #2456

jonas-greiner:
This PR fixes an assertion error in the following minimal example:
```
from pyscf import gto, dft

mol = gto.M(
    atom="""
        O  0.00000000  0.00000000  0.00000000
        H -0.75390364  0.00000000 -0.58783729
        H  0.75390364  0.00000000 -0.58783729
    """
)

mf = dft.RKS(mol).density_fit(only_dfj=True).run()
vk = mf.get_k(mol=mol)
```


pyscf/df/df_jk.py
```diff
@@ -132,28 +132,32 @@ def reset(self, mol=None):
...
+                    vj, vk = self.with_df.get_jk(dm.real, hermi, with_j, with_k,
+                                                self.direct_scf_tol, omega)
+                    if dm.dtype == numpy.complex128:
+                        vjI, vkI = self.with_df.get_jk(dm.imag, hermi, with_j, with_k,
+                                                    self.direct_scf_tol, omega)
+                        if with_j:
+                            vj = vj + vjI * 1j
+                        if with_k:
+                            vk = vk + vkI * 1j
+                    return vj, vk
```
susilehtola, line 153: this function isn't called

susilehtola: Also, I would refactor this function, since what it does is implement two code paths: one for real density matrices and another for complex. Generalize it to take in a `get_jk` function. Then it will be simple to implement all possibilities. On a quick look the code at present is a bit complicated and doesn't seem to cover all combinations?

jonas-greiner: Do you mean it is not called because of the changes in this PR? Because it is still passed to `scf.ghf.get_jk` right below.

susilehtola: sorry, nevermind - it's indeed passed as an argument to ghf's `get_jk`.

pyscf/df/df_jk.py
```diff
@@ -132,28 +132,32 @@ def reset(self, mol=None):
...
-                    vjI, vkI = self.with_df.get_jk(dm.imag, hermi, with_j, with_k,
-                                                   self.direct_scf_tol, omega)
-                    if with_j:
-                        vj = vj + vjI * 1j
-                    if with_k:
-                        vk = vk + vkI * 1j
-                return vj, vk
-            vj, vk = scf.ghf.get_jk(mol, dm, hermi, with_j, with_dfk,
-                                    jkbuild, omega)
+        if with_j or with_dfk:
```
susilehtola, line 141: I think there may problem here with the `else` clause: the call is to
```
                vj, vk = self.with_df.get_jk(dm, hermi, with_j, with_dfk,
                                            self.direct_scf_tol, omega)
```
and I wonder if this plays nicely for the various cases. I don't know the logic of the code well enough to review this TBH

jonas-greiner: There are no changes to this `if/else` structure except that this part of the code is only invoked if Coulomb integrals are calculated or density-fitted exchange integrals are available. The call to `self.with_df.get_jk()` will only cause issues AFAIK when `with_j` and `with_dfk` are both `False` which should not happen due to the added `if` clause and since either `with_j` and `with_k` are asserted to be `True`.

sunqm (approved):

PR #2459, opened 2024-10-10
Parity-violation contributions to electronic energy

juanjoaucar:
Dear @sunqm,

I have implemented the P-V contributions to the electronic energy for the punctual nucleus model. I've put the corresponding functions within the DHF class but I'm not sure if that is the best way to do it.

I know that some people was looking for this implementation in PySCF. Hope it serves to them.

Juan J. Aucar

### DIFF
--- a/pyscf/scf/dhf.py
+++ b/pyscf/scf/dhf.py
@@ -205,9 +205,13 @@ def get_ovlp(mol):
 
     s = mol.intor_symmetric('int1e_ovlp_spinor')
     t = mol.intor_symmetric('int1e_spsp_spinor')
+    u = mol.intor_symmetric("int1e_sp_spinor")
     s1e = numpy.zeros((n4c, n4c), numpy.complex128)
     s1e[:n2c,:n2c] = s
+    # Small (ket) over large (bra)
     s1e[n2c:,n2c:] = t * (.5/c)**2
+    # Large (ket) over small (bra)
+    s1e[n2c:, :n2c] = u.conj().T * (0.5 / c)
     return s1e
 
 make_rdm1 = hf.make_rdm1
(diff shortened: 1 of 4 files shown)


sunqm:
This will change the standard Dirac-Coulomb Hamiltonian results, right?

juanjoaucar:
No, this shouldn't change the Dirac-Coulomb Hamiltonian results. The PV contributions to the energy are negligible small, and they are treated only as a perturbation (they are less than 10^-10 hartrees for sure)

juanjoaucar:
Now I see that some tests are failing with this pull request. Maybe is due to the fact that I have added the LS|SS integrals to the overlap matrix in pyscf/scf/dhf.py.

juanjoaucar:
Dear @sunqm 

I've fixed some issues and I checked the results against the DIRAC code.

pyscf/scf/dhf.py
```diff
@@ -206,16 +206,22 @@ def get_hcore(mol):
...
 
-def get_ovlp(mol):
+def get_ovlp(mol, offdiagonal = False):
     n2c = mol.nao_2c()
     n4c = n2c * 2
     c = lib.param.LIGHT_SPEED
 
     s = mol.intor_symmetric('int1e_ovlp_spinor')
     t = mol.intor_symmetric('int1e_spsp_spinor')
+    u = mol.intor_symmetric("int1e_sp_spinor")
```
sunqm, line 216: 1. can be inside the `if offdiagonal` clause.
2. `get_ovlp(offdiagonal=True)` was not used anywhere. Is the offdiagonal necessary?

sunqm:
Thank you very much for contributing this feature! The implementation looks solid.

For maintenance purposes, I think this feature would be a good fit in pyscf-forge (https://github.com/pyscf/pyscf-forge), rather than adding it directly to the dhf.py module. It could work well as an independent module.

pyscf/scf/dhf.py
juanjoaucar: 1. Yes, sorry
2. Right. It will be needed for other parity-violating properties (also for energy in case of non-punctual nuclei). It will be used also in the pyECM code ([https://github.com/juanjoaucar/pyECM](https://github.com/juanjoaucar/pyECM))

sunqm:
I'll close this PR as it has been merged into pyscf-fore repo. Thanks for contributing to the pyscf repo!

This feature can be imported via `from pyscf.pv.energy import Epv_molecule` after installing pyscf-forge.

PR #2460, opened 2024-10-11
take cell.rcut into account when generating becke grids

simonganne01:
I got an error when generating Becke grids for 1D materials. I tracked is down to the rcut being to small. I suggest that we allow the cell.rcut to overwrite the estimation if defined.


pyscf/pbc/dft/gen_grid.py
```diff
@@ -157,8 +157,11 @@ def get_becke_grids(cell, atom_grid={}, radi_method=dft.radi.gauss_chebyshev,
         dimension = cell.dimension
     else:
         dimension = 3
-    rcut = pbc_eval_gto._estimate_rcut(cell)
-    Ls = pbc_eval_gto.get_lattice_Ls(cell, rcut=rcut.max())
+    if cell.rcut is None:
```
sunqm, line 160: This won't work because rcut can never be None.

sunqm:
Could you please post an issue with an example that demonstrates the problem? There might be some inaccuracies in estimating the "rcut" value during the DFT AO value generation. It's important to thoroughly investigate the cause of this underestimation rather than simply overriding the "rcut" value in a brute force way.

simonganne01:
I tried to recreate the error again, but I did not succeed. I do not know what changed. But I wil close the pull request.

simonganne01:
Okay, I encountered th error again. so because the unit cell size becomes irrelevant in two directions when working in 1D periodicity, you can place your 1D chain where you want. For example on the edge of a unit cell wich is really big in the two other directions. this will cause all the atoms to be outside the rcut sphere. this should not be possible. 
but I think I found two possible soulutions. 
1. change the following line `grids_edge = lib.cartesian_prod([[-.5, 1.]] * dimension).dot(a[:dimension])` to `grids_edge = lib.cartesian_prod([[-.5, 1.]] * 3).dot(a)` 
https://github.com/pyscf/pyscf/blob/2c7f7625fd2719b8f3e9715e40c3937cab65c06e/pyscf/pbc/gto/eval_gto.py#L237C5-L237C80
Because even if you have a periodicity in only 1 or 2 dimensions, youare still working in three dimensions. so the grid edges should be set in 3 dimensions. 
2. change the line `Ls_mask = (np.linalg.norm(grids2atm, axis=2) < rcut).any(axis=0)` to `Ls_mask = (np.linalg.norm(grids2atm[:, :, :dimension], axis=2) < rcut).any(axis=0)` in order to take into account the lower dimensionality of the periodicity. 
I personally prefer the first solution. but I am not an expert regarding this part of the code so please feel free to share what you think.

sunqm:
The `get_lattice_Ls` function is designed to generate the vectors for lattice sum. For low dimensional systems, these vectors on other directions should not be generated. The exception is dimension=2 which by default calls truncated Coulomb potential. This treatment actually works in 3-dimension. The truncated dimension is only enabled when `low_dim_ft_type` is set https://github.com/pyscf/pyscf/blob/2c7f7625fd2719b8f3e9715e40c3937cab65c06e/pyscf/pbc/gto/eval_gto.py#L194 .

In this sense, solution 2 makes more sense to me.

sunqm (approved):

PR #2497, opened 2024-11-04
Update boys.py

peggjt:
Make random orthogonal matrix


sunqm:
I'm not sure whether adding a completely random rotation to the initial guess is a good idea. The original intent was to introduce only a relatively small noise to the initial guess.

jeanwsr:
Close as this has been silent for a while.

Issue #2490, opened 2024-10-29
Becke radial integration scheme does not integrate numbers of electrons correctly


awvwgk:
From the available radial integration schemes the Becke one seems to not create a proper integration grid as the number of electrons is overestimated:

````python
from pyscf import dft, gto

mol = gto.M(atom="He")

for radi_method in (
   dft.radi.becke,
   dft.radi.treutler_ahlrichs,
   dft.radi.delley,
   dft.radi.gauss_chebyshev,
   dft.radi.mura_knowles,
):
   ks = dft.KS(mol)(xc="pbe", grids=dft.Grids(mol)(radi_method=radi_method), verbose=0)
   ks.kernel()
   rho = ks._numint.get_rho(mol, ks.make_rdm1(), ks.grids)
   nel = (rho * ks.grids.weights).sum()
   print(radi_method.__name__, nel)
````

Output:

````
becke 2.188070936036473
treutler_ahlrichs 1.9999999979072856
delley 1.9999999990380877
gauss_chebyshev 1.9999999994477295
mura_knowles 1.9999999997029547
````

cc Thijs @tvogels, Derk @derkkooi

susilehtola:
Yes, Becke's scheme was pioneering but it is not very good; the Treutler-Ahlrichs grid is probably the best one. Note that we have a collection of reproducible DFT grids in https://github.com/wavefunction91/IntegratorXX/ and a high-performance quadrature implementation thereof exists in https://github.com/wavefunction91/GauXC

awvwgk:
Thanks for the clarification. Closing this.

tvogels:
I still believe there might be a problem here. It makes sense that some integration schemes are more efficient than others--they might provide better approximations with a lower number of evaluations. But the implementation of the Becke scheme in PySCF doesn't even converge to the right values as the number of evaluation points goes to infinity.

<img width="459" alt="image" src="https://github.com/user-attachments/assets/a90f09d6-3a68-4376-ac30-f811ef0b01cd">

You can also see the bias in the following experiment, where I use the radial integration schemes from PySCF to integrate uniform spheres of varying radius.

<img src="https://github.com/user-attachments/assets/b7b6642c-1ab2-4110-a2e7-f22100da9cf6" width="500">

[...]

PR #2499, opened 2024-11-05
Fix biased implementation for the becke radi method

tvogels:
This fixed the Becke radial integration scheme as described in #2490.

### DIFF
--- a/pyscf/dft/radi.py
+++ b/pyscf/dft/radi.py
@@ -47,16 +47,24 @@ SG1RADII = numpy.array((
 def murray(n, *args, **kwargs):
     raise RuntimeError('Not implemented')
 
-# Gauss-Chebyshev of the first kind,  and the transformed interval [0,\infty)
 def becke(n, charge, *args, **kwargs):
     '''Becke, JCP 88, 2547 (1988); DOI:10.1063/1.454033'''
     if charge == 1:
         rm = BRAGG_RADII[charge]
     else:
         rm = BRAGG_RADII[charge] * .5
-    t, w = numpy.polynomial.chebyshev.chebgauss(n)
+
+    # Points and weights for Gauss-Chebyshev quadrature points of the second kind
+    # The weights are adjusted to integrate a function on the interval [-1, 1] with uniform weighting
+    # instead of weighted by sqrt(1 - t^2) = sin(i pi / (n+1)).
+    i = numpy.arange(n) + 1
+    t = numpy.cos(i * numpy.pi / (n + 1))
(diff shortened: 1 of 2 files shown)


sunqm:
Thanks for the fix. The Becke's 1988 paper states that the radial grids are Gauss-Chebyshev quadrature of the second kind.  The weights in your fix are different to the quadrature from `scipy.special.roots_chebyu` . Are there special treatments in your code?

chillenb:
The Gauss-Chebyshev rules in the NIST handbook Becke apparently used
[out2.pdf](https://github.com/user-attachments/files/17778266/out2.pdf)

are all for doing integrals with the weight function $w(x)=1/\sqrt{1-x^2}$. As @tvogels noticed, Becke must have divided out this weight function without saying anything. The resulting rule (the one in this PR's code) seems to be identical to the [Clenshaw-Curtis formula](https://en.wikipedia.org/wiki/Clenshaw%E2%80%93Curtis_quadrature#Integration_on_infinite_and_semi-infinite_intervals) mapped to $[0,\infty)$, with $L=1$. This follows from

$$\frac{1+\cos{x}}{1-\cos{x}}=\cot^2{\frac{x}{2}}.$$

I agree that `scipy.special.roots_chebyu` can be used for $t$ -- it returns the same values. But for $w$, it won't work without this trick.

tvogels:
Sorry for the delayed response.

Right. I followed the formulas from Wikipedia:

$$
x_i = \cos\left(\frac{i}{n+1}\pi\right),\quad w_i = \frac{\pi}{n+1}\sin^2\left(\frac{i}{n+1}\pi\right),
$$

but added a factor $\frac{1}{\sqrt{1-x^2}} = 1/\sin(\frac{i}{n+1}\pi)$ to the weights to compensate for the fact that this rule gives a weighed integral instead of the desired uniform one.

Becke is indeed not very clear about this in the paper, but something like this must have been implied.

Note that the same treatment appears in the Treutler-Ahlrichs rule:

https://github.com/pyscf/pyscf/blob/2c7f7625fd2719b8f3e9715e40c3937cab65c06e/pyscf/dft/radi.py#L144

https://github.com/pyscf/pyscf/blob/2c7f7625fd2719b8f3e9715e40c3937cab65c06e/pyscf/dft/radi.py#L146

which is also reported to be based on the Gauss-Chebyshev quadrature of the second kind.

sunqm (approved):

tvogels:
What’s your typical strategy for breaking changes like this?

sunqm:
If the code contains a bug, we then fix the bug and update the corresponding tests. In this case, it seems that the Becke radii code I implemented earlier might have a problem, as the integration of weights appears to be incorrect.

PR #2505, opened 2024-11-08
Add functions to read and write Cell geometry including lattice vectors

tberkel:


chillenb:
YES

sunqm:
I add code to rebuild the Cell instance after reading from string or file. Does it make sense in this case? @tberkel

sunqm (approved):

tberkel:
Yeah sure. The user might still have to specify basis set, pseudo, etc.,
and rebuild, but that’s no big deal.

PR #2515, opened 2024-11-11
some lightweight array helper functions

chillenb:
- Parallel elementwise 2D array multiplication
- Parallel matrix copy
- Parallel dset0 and zset0 to make sure an array is spread over NUMA nodes

### DIFF
--- a/pyscf/lib/numpy_helper.py
+++ b/pyscf/lib/numpy_helper.py
@@ -1241,39 +1241,44 @@ def omatcopy(a, out=None):
        ctypes.c_size_t(ld_out))
     return out
 
-def entrywise_mul(a, b):
+def entrywise_mul(a, b, out=None):
     """Entrywise multiplication of two matrices.
 
     Parameters
     ----------
     a : ndarray
     b : ndarray
-        Overwritten with a * b.
+    out : ndarray, optional
+        Output matrix. A new one is allocated if not provided.
 
     Returns
     -------
     ndarray
-        a * b (stored in b)
+        a * b
     """
     assert a.ndim == 2 and b.ndim == 2
     assert a.shape == b.shape and a.dtype == b.dtype
     lda, _, a_cshape = leading_dimension_order(a)
     ldb, _, b_cshape = leading_dimension_order(b)
-    assert a_cshape == b_cshape and a_cshape is not None
+    if out is None:
+        out = numpy.empty_like(b)
+    ld_out, _, out_cshape = leading_dimension_order(out)
+    assert a_cshape == b_cshape and b_cshape == out_cshape and a_cshape is not None
     if a.dtype == numpy.double:
         fn = _np_helper.NPomp_dmul
     elif a.dtype == numpy.complex128:
         fn = _np_helper.NPomp_zmul
     else:
-        b *= a
-        return b
+        return numpy.multiply(a, b, out=out)
     fn(ctypes.c_size_t(a_cshape[0]),
        ctypes.c_size_t(a_cshape[1]),
        a.ctypes.data_as(ctypes.c_void_p),
        ctypes.c_size_t(lda),
        b.ctypes.data_as(ctypes.c_void_p),
-       ctypes.c_size_t(ldb))
-    return b
+       ctypes.c_size_t(ldb),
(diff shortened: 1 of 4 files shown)


pyscf/lib/numpy_helper.py
```diff
@@ -1177,6 +1206,93 @@ def expm(a):
...
+        raise NotImplementedError
+    fn(ctypes.c_size_t(a_cshape[0]),
+       ctypes.c_size_t(a_cshape[1]),
+       a.ctypes.data_as(ctypes.c_void_p),
+       ctypes.c_size_t(lda),
+       out.ctypes.data_as(ctypes.c_void_p),
+       ctypes.c_size_t(ld_out))
+    return out
+
+def zeros(shape, dtype=numpy.double, order='C'):
```
sunqm, line 1244: np.zeros() should have the pages properly attached to sockets.

pyscf/lib/numpy_helper.py
```diff
@@ -1177,6 +1206,93 @@ def expm(a):
...
+    assert a.shape == b.shape and a.dtype == b.dtype
+    lda, _, a_cshape = leading_dimension_order(a)
+    ldb, _, b_cshape = leading_dimension_order(b)
+    assert a_cshape == b_cshape and a_cshape is not None
+    if a.dtype == numpy.double:
+        fn = _np_helper.NPomp_dmul
+    elif a.dtype == numpy.complex128:
+        fn = _np_helper.NPomp_zmul
+    else:
+        raise NotImplementedError
```
sunqm, line 1287: put `b *= a` here

sunqm:
The code implemented in this PR is correct. However, these optimizations don’t seem necessary unless there is evidence or specific examples showing that these operations are actual bottlenecks. I think it's better to keep these operations in NumPy.

chillenb:
Dear Qiming,

Here is a set of benchmarks showing how these functions are useful! I can't upload the actual ipynb here, but I can send it to you if you want to do experiments with it.
[numpybench.pdf](https://github.com/user-attachments/files/17788639/numpybench.pdf)

pyscf/lib/numpy_helper.py
chillenb: True. I'll get rid of this then.

pyscf/lib/numpy_helper.py
chillenb: OK!

pyscf/lib/numpy_helper.py
chillenb: After some digging, I learned that np.zeros() calls calloc(). It seems the OS does not actually decide allocate the pages until they are accessed for large enough calloc! In other words, if you write to a big array created with np.zeros() using a single thread, it will all stay on the same socket; if you use multiple, it'll be spread out.

First touch can be accomplished with the other routines, so I'm happy to remove this.

sunqm: Thanks for providing the performance benchmark.

The profiling for matmul using np.zeros vs lib.zeros is not directly comparable, since lib.zeros allocate pages out of the `timeit` , while the page allocation cost is included in the tests for np.zeros. Page allocation typically is the most time consuming step when manipulating a large trunk of memory. If you run the profiling second time, the memory allocated by np.zeros or lib.zeros should not have performance difference, as the pages allocated in both cases should be properly attached to numa sockets.

Calloc cannot utilize multiple threads to execute the zeroing operations. lib.zeros is beneficial from the parallel threads for this operation. In this sense, lib.zeros is more favorable given multiple cores become so normal these days. It can be kept here.

sunqm (approved):

pyscf/lib/numpy_helper.py
chillenb: @sunqm 

Thanks for your feedback!

I ran the profiling again, this time with warm-ups as you suggested. The results are here:
https://gist.github.com/chillenb/e3f533fa2566ca40e48598e32274eb68

The results did not really change much -- they suggest np.zeros is not spreading out pages over sockets on the test system. If you have observed the opposite behavior on your test system, then this is really puzzling! Perhaps the kernel configuration is the cause.

PR #2528, opened 2024-11-19
Adjust TDDFT amplitudes to follow the CIS convention

### LINKED ISSUE #2246: Complex conjugation bug in tdscf?
I think that the tdscf codes in PySCF return the complex conjugate of the eigenvector of the "conventional" CIS/TDHF matrix (equivalently, they calculate the left eigenvector or the eigenvector of the matrix conjugate transpose). Of course, the eigenvalues are fine, and this only affects methods with complex orbitals—including, but not limited to, periodic calculations. Assuming that this is *not* intentional, then ...

I think that the problem is in the transformation of the pseudodensity matrix to the AO basis and the reverse transformation of the Fock-like matrix back to the MO basis.

I have only been able to test this easily in periodic calculations. For `pbc/tdscf/krhf.py`, I think the following change is necessary:
```
133                     dmov[i,k] = reduce(numpy.dot, (orbo[k].conj(), dm1, orbv[k].T)).T  # changed
134                     #dmov[i,k] = reduce(numpy.dot, (orbo[k], dm1, orbv[k].conj().T))

143                     v1vo = reduce(numpy.dot, (orbv[k].conj().T, v1ao[i,k], orbo[k])).T  # changed
144                     #v1vo = reduce(numpy.dot, (orbo[k].conj().T, v1ao[i,k], orbv[k]))
```
This produces an eigenvector `x` that is the complex conjugate of the eigenvector returned without these changes. Two checks suggest that this change is correct:

- When comparing to a supercell calculation with all real orbitals, only this modified version (with k-point sampling and complex orbitals) agrees.
- This modified version agrees with the eigenvector of the more explicit periodic CIS code in `pbc/ci/kcis_rhf.py`

If this is a correct bug report, I think that the analogous change needs to be made throughout PySCF, for example [L97-99](https://github.com/pyscf/pyscf/blob/master/pyscf/tdscf/rhf.py#L97-L99) of molecular `pyscf/tdscf/rhf.py`, etc.

I welcome any corrections, thoughts, or suggestions for further testing.

This problem was found with @SylviaJoy1, and anecdotally supported by @hongzhouye in a different context.

sunqm:
Fix #2246

### DIFF
--- a/pyscf/pbc/scf/_response_functions.py
+++ b/pyscf/pbc/scf/_response_functions.py
@@ -64,6 +64,7 @@ def _gen_rhf_response(mf, mo_coeff=None, mo_occ=None,
                 if hermi == 2:
                     v1 = numpy.zeros_like(dm1)
                 else:
+                    assert kshift == 0
                     v1 = ni.nr_rks_fxc(cell, mf.grids, mf.xc, dm0, dm1, 0, hermi,
                                        rho0, vxc, fxc, kpts, max_memory=max_memory)
                 if hybrid:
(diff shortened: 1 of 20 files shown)


fishjojo (approved):

tberkel:
Looks good, thanks @sunqm .

PR #2538, opened 2024-12-02
a few rst fixes about RST213 in the doc, plus fixing one bad \[

fchapoton:
just fixing a few wrong uses of `*` in the doc and also one bad use of `\[` in a regexp

### DIFF
--- a/pyscf/pbc/df/fft_jk.py
+++ b/pyscf/pbc/df/fft_jk.py
@@ -41,7 +41,7 @@ def get_j_kpts(mydf, dm_kpts, hermi=1, kpts=np.zeros((1,3)), kpts_band=None):
         kpts : (nkpts, 3) ndarray
 
     Kwargs:
-        kpts_band : (3,) ndarray or (*,3) ndarray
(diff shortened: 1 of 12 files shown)


pyscf/pbc/df/fft_jk.py
```diff
@@ -41,7 +41,7 @@ def get_j_kpts(mydf, dm_kpts, hermi=1, kpts=np.zeros((1,3)), kpts_band=None):
         kpts : (nkpts, 3) ndarray
 
     Kwargs:
-        kpts_band : (3,) ndarray or (*,3) ndarray
+        kpts_band : ``(3,)`` ndarray or ``(*,3)`` ndarray
```
sunqm, line 44: what's the issue of `(3,)` in doc?

fchapoton: The issue is that `*` has a meaning in the rst language and should not be used like that. I chose to change both things on the same line.

sunqm (approved):

PR #2570, opened 2024-12-18
Add another _decompose_rdm1 for non-hermitian density matrices in df.grad.rhf

Warlocat:
The original _decompose_rdm1 seems to only work for hermitian density matrices. I add another decomposition based on SVD for non-hermitian cases, which is helpful for post-HF gradient implementations.

### DIFF
--- a/pyscf/df/grad/rhf.py
+++ b/pyscf/df/grad/rhf.py
@@ -94,7 +94,7 @@ def get_jk(mf_grad, mol=None, dm=None, hermi=0, with_j=True, with_k=True,
     dm_tril[:,idx] *= .5
 
     # For k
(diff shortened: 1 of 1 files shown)


sunqm:
I have the impression that hermi=0 for df.grad.get_veff is not used by any modules. Could you create a test case to evaluate this optimization?

Warlocat:
> I have the impression that hermi=0 for df.grad.get_veff is not used by any modules. Could you create a test case to evaluate this optimization?

I am implementing the pprpa gradients which has an anti-hermitian "dm" for get_jk. It will be an external library or merged into PySCF in the future. For now, do you want me to add a test to show hermi=0 and hermi=1 give the same results for hermitian densities?

sunqm (approved):

PR #2576, opened 2024-12-23
Fix TypeError when cc.frozen is nonzero numpy.integer

anj1:
The current implementation of the frozen orbital check can raise a `TypeError` when cc.frozen is set to a nonzero `numpy.integer` value. This occurs because the logic attempts to call `len()` on the `numpy.int64` value.

This PR adds a `hasattr` check to properly validate if an object supports `len()` before attempting to call it, fixing the `TypeError` while maintaining the existing behavior for all valid input types (None, 0, sequences).

Changes:
- Add `hasattr(frozen, '__len__')` check before attempting `len()`
- Maintain existing behavior for all valid input types

Test Plan:
- Tested with `cc.frozen = None  # Works as before`
- Tested with `cc.frozen = 0     # Works as before`
- Tested with `cc.frozen = []    # Works as before`
- Tested with `cc.frozen = np.int64(1)  # Now works correctly`

### DIFF
--- a/pyscf/grad/ccsd.py
+++ b/pyscf/grad/ccsd.py
@@ -33,7 +33,7 @@ from pyscf.cc import ccsd_rdm
 from pyscf.ao2mo import _ao2mo
 from pyscf.scf import cphf
 from pyscf.grad import rhf as rhf_grad
-from pyscf.grad.mp2 import _shell_prange, _index_frozen_active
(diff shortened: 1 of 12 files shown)


sunqm:
Thank you for identifying this issue!

sunqm (approved):

anj1:
Thanks for the fixes!

PR #2580, opened 2024-12-26
Multigrid v2 for non-orthogonal cells

fishjojo:


pyscf/pbc/gto/test/test_rcut.py
```diff
@@ -54,6 +54,21 @@ def test_rcut(self):
...
+        s0 = numpy.asarray(cell.pbc_intor('int1e_ovlp_sph', hermi=1, kpts=kpts))
+        cell1 = cell.copy()
+        cell1.use_loose_rcut = True
+        for i in range(3):
+            cell1.precision = prec = 1e-8 * 10**i
+            cell1.build()
+            t1 = numpy.asarray(cell1.pbc_intor('int1e_kin_sph', hermi=1, kpts=kpts))
+            s1 = numpy.asarray(cell1.pbc_intor('int1e_ovlp_sph', hermi=1, kpts=kpts))
+            print(prec, 'error = ', abs(t1-t0).max(), abs(s1-s0).max())
+            self.assertTrue(abs(t1-t0).max() < prec*2e3)
```
sunqm, line 69: Is an error ~1e3 too large? Should the loose_rcut be tested using a different metric like the value of orbital or electron density at that rcut?

pyscf/pbc/grad/krks.py
```diff
@@ -81,6 +81,8 @@ def get_vxc(ni, cell, grids, xc_code, dms, kpts, kpts_band=None, relativity=0, h
             for i in range(nset):
                 rho = make_rho(i, ao_k2[:,0], mask, xctype)
                 vxc = ni.eval_xc(xc_code, rho, 0, relativity, 1)[1]
+                if vxc is None:
```
sunqm, line 84: why can vxc be None? Is it from xctype=None or HF? should this special vxc be handled in a separated case?

pyscf/pbc/dft/test/test_multigrid2.py
```diff
@@ -62,33 +91,145 @@ def _multigrid2_energy_grad(cell, xc, spin=0):
...
+        assert abs(e1-e0) < 1e-7
+        assert abs(g-g0).max() < 1e-7
+        assert abs(g1-g0).max() < 1e-7
+
+    def test_orth_j_dft(self):
+        xc = ''
+        e0, g0 = _fftdf_energy_grad(He_orth, xc)
+        e,  g  = _multigrid2_energy_grad(He_orth, xc, 0)
+        e1, g1 = _multigrid2_energy_grad(He_orth, xc, 1)
+        assert abs(e-e0) < 1e-6
```
sunqm, line 219: If cell.precision=1e-8, 1e-6 error seems too big. I observed that in certain scenario, the reference provided by the default fftdf is not accurate enough, with errors slightly larger than the required cell.precision. Is this the case here?

pyscf/pbc/dft/test/test_multigrid2.py
```diff
@@ -62,33 +91,145 @@ def _multigrid2_energy_grad(cell, xc, spin=0):
...
+        assert abs(e1-e0) < 1e-6
+        assert abs(g-g0).max() < 1e-6
+        assert abs(g1-g0).max() < 1e-6
+
+    def test_orth_gga_dft(self):
+        xc = 'pbe, pbe'
+        e0, g0 = _fftdf_energy_grad(He_orth, xc)
+        e,  g  = _multigrid2_energy_grad(He_orth, xc, 0)
+        e1, g1 = _multigrid2_energy_grad(He_orth, xc, 1)
         assert abs(e-e0) < 1e-6
```
sunqm, line 189: Is the tolerance too loose?

pyscf/pbc/grad/krks.py
fishjojo: This happens when `mf.xc=""`, which corresponds to a J only calculation.

pyscf/pbc/gto/test/test_rcut.py
fishjojo: I relace the test by checking the AO value at `cell.rcut` smaller than `cell.precision`.

PR #2728, opened 2025-03-06
Better initial guess for smearing

hongzhouye:
When doing PBE calculations for a metal (fcc aluminum, 3x3x3 kpt), I found that the energy fluctuates significantly depending on the `sigma` parameter, regardless of using Fermi or Gaussian smearing. This is traced back to the `_smearing_optimization` function in `pyscf/scf/addons.py` ([link](https://github.com/pyscf/pyscf/blob/master/pyscf/scf/addons.py#L83)) converging to a solution with incorrect summed nocc. In the following table, the "old" column lists nocc obtained using the current `_smearing_optimization`. Multiple violations of the correct nocc (which is 702 in this case) are seen. The table is obtained using Gaussian smearing, but a very similar trend is seen for Fermi smearing.
```
sigma/eV   old      new
1.000e-01  702.000  702.000
5.000e-02  699.000  702.000
1.000e-02  699.000  702.000
5.000e-03  699.000  702.000
1.000e-03  699.000  702.000
5.000e-04  702.000  702.000
1.000e-04  702.000  702.000
```
The said function uses the "fermi energy" obtained from `_get_fermi` in the same file as initial guess for a 1D Powell search. This PR implements a function to calculate a better initial guess by brute-force search for the best mu on a grid within a window around the original fermi energy guess (default: +/- 1 eV with a step of 1 meV, i.e., searching over 2000 points in total). The new initial guess is found to be more robust as seen in the "new" column of the table above.

------
I have to lower the precision in following assertions in the test `test_rhf_smearing_nelec` in `pyscf/scf/test/test_addons.py`
```
self.assertAlmostEqual(abs(mf.mo_occ - mf_smear.mo_occ).max(), 0, 3)
self.assertAlmostEqual(e_frac, e_smear, 8)
```
(The precision was 4 and 9, respecitively). I do not quite understand the point of this particular test. It seems to check if the mo_occ and SCF energy before and after smearing remain the same. But why would one expect them to be the same given that smearing is applied?

### DIFF
--- a/pyscf/pbc/mpitools/mpi.py
+++ b/pyscf/pbc/mpitools/mpi.py
@@ -290,7 +290,6 @@ def _assert(condition):
         comm.Abort()
 
 def register_for(obj):
-    global _registry
     key = id(obj)
     # Keep track of the object in a global registry.  On slave nodes, the
     # object can be accessed from global registry.
--- a/pyscf/scf/addons.py
+++ b/pyscf/scf/addons.py
@@ -85,13 +85,32 @@ def _smearing_optimize(f_occ, mo_es, nocc, sigma):
         mo_occ = f_occ(m, mo_es, sigma)
         return (mo_occ.sum() - nocc)**2
 
-    fermi = _get_fermi(mo_es, nocc)
+    mu0 = _init_guess_mu(f_occ, mo_es, nocc, sigma)
     res = scipy.optimize.minimize(
-        nelec_cost_fn, fermi, method='Powell',
+        nelec_cost_fn, mu0, method='Powell',
         options={'xtol': 1e-5, 'ftol': 1e-5, 'maxiter': 10000})
     mu = res.x
     mo_occs = f_occ(mu, mo_es, sigma)
     return mu, mo_occs
+
+def _init_guess_mu(f_occ, mo_es, nocc, sigma, box=[-1.,1.], step=1e-3):
+    ''' Generate an initial guess for `mu` by manually minimizing the nocc error on a grid
+        around `fermi`. The grid is constructed as:
+
+            mus = fermi + numpy.arange(*box, step) / 27.211399
+
+        By default, `box = [-1, 1]` and `step = 1e-3`, which corresponds to searching
+        within a +/-1 eV window around `fermi` with a resolution of 1 meV (resulting
+        in 2000 grid points).
+
+        Return:
+            mu0 : float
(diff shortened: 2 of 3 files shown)


sunqm (approved):

sunqm:
> When doing PBE calculations for a metal (fcc aluminum, 3x3x3 kpt), I found that the energy fluctuates significantly depending on the `sigma` parameter, regardless of using Fermi or Gaussian smearing. This is traced back to the `_smearing_optimization` function in `pyscf/scf/addons.py` ([link](https://github.com/pyscf/pyscf/blob/master/pyscf/scf/addons.py#L83)) converging to a solution with incorrect summed nocc. In the following table, the "old" column lists nocc obtained using the current `_smearing_optimization`. Multiple violations of the correct nocc (which is 702 in this case) are seen. The table is obtained using Gaussian smearing, but a very similar trend is seen for Fermi smearing.
> 
> ```
> sigma/eV   old      new
> 1.000e-01  702.000  702.000
> 5.000e-02  699.000  702.000
> 1.000e-02  699.000  702.000
> 5.000e-03  699.000  702.000
> 1.000e-03  699.000  702.000
> 5.000e-04  702.000  702.000
> 1.000e-04  702.000  702.000
> ```
> 
> The said function uses the "fermi energy" obtained from `_get_fermi` in the same file as initial guess for a 1D Powell search. This PR implements a function to calculate a better initial guess by brute-force search for the best mu on a grid within a window around the original fermi energy guess (default: +/- 1 eV with a step of 1 meV, i.e., searching over 2000 points in total). The new initial guess is found to be more robust as seen in the "new" column of the table above.
> 
> I have to lower the precision in following assertions in the test `test_rhf_smearing_nelec` in `pyscf/scf/test/test_addons.py`
> 
> ```
> self.assertAlmostEqual(abs(mf.mo_occ - mf_smear.mo_occ).max(), 0, 3)
> self.assertAlmostEqual(e_frac, e_smear, 8)
> ```
> 
> (The precision was 4 and 9, respecitively). I do not quite understand the point of this particular test. It seems to check if the mo_occ and SCF energy before and after smearing remain the same. But why would one expect them to be the same given that smearing is applied?

This test checks the small sigma limit. When sigma is set to a value near to zero, the smearing SCF should converge to a solution that is identical to the frac_occ result. However, this calculation may encounter numerical uncertainty when sigma is very small. It does not guarantee to accurately produce the frac_occ result.

PR #2734, opened 2025-03-12
Changed wrong < into <=

mcocdawc:
Enables slices over the whole auxiliary basis in density fitting. (before this PR the last shell was not allowed)


sunqm (approved):

sunqm (approved):

mcocdawc:
May I ask how long it roughly takes for such a fix to end up in a stable release?

PR #2744, opened 2025-03-24
Bugfix in BSE ECP parser

wxj6000:
BSE parser was not able to handle the ECP basis set from BSE.

### DIFF
--- a/pyscf/gto/basis/__init__.py
+++ b/pyscf/gto/basis/__init__.py
@@ -315,7 +315,7 @@ ALIAS = {
     'pobtzvp'       :  'pob-tzvp.dat',
     'pobtzvpp'      :  'pob-tzvpp.dat',
     'crystalccpvdz' :  'crystal-cc-pvdz.dat',
-# ccECP 
+# ccECP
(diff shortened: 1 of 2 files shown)


pyscf/gto/basis/__init__.py
```diff
@@ -708,7 +708,11 @@ def load_ecp(filename_or_basisname, symb):
                 filename_or_basisname, elements=symb)
         except KeyError:
             raise BasisNotFoundError(filename_or_basisname)
-        return bse._ecp_basis(bse_obj)[0][symb]
+        ecp_basis = bse._ecp_basis(bse_obj)
+        if len(ecp_basis) > 0:
+            return ecp_basis[symb]
+        else:
+            return {}
```
sunqm, line 715: raise BasisNotFoundError

sunqm (approved):

pyscf/gto/basis/__init__.py
wxj6000: `ecp_basis` is supposed to be empty if there is no basis for the specific element.

PR #2771, opened 2025-04-09
fix near-symmetry issue with fromstring/tostring

kylebystrom:
The default symmetry tolerance in the `pbc` module is `1e-6`, while `fromfile` truncates the atomic coordinates to five decimal places. When symmetry is turned on, this leads to "near-symmetry issues" where the symmetrized mesh can be huge (e.g. 500000 x 500000 x 500000) and cause memory issues. I made two changes to address this:

- All the `fromstring` and `tostring` functions for isolated and periodic systems use 8 decimals for float formatting now.
- If symmetrizing `mesh` to `mesh1` results in the condition `m1size > 8 * msize and m1size > 1000 + msize`, a warning is raised suggesting that the user symmetrize the structure, increase symmetry tolerance, or turn off symmetry.

Please let me know if you would like anything done differently, want more tests, etc. I realize the condition above is a bit arbitrary; I just picked what seemed intuitively reasonable.

I have included a script below that fails with out-of-memory error in the previous version and runs smoothly in this version.
```
from pyscf.pbc import gto, scf
from pyscf.pbc.gto.cell import fromfile

a, atom = fromfile("Li.poscar")
# Uncomment in new version to trigger warning and OOM error
# atom = atom.replace("1.71965623", "1.71966")

cell = gto.Cell(
    a=a,
    atom=atom,
    basis='gth-szv',
    pseudo='gth-pade',
    verbose=4,
    space_group_symmetry=True,
    symmorphic=False,
)
cell.build()


kpts = cell.make_kpts(
    [2, 2, 2],
    space_group_symmetry=True,
    time_reversal_symmetry=True,
)
kmf = scf.KRHF(cell, kpts).density_fit()
ehf = kmf.kernel()
```

And here is `Li.poscar`:
```
Li2
1.0
   3.4393124531669552    0.0000000000000000    0.0000000000000002
  -0.0000000000000002    3.4393124531669552    0.0000000000000002
   0.0000000000000000    0.0000000000000000    3.4393124531669552
Li
2
direct
   0.0000000000000000    0.0000000000000000    0.0000000000000000 Li0+
   0.5000000000000000    0.5000000000000000    0.5000000000000000 Li0+
```

### DIFF
--- a/pyscf/gto/mole.py
+++ b/pyscf/gto/mole.py
@@ -2088,7 +2088,7 @@ def tostring(mol, format='raw'):
         for i in range(mol.natm):
             symb = mol.atom_pure_symbol(i)
             x, y, z = coords[i]
-            output.append('%-4s %14.5f %14.5f %14.5f' %
+            output.append('%-4s %17.8f %17.8f %17.8f' %
                           (symb, x, y, z))
         return '\n'.join(output)
     elif format == 'zmat':
--- a/pyscf/gto/test/test_mole.py
+++ b/pyscf/gto/test/test_mole.py
@@ -956,9 +956,9 @@ O    SP
         out1 = mol.tofile(tmpfile.name, format='xyz')
         ref = '''3
 XYZ from PySCF
-H           0.00000        1.00000        1.00000
-O           0.00000        0.00000        0.00000
-H           1.00000        1.00000        0.00000
+H           0.00000000        1.00000000        1.00000000
+O           0.00000000        0.00000000        0.00000000
+H           1.00000000        1.00000000        0.00000000
 '''
         with open(tmpfile.name, 'r') as f:
             self.assertEqual(f.read(), ref)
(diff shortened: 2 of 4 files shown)


tberkel:
Thanks @kylebystrom for this fix. What I still don't understand is why a system that is *almost* symmetric generates such a large mesh (larger than if symmetry is turned off). Any ideas @fishjojo ?

fishjojo:
@tberkel It is due to the fractional translation part of non-symmorphic space group operations. The default precision I set for matching the grid symmetry to the lattice symmetry is 1e-6, which seems too small. And the smaller the error is, (in this example, it's ~1e-6), the more mesh grid points are needed to satisfy the symmetry requirement. An easy workaround is to set the pyscf configuration keyword `pbc_symm_space_group_symprec` to a bigger number (1e-5 will have no issue for the current system), or I could increase the default value.

PR #2774, opened 2025-04-14
Add a general driver for finite difference gradients and hessian

sunqm:


examples/tools/07-finite_difference.py
```diff
@@ -0,0 +1,47 @@
...
+#
+method = mol.RHF().Gradients()
+H = finite_diff.kernel(method)
+print('Finite difference Hessian:')
+print(H)
+
+print('Analytical Hessian:')
+print(mol.RHF().run().Hessian().kernel())
+
+polish comments in a python module
```
fishjojo, line 38: delete?

pyscf/tools/finite_diff.py
```diff
@@ -0,0 +1,175 @@
...
+'''
+
+import numpy as np
+from pyscf import gto
+from pyscf import lib
+from pyscf.lib import logger
+from pyscf.grad.rhf import GradientsBase
+from pyscf.hessian.rhf import HessianBase
+
+def kernel(method, displacement=1e-2):
```
fishjojo, line 27: 1e-2 seems big

fishjojo (approved):

pyscf/tools/finite_diff.py
matthew-hennefarth: I disagree. Also, (assuming same units), this is the same default as in [OpenMolcas](https://molcas.gitlab.io/OpenMolcas/sphinx/users.guide/programs/alaska.html). In Orca it appears to be 0.005 Bohr.

sunqm: 1e-2 typically gives an error ~1e-4. A smaller displacement would require tightening the convergence tolerance of the underlying methods, at least to 1e-6. Otherwise, the errors in the underlying methods would dominate the finite-difference error. 1e-6 is already near the limit of many post-HF methods. This is because the orbital coefficients usually have errors on the order of the square root of the HF convergence threshold.

When computing the Hessian via finite differences, the gradients tend to carry the error in the orbitals, or even larger. ~1e-2 is almost the smallest you can use to get reliable Hessian.

PR #2776, opened 2025-04-17
Proposal: user-defined basis set aliases

chillenb:
If you need to work with more than a handful of custom basis sets at a time, it can be cumbersome to refer to them by filename, so many users choose to add them directly to `pyscf/gto/basis`, then edit the `ALIAS` dictionary in `pyscf/gto/basis/__init__.py`.

However, if you installed PySCF with pip/conda, or you are using a cluster-wide installation, you might prefer not to edit those files, or you might not have permission to do so.

 A simple solution, implemented here, is to read `USER_BASIS_DIR`, `USER_BASIS_ALIAS`, etc, from `__config__`. Just put your basis set files in USER_BASIS_DIR, and register their names in USER_BASIS_ALIAS. It's a minor quality of life improvement, but I think it would be a nice feature to have.
 
 Feedback, suggestions for improvement, criticisms, etc. are welcome!

### DIFF
--- a/examples/gto/04-input_basis.py
+++ b/examples/gto/04-input_basis.py
@@ -83,6 +83,25 @@ O    SP
 )
 
 #
+# You can define custom aliases for basis sets defined in files.
+# The variables USER_BASIS_DIR and USER_BASIS_ALIAS should be defined
+# in your PySCF configuration file (e.g. ~/.pyscf_conf.py) like this:
+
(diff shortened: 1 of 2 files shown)


jeanwsr:
Looks great. Could you add an example for using these features?

chillenb:
@jeanwsr I added a simple example. Thanks for the suggestion

sunqm (approved):

Issue #2427, opened 2024-09-18
JKFIT vs RIFIT for correlation methods


sunqm:
In the current implementation of density fitting methods, the same auxiliary basis sets are used for both the SCF and correlation methods. This may lead to potential accuracy and computational cost issues. Should this be improved? I can imagine two possible treatments:

* Automatically selecting RIFIT for correlation methods:
Like what was adopted in Psi4, the correlation methods (such as DFMP2, DFCCSD) automatically switch to the RIFIT basis set if it's available based on the orbital basis sets.

* Warning for absent RIFIT basis sets:
If RIFIT basis sets were not specified in the input, a warning is printed to notify the user. This warning can suggest how to enable the RIFIT basis set.

What do you think about this change? Please leave your comments.

fishjojo:
I prefer to print a warning. Automatically switching basis will add additional computation cost.

sokolov-group:
Thanks for opening this issue. I think the first option would be my preference. Most users do not read every line of the output file. Using JKFIT instead of RIFIT for the correlated methods can lead to significant losses in accuracy. We found that in some cases it can even affect convergence. Unless the user defines the basis set for SCF and correlation methods separately, automatically assigning JKFIT and RIFIT basis sets will improve the accuracy of density fitting.

tberkel:
I'm eight months late to this discussion, but I agree with @sokolov-group that the default behavior should automatically select RIFIT for correlated calculations. We can have an option to re-use the SCF auxiliary basis set for experts who want to save on cost and know what they're doing. This would break a lot of examples though, so maybe worth considering for PySCF 3.0?

PR #2796, opened 2025-05-02
ECP and auxiliary basis set auto assignment

sunqm:
This PR improves basis set handling for ECP validation and auxiliary basis set assignment.
* The Mole.build() checks whether ECP is required and correctly assigned for the specified elements and basis sets.
* The density fitting module now automatically assign a corresponding auxiliary basis set based on recommendations from Psi4's predefined lists or from the BSE recommendations.


tberkel:
I like the idea of better default assignments. And maybe this is better for a Discussion thread, but PySCF currently has the somewhat atypical behavior that it will re-use the DF from the mean-field (e.g., -JKFIT) in subsequent correlated calculations, rather than switching to -RI type auxiliary basis sets. Is this the default behavior we want?

jeanwsr:
> Is this the default behavior we want?

Discussed in #2427 ?

tberkel:
Yes, thanks, I missed that thread. I'll add thoughts there.

fishjojo (approved):

sunqm:
> I like the idea of better default assignments. And maybe this is better for a Discussion thread, but PySCF currently has the somewhat atypical behavior that it will re-use the DF from the mean-field (e.g., -JKFIT) in subsequent correlated calculations, rather than switching to -RI type auxiliary basis sets. Is this the default behavior we want?

This update generally does not change auxiliary basis set assignments. For orbital basis sets that were previously missing from the preferred combinations, it now queries the BSE database for recommendations. The main update here is adding more checks and improved warning messages.

Issue #2805, opened 2025-05-13
[FEATURE REQUEST] Calculate absolute overlap integrals.


mcocdawc:
**Describe the feature you'd like to see implemented**
It would be great to be able to compute the absolute overlap matrix, i.e.
$S_{ij} = \int | \phi_i(x) | | \phi_j(x) | dx$
in pyscf.

**Is your feature request related to a scientific problem? Please describe.**
I would like to use it for screening away negligible exchange-interactions between AO pairs.

**Describe alternatives you've considered**
I tried to use the normal overlap matrix and artificially set all AOs that reside on the same atom to have non-zero overlap. This did not reliably work.

On the other hand, an evaluation on a grid via the following function works, but is very slow. (Here I want to acknowledge Diptarka Haid for the suggested code.)
```python3
def calculate_abs_overlap(mol: Mole, grid_level: int = 0) -> Matrix[np.float64]:
    r"""
    Calculates the overlap matrix S_ij = \int |phi_i(r)| |phi_j(r)| dr
    using numerical integration on a DFT grid.
    Args:
        mol (gto.Mole): PySCF Mole object.
    Returns:
        numpy.ndarray: The modified overlap matrix.
    """
    grids = dft.gen_grid.Grids(mol)
    grids.level = grid_level
    grids.build()
    AO_abs_val = np.abs(dft.numint.eval_ao(mol, grids.coords, deriv=0))
    return (AO_abs_val * grids.weights[:, np.newaxis]).T @ AO_abs_val
```

susilehtola:
Overlap integrals are usually very easy to do: you use the Cartesian representation, and the integrals factorize in $x$, $y$, and $z$. At the end, you do the spherical transform to end up in the spherical basis.

The absolute value makes this impossible. You should still be able to determine analytical equations for the Cartesian integrals, since $|\phi(x,y,z)| = |\phi(x)| |\phi(y)| |\phi(z)|$. In each direction, you will then just have to find the roots $\phi(x)=0$ in the polynomial part, and divide the regions of integration so that everything is non-negative. Easy peasy. The problem is that you can't transform these integrals in the spherical basis.

A better alternative might be to just do $\phi_i^2 \phi_j^2$, instead.

mcocdawc:
Thank you for your detailed answer.

I am absolutely happy with using
$\int  |\phi_i(x)|^2  |\phi_j(x)|^2 \mathrm{d}x $
instead, but does an implementation using analytical expressions already exist for that?

[...]

PR #2825, opened 2025-05-21
Add an example for #2805

sunqm:


examples/gto/40-mole_api_and_numba_jit.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env python
+
+Polish docstring:
+'''
+This example demonstrates some APIs of the Mole class and utilizes Numba JIT
+compilation to compute integrals. The evaluated integrals are the ovelap of
```
h-larsson, line 6:
```suggestion
compilation to compute integrals. The evaluated integrals are the overlap of
```

examples/gto/40-mole_api_and_numba_jit.py
```diff
@@ -0,0 +1,109 @@
...
+absolute value of Cartesian GTOs.
+
+The overlap integral between the absolute values of Cartesian GTOs can be
+computed as the square root of the overlap of four orbitals: sqrt(<ii|jj>) .
```
susilehtola, line 10: The overlap of the squares is not the same as overlap of the absolute values.

PR #2843, opened 2025-06-10
Solvent model cleanup

sunqm:
* Fix the .TDA(), .TDDFT(), .MP2, etc methods for solvent mean-field methods.
* Update PCM TDDFT to align with the LR-PCM in QChem. Related to https://github.com/pyscf/gpu4pyscf/pull/410
* Update TDDFT interface to incorporate special treatment in solvent model.


yizeyi18:
`make_hess_object` in `pyscf.solvent.hessian.smd` wan deleted in this PR, and the `Hessian` method from `pyscf.solvent.smd` is also  removed. Are there succeedor for smd's hessian method planned?

PR #2849, opened 2025-06-12
fix input atom with zmat file 

jeanwsr:


jeanwsr:
The current `gto.M(atom='xx.zmat')` does not work correctly, since the `from_zmatrix` is called twice in the `fromstring` and `format_atom` function. This PR fixes it by making `fromstring` returns a string for zmat input (do nothing), like xyz input.

However this changes the behavior of `fromstring` function so I'm not sure whether this is correct solution.

sunqm:
Yes, the fromstring function does not operate as it was designed. I noticed this earlier as well. Your fix can be used as a workaround for the moment

sunqm (approved):

PR #2883, opened 2025-07-03
fix multigrid dft gradient for no pseudo potential

### LINKED ISSUE #2882: [BUG REPORT] Nan in multigrid gradients
**Describe the bug**
Multigrid function for gradients outputs Nan

**To Reproduce**
```
import numpy as np
import pyscf
from pyscf.pbc.dft.multigrid import MultiGridFFTDF2, pp
cell = pyscf.M(
    atom = 'H 0 0 0; H 1 1 1',
    a = np.eye(3) * 2,
    basis = [[0, [1.3, 1]],
             [1, [0.8, 1]]
            ],
    unit = 'bohr',
    ke_cutoff = 20)
mf = cell.RKS(xc='lda,vwn')
mf.with_df = MultiGridFFTDF2(cell).build()
mf.run()
dm = mf.make_rdm1()
print(pp.vpploc_part1_nuc_grad(mf.with_df, dm))
```

**Expected behavior**
```
[[-0. -0. -0.]
 [-0. -0. -0.]]
```

- commit Id 142756584

fishjojo:
Fixes #2882

### DIFF
--- a/pyscf/lib/dft/grid_common.c
+++ b/pyscf/lib/dft/grid_common.c
@@ -30,6 +30,7 @@
 #if defined(HAVE_LIBXSMM)
     #include "libxsmm.h"
 #endif
+#define EXPMIN          -700
 
 int get_lmax(int ish0, int ish1, int* bas)
 {
@@ -231,6 +232,9 @@ static int _orth_components(double *xs_exp, int* bounds, double dx, double radiu
     double base_x = dx * xij_latt;
     double x0xij = base_x - xij;
     double _x0x0 = -aij * x0xij * x0xij;
+    if (_x0x0 < EXPMIN) {
+        return 0;
+    }
 
     double *gridx = cache;
     double *xs_all = xs_exp;
@@ -242,8 +246,7 @@ static int _orth_components(double *xs_exp, int* bounds, double dx, double radiu
     double _x0dx = -2 * aij * x0xij * dx;
     double exp_dxdx = exp(_dxdx);
     double exp_2dxdx = exp_dxdx * exp_dxdx;
-    double exp_x0dx_cache = exp(_x0dx);
-    double exp_x0dx = exp_dxdx * exp_x0dx_cache;
+    double exp_x0dx = exp(_dxdx + _x0dx);
     double exp_x0x0_cache = exp(_x0x0);
     double exp_x0x0 = exp_x0x0_cache;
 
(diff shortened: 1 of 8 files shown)


pyscf/lib/dft/grid_common.c
```diff
@@ -255,7 +255,7 @@ static int _orth_components(double *xs_exp, int* bounds, double dx, double radiu
         exp_x0dx *= exp_2dxdx;
     }
 
-    exp_x0dx = exp_dxdx / exp_x0dx_cache;
```
fishjojo, line 258: I did this to reduce some FLOPs, and it should be fine unless we have a very big exponent. I guess we shouldn't use this function for point charges. @sunqm what do you think?

sunqm: The optimization for flop here is unsafe. It's not just an issue with point charges. exp_x0dx_cache can be zero or infinity for very compact basis functions. The two cases both lead to nan. To avoid this issue, in my early implementation, I explicitly removed the cases where exp_x0x0 is ~0. Additionally, instead of computing these terms as products or divisions, I directly compute both exp(x0x0 + x0dx) and exp(x0x0 - x0dx).
https://github.com/pyscf/pyscf/blob/265b98be4dee8fd59fb3d183d7a99d0b7738ef26/pyscf/lib/dft/numint_uniform_grid.c#L314

pyscf/pbc/dft/multigrid/multigrid_pair.py
```diff
@@ -1038,10 +1038,11 @@ def get_nuc_nuc_grad(mydf, dm, kpts=None):
     charge = -cell.atom_charges()
     grad = np.zeros((cell.natm,3))
     # TODO improve performance
+    rhoG = _eval_rhoG(mydf, dms, hermi=1, kpts=kpts).ravel()
```
sunqm, line 1041: Should this be computed as `_eval_rhoG(mydf, dms.conj(), hermi=1, kpts=kpts).ravel().conj()`? The `_get_j_pass2` applies a IFFT on vG while the _eval_rhoG applies FFT on rhoR.  Here, to compute IFFT(rhoR) = conj(IFFT(rhoR.conj()))

fishjojo: It should be the same. If `rhoG = _eval_rhoG().conj()`, then the following line will be `vG = -1j * exp(-1j * dot(Gv, R)) * charge * coulG * Gv`

PR #2892, opened 2025-07-10
Fix S_abs example

mcocdawc:
In #2825 an example to compute the absolute overlap matrix 

$$ S_{ij} = \int | \phi_i(r)||\phi_j(r)| \mathrm{d}r $$

was added. Unfortunately it still missed the absolute values in the inner loop and produced wrong results already for the H2 dimer, e.g. the absolute overlap between the 1s and 2p functions on the same atom was zero.

The loop to be changed was this here:
```python3
                    for n in range(nroots):
                        w = weights[n]
                        Ix += abs(mu[ix, 0, n] * nu[jx, 0, n] * w)
                        Iy += abs(mu[iy, 1, n] * nu[jy, 1, n] * w)
                        Iz += abs(mu[iz, 2, n] * nu[jz, 2, n] * w)
```

In addition, this PR is written in a more vectorized manner and uses parallelization where possible.
This gives a speedup by a factor of four for non-trivial molecules.

(e.g. 160 carbon polyacetylene chain with cc-pvdz goes down from 40s to 10s.)

### DIFF
--- a/examples/1-advanced/40-mole_api_and_numba_jit.py
+++ b/examples/1-advanced/40-mole_api_and_numba_jit.py
@@ -110,7 +110,7 @@ def primitive_overlap(li, lj, ai, aj, ci, cj, Ra, Rb, roots, weights) -> np.ndar
 @njit(cache=True, parallel=True)
 def primitive_overlap_matrix(ls, exps, norm_coef, bas_coords, roots, weights):
     nbas = len(ls)
-    dims = [(ls[k] + 1) * (ls[k] + 2) // 2 for k in range(nbas)]
+    dims = [(l + 1) * (l + 2) // 2 for l in ls]
     nao = sum(dims)
     smat = np.zeros((nao, nao))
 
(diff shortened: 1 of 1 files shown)


sunqm (approved):

examples/1-advanced/40-mole_api_and_numba_jit.py
```diff
@@ -17,114 +17,124 @@
...
+                    Iz = 0.0
                     for n in range(nroots):
-                        Ix += mu[ix,0,n] * nu[jx,0,n] * weights[n]
-                        Iy += mu[iy,1,n] * nu[jy,1,n] * weights[n]
-                        Iz += mu[iz,2,n] * nu[jz,2,n] * weights[n]
-                    s[i,j] = Ix * Iy * Iz * norm_fac
+                        w = weights[n]
+                        Ix += abs(mu[ix, 0, n] * nu[jx, 0, n] * w)
+                        Iy += abs(mu[iy, 1, n] * nu[jy, 1, n] * w)
+                        Iz += abs(mu[iz, 2, n] * nu[jz, 2, n] * w)
```
susilehtola, line 103: You are assuming that the weight is positive here, which is not the case for some of the Lebedev rules. This code will give the wrong result.

PR #2893, opened 2025-07-11
Molecular ADC for electronic excitations (EE-ADC)

sokolov-group:
This PR contributes an implementation of algebraic diagrammatic construction theory for electronic excitations (EE-ADC) written by Terrence Stahl (@stahl225). 

- Features:
    - Restricted ADC(2), ADC(2)-X, and ADC(3) for closed-shell molecules (singlet excitation energies only)
    - Unrestricted ADC(2), ADC(2)-X, and ADC(3) for open-shell molecules (excitation energies, oscillator strengths, one- and two-body density matrices for the ground and excited states, expectation values of spin-square operator)
    - In-core, out-of-core, and density fitting handling of two-electron integrals
    - EE-ADC tests and updated examples


pyscf/adc/radc_ee.py
```diff
@@ -0,0 +1,1262 @@
...
+
+
+    if (adc.method == "adc(3)"):
+        t2_ccee = t2[1][:]
+
+        if isinstance(eris.ovvv, type(None)):
+            chnk_size = radc_ao2mo.calculate_chunk_size(adc)
+            a = 0
+            for p in range(0,nocc,chnk_size):
+                v_ceee = dfadc.get_ovvv_df(adc, eris.Lov, eris.Lvv, p, chnk_size).reshape(-1,nvir,nvir,nvir)
```
sunqm, line 147: Does ADC(3) currently support only the density fitting integrals? If so, raise an error in ao2mo transformation functions. or the RADCEE instance initialization

pyscf/adc/radc_ee.py
```diff
@@ -0,0 +1,1262 @@
...
+                k = v_ceee.shape[0]
```
sunqm, line 148: Is it correct `a == p, k == min(p+chnk_size, nocc)`? If so, the slicing bounds can be generated by the `for a, b in lib.prange(0, nocc, chnk_size)`

pyscf/adc/radc_ee.py
```diff
@@ -0,0 +1,1262 @@
...
+        max_space : int
+            Space size to hold trial vectors for Davidson iterative
+            diagonalization.  Default is 12.
+
+    Kwargs:
+        nroots : int
+            Number of roots (eigenvalues) requested. Default value is 1.
+
+            >>> myadc = adc.RADC(mf).run()
+            >>> myadcea = adc.RADC(myadc).run()
```
sunqm, line 1179: adc.RADCEE(myadc).run() . The docstring in radc_ea.py and radc_ip.py needs to be updated as well.

sunqm (approved):

ssh2:
The RADC code is for singlets only (spin-conserved). The UADC code is general (no restriction on spin).

pyscf/adc/radc_ee.py
ssh2: Yes, this is correct, thanks for this suggestion. I am going through the code to address this.

pyscf/adc/radc_ee.py
ssh2: Thanks, working on address this too.

ssh2:
Addressed @sunqm's comments. We also did a lot of code clean up, added more new tests and examples, improved formatting of some of the output. Should be ready for merge but I welcome any further comments.

PR #2997, opened 2025-10-12
Using relative tolerance to check space group symmetry

### LINKED ISSUE #2996: [BUG REPORT] Unreasonable cell.mesh when space group symmetry is enabled
**Describe the bug**
Building cell with space group symmetry causes huge mesh and memory consumption

**To Reproduce**
```

cell = gto.Cell() 
cell.atom = "He 0 0 0; He 1 1 1" 
cell.basis = "6-31g" 
cell.spin = 2 
cell.a = np.eye(3) * 3 
cell.verbose = 0 
cell.precision = 1e-8
cell.space_group_symmetry = True
cell.build() 
print(cell.mesh)
```

**Expected behavior**
The reasonable mesh should be around [150, 150, 150]

**Error message**
```
mesh [139 139 139] is symmetrized as [100000 100000 100000]
WARNING!
  Symmetrization significantly increased the mesh size,
  from [139 139 139] to [100000 100000 100000]. This might indicate a nearly symmetric input
  structure, and it might cause memory issues. Consider symmetrizing your
  structure, increasing the symmetry tolerance `pbc_symm_space_group_symprec`,
  or turning off symmetry.
```

PySCF 2.10

sunqm:
fix #2996

### DIFF
--- a/pyscf/pbc/symm/symmetry.py
+++ b/pyscf/pbc/symm/symmetry.py
@@ -97,13 +97,14 @@ def check_mesh_symmetry(cell, ops, mesh=None, tol=SYMPREC,
                         return_mesh=False):
     if mesh is None:
         mesh = cell.mesh
+    mesh = np.asarray(mesh)
     ft = []
     rm_list = []
     for i, op in enumerate(ops):
         if not op.trans_is_zero:
             ft.append(op.trans)
-            tmp = op.trans * np.asarray(mesh)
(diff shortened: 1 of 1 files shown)


pyscf/pbc/symm/symmetry.py
fishjojo: should also divide `mesh` here to be consistent.

PR #3014, opened 2025-10-20
Adjust analyze() functions for UHF and ROHF to be compatible with msym

bogdanoff:
I did not notice that there are different `analyze()` functions for ROHF and UHF.
There is quite some code duplication in those, especially confusing in `hf_symm.py`, where RHF uses the function from outside, while ROHF defines its own.

I also tested cmake to build msym, which can in principle be added to `pyscf/lib/CMakeLists.txt`.
```
option(BUILD_MSYM "Download and build libmsym library" OFF)
if(BUILD_MSYM)
   find_package(PythonInterp REQUIRED)
    # message(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
#  if(NOT EXISTS "${PROJECT_SOURCE_DIR}/deps/include/libmsym/msym.h")
    ExternalProject_Add(libmsym
      GIT_REPOSITORY https://github.com/mcodev31/libmsym.git
      GIT_TAG master
      CMAKE_ARGS -DMSYM_BUILD_PYTHON:BOOL=ON -DBUILD_SHARED_LIBS:BOOL=ON
            -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR> -DMSYM_BUILD_PYTHON=ON
            -DPYTHON=${PYTHON_EXECUTABLE} -DMSYM_PYTHON_INSTALL_OPTS="--prefix <INSTALL_DIR>"
    )
#  endif()
endif()
```

There is, however, a python wrapper that has to be copied somewhere to `PYTHONPATH`, but it is not clear to me how to do it properly.

### DIFF
--- a/pyscf/scf/hf_symm.py
+++ b/pyscf/scf/hf_symm.py
@@ -838,12 +838,12 @@ class SymAdaptedROHF(rohf.ROHF):
                 ir_in_d2h = mol.irrep_id[k] % 10  # convert to D2h irreps
                 wfnsym ^= ir_in_d2h
 
-            if mol.groupname in ('SO3', 'Dooh', 'Coov'):
-                # TODO: check wave function symmetry
-                log.note('Wave-function symmetry = %s', mol.groupname)
-            else:
+            if mol.groupname in symm.param.POINTGROUP:
                 log.note('Wave-function symmetry = %s',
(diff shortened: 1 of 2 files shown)


bogdanoff:
Indeed, that is an easier option. I was just not aware of the PyPI package available.

PR #3047, opened 2025-11-15
In pseudopotential nonlocal term gradient, if large imaginary is found, print warning instead of raise error

henryw7:


YuzeHao2023 (approved):
lgtm

Issue #2591, opened 2025-01-06
GHF convergence issues


mc-rohan:
I am trying to calculate magnetization of a Lithium trimer using GHF w/in PySCF. The ground state of this system should be a spin-frustrated state, with a net magnetization of 0,0,0. This paper correctly obtained that ground state (https://doi.org/10.1063/1.4902884), shown below. This was using 3-21G with GHF (in a development version of Gaussian)
![Screenshot 2025-01-06 133008](https://github.com/user-attachments/assets/78fd9037-d897-4035-b1d4-fb81500ee24e)

When I run an analogous calculation in PySCF, the calculation converges on this ground state.
![Screenshot 2025-01-06 131849](https://github.com/user-attachments/assets/94c8e5c2-0417-4218-9bb2-e5d34c23a01d)

After using the ghf_stability function from scf.stability, I realized this was unstable. When I run stability analysis to find a stable minima, the GHF then converges on this ground state.
![Screenshot 2025-01-06 134253](https://github.com/user-attachments/assets/bae9763b-7627-4fc1-a241-10db5eaec998)

This is closer to the true ground state (net magnetization is 0,0,0), however it still doesn't quite match what was published in https://doi.org/10.1063/1.4902884. 

What could explain this discrepancy?
Thank you!

tberkel:
You've probably only a found a local minimum (stability analysis only moves you off of saddle points). You can try modifying the initial guess to find the published state.

mc-rohan:
Is there a way to obtain a complex-valued initial guess density matrix? 

All of the available initial guess methods in PySCF do not converge to the published state, and I suspect it may be because they all return dms of type np.float64. 

I see that there is a way to perform "complex GHF" by simply changing the initial guess to dtype=np.complex128 (and maybe adding some small imaginary perturbations), however this still doesn't seem enough to converge onto the published state.

tberkel:
Sorry, I don't have enough experience but maybe someone else can comment.

xubwa:
You may achieve this uniformly or randomly add a small imaginary part to a current real valued densitry matrix.

jeanwsr:
With complex initial guess (like @xubwa suggested) I can get a few complex solutions, with non-zero mx and my population on atoms, although still not equals the value in reference.

However, the ghf_stability function may not be able to detect complex GHF internal instability or real2complex instability, I believe. So it still remains a question how we can tell the complex solution is right.

jeanwsr:
[...]

PR #3057, opened 2025-11-21
Change default breaksym to False in _from_rhf_init_dm to improve GHF convergence

puzhichen:
This PR changes the default value of the `breaksym` argument in the `_from_rhf_init_dm` function (in `pyscf/scf/ghf.py`) from True to False.

Currently, when initializing a GHF density matrix from an RHF density matrix, the function defaults to `breaksym=True`, which introduces a perturbation to the off-diagonal blocks ($\alpha$-$\beta$ coupling). While testing, I observed that for certain closed-shell systems, this forced symmetry breaking in the initial guess can lead to SCF convergence issues or instabilities. In these cases, starting from a clean RHF guess (without spin-mixing) provides a more stable starting point.

### DIFF
--- a/pyscf/scf/ghf.py
+++ b/pyscf/scf/ghf.py
@@ -553,7 +553,7 @@ employing the updated GWH rule from doi:10.1021/ja00480a005.''')
 
     to_gpu = lib.to_gpu
 
-def _from_rhf_init_dm(dm, breaksym=True):
+def _from_rhf_init_dm(dm, breaksym=False):
     dma = dm * .5
     dm = scipy.linalg.block_diag(dma, dma)
     if breaksym:


jeanwsr:
Could you post some example which shows `breaksym=True` is problematic?

jeanwsr:
Your change is breaking quite a few tests. Are the results from new guess more correct or not correct in these cases?

The brokensym initial guess for GHF is quite tricky when brokensym is really needed. There're some discussions in #3058, #2591, etc. And your issue is another problem that brokensym may be a trouble when brokensym is not needed.

In the "no-brokensym" condition I think using `mol.RHF().run().to_ghf().run()`, or something similar with UHF, can avoid the GHF initial guess problem. And this makes sense since it indicates that the user has already known there's no brokensym, while `mol.GHF()` sounds like the user does not know whether there's brokensym and let it decided by program. 

However, it still remains a question whether the default behavior of `mol.GHF()` should be brokensym or not. Setting no brokensym as default may cause troubles when intrinsic brokensym exists.
I think it depends on whether the new guess can pass the tests (or fix the tests if previous tests are wrong).

puzhichen:
> Your change is breaking quite a few tests. Are the results from new guess more correct or not correct in these cases?
> 
> The brokensym initial guess for GHF is quite tricky when brokensym is really needed. There're some discussions in #3058, #2591, etc. And your issue is another problem that brokensym may be a trouble when brokensym is not needed.
> 
> In the "no-brokensym" condition I think using `mol.RHF().run().to_ghf().run()`, or something similar with UHF, can avoid the GHF initial guess problem. And this makes sense since it indicates that the user has already known there's no brokensym, while `mol.GHF()` sounds like the user does not know whether there's brokensym and let it decided by program.
> 
> However, it still remains a question whether the default behavior of `mol.GHF()` should be brokensym or not. Setting no brokensym as default may cause troubles when intrinsic brokensym exists. I think it depends on whether the new guess can pass the tests (or fix the tests if previous tests are wrong).

Thanks for the feedback.

You are right, I noticed that this commit broke a number of unit tests. I am currently investigating the root causes of these failures. Apart from the tests that verify the initial guess itself, some errors seem to stem from deviations in the subsequent calculations caused by the change in the guess. This certainly highlights the sensitivity of the results to the initial guess.

I also agree with your point that hard-coding `brokensym=False` as the default might be too arbitrary, as it could suppress intrinsic symmetry breaking when it is actually needed. Instead of a hard switch, I think exposing an interface or argument (e.g., passing a parameter to control the symmetry breaking behavior) might be a more reasonable approach.

I will update the PR once I verify whether the currently "broken" tests are producing physically incorrect results or just different ones.

PR #3058, opened 2025-11-21
GHF stability_slow

NathanGillispie:
Adding GHF stability to stability_slow. The Hessian is constructed through A, B matrices from get_ab function in [tdscf/ghf.py:109](https://github.com/pyscf/pyscf/blob/77266394415c17cab0c6a678d4bb1ed85c7929d9/pyscf/tdscf/ghf.py#L109-L172). Despite having GKS code there, I'm not comfortable with the theory.

Compared to `ghf_stability`, this will occasionally find lower energy solutions, or stabilize in a fewer number of iterations. On complex orbitals, it seems to work, but I could not find cases where `ghf_stability` didn't. So I cannot conclude that I have fixed the complex internal stability, assuming there is a problem. Perhaps @jeanwsr would have some test cases, or insight.

Another thing I have noticed is that this is able to match Hessian eigenvalues with `ghf_stability` (with sufficiently low `tol`) but the Davidson seems to be missing a lot of values. For example, given input
```py
import pyscf, pyscf.gto
from pyscf.scf.stability_slow import ghf_internal

mol = pyscf.gto.M(atom='O -.6 0 0; O .6 0 0', basis='cc-pvdz', verbose=4)
mf = pyscf.scf.GHF(mol).set(init_guess='hcore', conv_tol=1e-12).run()

ghf_internal(mf)
mf.stability(tol=1e-12)
```
I get 
```
converged SCF energy = -149.62899231417  <S^2> = 2.0326472  2S+1 = 3.0216864
ghf_internal: lowest eigs of H = [-1.53994323e-09 -1.53993634e-09  5.83048185e-02  5.83048185e-02
  8.95155351e-02]
<class 'pyscf.scf.ghf.GHF'> wavefunction is stable in the internal stability analysis
ghf_stability: lowest eigs of H = [0.05830482 0.05830482 0.40681822]
<class 'pyscf.scf.ghf.GHF'> wavefunction is stable in the internal stability analysis
```
Which doesn't seem to match until you look at all the eigenvalues:
```
ghf_internal: lowest eigs of H = [-5.23684871e-14 -2.74206340e-14  5.83048201e-02  5.83048201e-02
  8.95155367e-02  1.10808078e-01  1.10808078e-01  1.10808078e-01
  1.10808078e-01  2.37995350e-01  2.37995350e-01  4.06818222e-01
  4.45437914e-01  4.45437914e-01  4.45437914e-01  4.45437914e-01
  4.60494732e-01  4.60494732e-01  5.08729696e-01  5.08729696e-01]
```
Where it seems that `ghf_stability` misses two, gets two, then misses seven before getting the next eigenvalue. Maybe this is expected behavior though.

### DIFF
--- a/pyscf/scf/stability_slow.py
+++ b/pyscf/scf/stability_slow.py
@@ -288,18 +288,21 @@ def ghf_internal(mf, return_status=False, verbose=None):
         eri_mo_a+= lib.einsum('pqrs,pi,qj->ijrs', eri_ao, orbob.conj(), mob)
         eri_mo = lib.einsum('ijrs,rk,sl->ijkl', eri_mo_a, moa.conj(), moa)
         eri_mo+= lib.einsum('ijrs,rk,sl->ijkl', eri_mo_a, mob.conj(), mob)
+    # The orbital hessian is constructed as (ai|jb) in soscf/newton_ah.py
+    # (ai|jb) == iabj.conj()
     a += numpy.einsum('iabj->iajb', eri_mo[:nocc,nocc:,nocc:,:nocc].conj())
     a -= numpy.einsum('ijba->iajb', eri_mo[:nocc,:nocc,nocc:,nocc:].conj())
     b += numpy.einsum('iajb->iajb', eri_mo[:nocc,nocc:,:nocc,nocc:].conj())
     b -= numpy.einsum('jaib->iajb', eri_mo[:nocc,nocc:,:nocc,nocc:].conj())
 
-    a, b = numpy.reshape((a, b), (2, nocc*nvir, nocc*nvir))
+    a = a.reshape(nocc*nvir, nocc*nvir)
+    b = b.reshape(nocc*nvir, nocc*nvir)
     hessian = numpy.block([[a,b],[b.conj(),a.conj()]])
 
     # This factor is necessary to replicate the eigenvalues of ghf_stability
     hessian *= 2
 
-    e, v = numpy.linalg.eigh(hessian)
+    e, v = scipy.linalg.eigh(hessian)
     log.info('ghf_internal: lowest eigs of H = %s', e[:5])
 
     e = e[0]


jeanwsr:
If you want to find a "real" testcase for complex internal, you can try the tetrahedral H4 example [here](https://github.com/pyscf/pyscf/pull/1416/files#diff-d22de2dd1eb1d2a6239080c287177649fc1b92565aaa06d29c860dbd737cd68e). I'm sorry I have forgotten whether my previous slow version is correct (agrees with literature) or not.

NathanGillispie:
> 1. If you make `with_symmetry` an option in `ghf_stability`, as other stability functions do, you will find the missing roots come back when `with_symmetry=False`. I think this is reasonable since the exact diag does not limit the symmetry. I'm not sure whether the default value should be True or False, but we can make it an option first in the mf.stability.

That is very helpful. I'll look into adding the option.

> 3. I don't know why but when i run the example I occasionally get a higher SCF solution. Is it reproducible for you? Maybe the GHF initial guess has some problem.

Yes I'm running this on a different computer and I've noticed that too. It must have to do with the guess, because I can still match some eigenvalues.

> If you want to find a "real" testcase for complex internal, you can try the tetrahedral H4 example here. I'm sorry I have forgotten whether my previous slow version is correct (agrees with literature) or not.

Thanks, I'll look into that.

pyscf/scf/stability_slow.py
```diff
@@ -227,6 +230,99 @@ def uhf_external(mf, verbose=None):
...
+        eri_mo_a = lib.einsum('pqrs,pi,qj->ijrs', eri_ao, orboa.conj(), moa)
+        eri_mo_a+= lib.einsum('pqrs,pi,qj->ijrs', eri_ao, orbob.conj(), mob)
+        eri_mo = lib.einsum('ijrs,rk,sl->ijkl', eri_mo_a, moa.conj(), moa)
+        eri_mo+= lib.einsum('ijrs,rk,sl->ijkl', eri_mo_a, mob.conj(), mob)
+    a += numpy.einsum('iabj->iajb', eri_mo[:nocc,nocc:,nocc:,:nocc].conj())
+    a -= numpy.einsum('ijba->iajb', eri_mo[:nocc,:nocc,nocc:,nocc:].conj())
+    b += numpy.einsum('iajb->iajb', eri_mo[:nocc,nocc:,:nocc,nocc:].conj())
+    b -= numpy.einsum('jaib->iajb', eri_mo[:nocc,nocc:,:nocc,nocc:].conj())
+
+    a, b = numpy.reshape((a, b), (2, nocc*nvir, nocc*nvir))
```
sunqm:
```
a = a.reshape()
b = b.reshape()
```

sunqm:
@jeanwsr I agree that `with_symmetry` can be made as an argument for all stability functions. Other additional functionalities like distinguishing complex and real cases, I think they can be added when updating the iterative implementation in `stability.py`,  maybe in a separate PR. If you don't have more concerns, I will go ahead and merge this PR.

jeanwsr:
> @jeanwsr I agree that `with_symmetry` can be made as an argument for all stability functions. Other additional functionalities like distinguishing complex and real cases, I think they can be added when updating the iterative implementation in `stability.py`, maybe in a separate PR. If you don't have more concerns, I will go ahead and merge this PR.

please go ahead.

PR #3068, opened 2025-12-13
Pytblis interface

sunqm:
This PR adds a configurable backend for `einsum`. Supported backends include https://github.com/chillenb/pytblis and https://github.com/pyscf/pyscf-tblis

pytblis install instruction: https://github.com/pyscf/pyscf.github.io/pull/181


.github/workflows/ci_linux/python_deps.sh
```diff
@@ -8,6 +8,7 @@ version=$(python -c 'import sys; version=sys.version_info[:2]; print("{0}.{1}".f
 if [ $version != '3.12' ]; then
     pip install geometric
     pip install spglib
+    pip install pytblis
```
chillenb: The pypi wheels for pytblis are currently pthreads-based, but libgomp is better for high core counts. The numpy/scipy people don’t like to see openmp in wheels, but pyscf, pytorch, sklearn and others all do it.. Would you prefer libgomp-based wheels?

chillenb (commented):
Here are a few comments and questions. If there are things I can change on the pytblis side to make the integration easier, please let me know!

PR #3076, opened 2025-12-17
Faster SGX and SGX Gradients

kylebystrom:
This PR contains changes mostly focused on the SGX module, but with additional changes in a few related things.
* Analytical gradients for SGX.
* Improved "P-junction" (density matrix) screening for SGX. The aggressiveness of the screening can now be set with intuitive parameters that place approximate upper bounds on the energy and potential matrix error.
* Improved SGX parallelization (My previous SGX PR didn't scale well to many cores).
* Use optimized Orca-like grid scheme for SGX.
* Added a modified Becke partitioning scheme based on https://doi.org/10.1063/1.5049435 that is more accurate for weakly bound systems. It is used by default in SGX now. It contains a slight modification from the previous paper to make the analytical gradients smooth.
* Modified the Einstein summations in the J-matrix density fitting analytical gradients to significantly accelerate them.

As part of this PR, I made some significant changes to the user settings for the SGX module, so I will make a PR for the docs module as well. This is a draft while I perform a couple more tests, but I will prepare it for review ASAP. I will also attach some plots and benchmarks of the new version.

I have been slowly working on this on and off for a couple years, and there are many commits, so I figure it might be best to squash the commits when we merge.


pyscf/sgx/sgx_jk.py
```diff
@@ -53,14 +44,921 @@
...
+    @property
+    def _cintopt(self):
+        return self._opt._cintopt
+
+    @property
+    def use_dm_screening(self):
+        return self._screen_energy or self._screen_potential
+
+    def reset(self):
+        self._ovlp_proj = numpy.identity(self.mol.nao_nr())
```
sunqm, line 303: * This attr is initialized to None in `build()`. Is it needed to set it to identity matrix here?

* _stilde_ij and _mtilde_ij are not used by any functions.

sunqm (approved):

PR #3114, opened 2026-02-04
Add DF complex RCCSD

RHSong:
This PR adds DF complex RCCSD to the CC module. The code is largely based on rccsd.py, with some modifications to make it affordable for around 500 nmos.


pyscf/cc/__init__.py
```diff
@@ -111,7 +112,10 @@ def RCCSD(mf, frozen=None, mo_coeff=None, mo_occ=None):
         mf = mf.to_rhf()
 
     if isinstance(mf, _DFHF) and mf.with_df:
-        return dfccsd.RCCSD(mf, frozen, mo_coeff, mo_occ)
+        if numpy.iscomplexobj(mo_coeff) or numpy.iscomplexobj(mf.mo_coeff):
```
sunqm: Based on the eris generation logic in cmplx_dfccsd.py,  the complex dfccsd is intended for the mf objects of the off-gamma-point PBC HF calculations. However, this mf object is not an instance of _DFHF. this factory function will not produce a correct complex dfccsd for this mf.

Issue #3118, opened 2026-02-05
exxdiv problem of GDF band for hybrid DFT


jeanwsr:
In the documentation I found 
> For Hartree-Fock or hybrid DFT, the discontinuity of the exchange potential at G=0 is problematic for band structure calculations. Using :attr:`exxdiv='vcut_sph'` with FFTDF is recommended instead. Alternatively, the SCF procedure can be repeated at each k-point, which is much more expensive but allows the use of any :attr:`exxdiv` or density fitting.

which I do observe at hybrid DFT calculation with GDF. Although `exxdiv=vcut_sph` with FFTDF is recommended, it might be not applicable due to FFTDF's high cost.

Could I ask what is the current best workaround for it when using GDF? Is "repeat SCF at each k-point" still the only way?

PR #3127, opened 2026-02-15
Add PBC lo + Refactor molecular lo

hongzhouye:
A module `pbc/lo` is added to generate localized Wannier functions using the Pipek-Mezey method, as described in this [arXiv preprint](https://arxiv.org/abs/2602.12382). The molecular `lo` module is refactorized accordingly with improved efficiency.

@Genz17 contributed to the initial implementation.

Many thanks to @sunqm, @jeanwsr, and @xubo-wang for discussion on complex orbital rotations, to @tberkel for discussion on Wannier interpolations, and to @welltemperedpaprika for testing the initial version of the code.

## Refactor of molecular `lo` module
1. Refined `CIAHOptimizerMixin` for generality.
2. Added `CIAHOptimizerMixinComplex` for complex-valued rotations.
3. Added BFGS-based optimizer for `PM`, `Boys`, and `Edmiston`, complementary to existing CIAH optimizer.
4. Improved cost scaling for `PipekMezey` from `N^4` to `N^3`.
5. Added`PipekMezeyComplex` for complex unitary rotations.
6. Refined stability analysis: now supports both Newton-based stability analysis for `PM`, `Boys`, and `Edmiston` and Jacobi-based stability analysis for `PM`, with improved performance.
7. Added new tests and examples.

## Add `pbc/lo` module
1. `pbc/soscf/kciah.py`: Base class for k-point-based unitary rotations.
2. `base.py`: Base classes for k-point orbital rotations with both complex and real unitaries.
3. `kpipek.py`: k-point PM with both complex and real rotations
    - Support both Jacobi and Newton-based stability check.
    - Support both CIAH and BFGS optimizer
    - Support convenient extraction of Wannier functions in supercell
    - Support `meta-lowdin`, `iao`, and `iao-biorth` projectors
5. `interpolation.py`: Wannier interpolation (may be relevant to e.g., #3118).
6. Added new tests and examples.

## Note
To obtain real Wannier functions, one must
1. use real-valued unitary rotations, handled by this PR through `pbc.lo.KPMReal`
2. use Bloch orbitals that have time-reversal symmetry, dependent on a previous PR: #3126

## Performance benchmarks

The following shows the growth of CPU cost as system size for h-BN and CO/MgO by increasing the density of the k-point mesh in xy directions. Four methods are compared:
- `Gamma-CIAH (old)`: Old molecular PM code applied to supercell. The cost scales as `O(N^4)` due to Hessian--vector product (but in these tests some `O(N^3)` scaling terms dominate).
- `Gamma-CIAH (new)`: New molecular PM code from this PR applied to supercell. The cost scales as `O(N^3)`.
- `k-BFGS`: PBC k-point BFGS from this PR. The cost scales as `O(N^2)`.
- `k-CIAH`: PBC k-point CIAH from this PR. The cost scales as `O(N^2)`.

pr


pyscf/lo/stability.py
```diff
@@ -0,0 +1,140 @@
...
+    dump_status(log, stable, f'{mlo.__class__.__name__}', 'internal')
+    if stable:
+        mo = mlo.mo_coeff
+    else:
+        u = mlo.extract_rotation(v)
+        mo = mlo.rotate_orb(u)
+    if return_status:
+        return mo, stable
+    else:
+        return mo
```
jeanwsr: I think we can always return mo, stable here. In scf.stability we use return_status because of backward compatibility with old APIs, which seems not a problem in lo.

examples/pbc/34-pm_localization.py
```diff
@@ -2,30 +2,104 @@
...
+mo = np.asarray([x[:,:nocc] for x in mf.mo_coeff])
+mlo = lo.KPM(cell, mo, kpts)    # k-point PM with complex rotations
+mlo.kernel()
+
+# stability check
+while True:
+    mo, stable = mlo.stability(return_status=True)
+    if stable:
+        break
+    mlo.kernel(mo)
```
jeanwsr: Could this loop be wrapped into a function for convenience?

pyscf/lo/stability.py
hongzhouye: Thanks for the suggestion. Would it be better to make the default `True`? Having this switch makes it consistent with the previous Jacobi stability check for PM localization.

examples/pbc/34-pm_localization.py
tberkel: Looks fine to me since it's just an example, and I find it readable.

sunqm (approved):

examples/local_orb/07-boys_localization.py
sunqm, line 1: I've removed the `if __name__ == "__main__":` code blocks from several modules and put their contents in new examples. If these examples don't make much sense, please revise or remove them. @hongzhouye

PR #3132, opened 2026-02-18
Fix `import pyscf` under Python 3.7

wangenau:
While there is probably almost no one left using Python 3.7, one simple fix allows the execution of PySCF under it.
The current PySCF versions (2.12.x) fail during the import.
While at it, I have also updated the version classifiers and added a bound for the supported Python version, since PySCF does not run under Python 3.6 anyway.


wangenau:
CI failures are unrelated.

PR #3191, opened 2026-04-13
Canonical orthogonalization for eigenvalue solver in SCF

sunqm:
To improve numerical stability, especially in the PBC SCF calculations, for basis sets with diffuse functions, use canonical orthogonalization for eigenvalue solver by default for the SCF driver. Corresponding to https://github.com/pyscf/gpu4pyscf/pull/661

Key changes include:
* Removal of linearly dependent basis by default. Eigenvalues of the overlap matrix < 1e-6 are discarded. It is configurable by two flags at module level:
```
pyscf.scf.hf.remove_overlap_zero_eigenvalue = True
pyscf.scf.hf.overlap_zero_eigenvalue_threshold = 1e-6
```

* Phase for complex orbitals
The phase of complex orbitals is now fixed by ensuring that the largest component in each orbital coefficient is real.

* `scf.addons.remove_linear_dep_` function is deprecated

This PR only updates linear dependency treatment for the orbital basis. Linear dependency removal for auxiliary basis sets in density fitting methods is not included. The same treatment to density fitting can be considered in the future to improve numerical stability.


jeanwsr (approved):

jeanwsr:
Should we also remove this 
https://github.com/pyscf/pyscf/blob/c8a4215308657274f221937e5a8b8908a039c3a9/pyscf/pbc/mp/test/test_padding.py#L23

PR #3192, opened 2026-04-13
Disable density-based grid pruning by default

sunqm:
Corresponding to https://github.com/pyscf/gpu4pyscf/pull/715


fishjojo (approved):

pyscf/pbc/dft/uks.py
```diff
@@ -156,6 +156,15 @@ def dump_flags(self, verbose=None):
...
         return self
 
+    def initialize_grids(self, cell, dm, kpts, ground_state=True):
+        '''Initialize self.grids the first time call get_veff'''
+        if isinstance(self.grids, gen_grid.UniformGrids):
+            return self
+        elif self.grids.coords is not None:
+            return self
+        return rks.KohnShamDFT.initialize_grids(self, cell, dm[0]+dm[1], kpts,
+                                                ground_state)
```
MatthewRHermes: Just checking if `self.grids.coords is None` (here and in several other places) skips over a check related to NLC functionals. Is that OK?

PR #3208, opened 2026-05-08
Better SG1 radii parameter for ghost atom

henryw7:


MatthewRHermes:
Could you explain briefly what problem this solves?

sunqm:
SG1 scheme currently does not generate integration grids for ghost atoms. In a CP correction calculation, electron density associated with basis on ghost atoms is not properly integrated. In the default Treutler-Ahlrichs grids, grids are generated for ghost atoms. This PR applies the similar treatment for SG1 grids.

sunqm (approved):

PR #3209, opened 2026-05-12
upgrade dispersion

JiashuLiang:
* upgrade dftd4 and set atm of D4 default to True to be consistent with before

* allow format of disp 'd4:wb97x-3c' and update priority logic

### DIFF
--- a/pyscf/scf/test/test_uhf.py
+++ b/pyscf/scf/test/test_uhf.py
@@ -176,9 +176,9 @@ class KnownValues(unittest.TestCase):
         mol = make_disp_mol()
         mf = scf.UHF(mol)
         mf.disp = 'd3bj'
-        mf.conv_tol = 1e-10
-        e_tot = mf.kernel()
-        self.assertAlmostEqual(e_tot, -150.7949833081, 6)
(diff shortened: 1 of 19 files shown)


jeanwsr:
Is it possible to add a full wB97X-3c example or test here, I mean, with its specific basis set.

pyscf/scf/test/test_uhf.py
```diff
@@ -23,6 +23,25 @@
...
 
+try:
+    from pyscf.dispersion import dftd3, dftd4
+except (ImportError, OSError):
+    dftd3 = dftd4 = None
+
+
+def make_disp_mol():
+    return gto.M(
+        atom='''
```
sunqm, line 34: Consider using a different system. This one might be unstable for SCF convergence, introducing uncertainty into the results.

JiashuLiang:
> Is it possible to add a full wB97X-3c example or test here, I mean, with its specific basis set.

Ok, I will add one wB97X-3c example in pyscf/examples/dft/33-custom_disp.py

sunqm (approved):

PR #3214, opened 2026-05-17
Fix a batch of C-side correctness bugs across pbc, gto, vhf, dft, mcscf, ao2mo, mp, agf2

susilehtola:
## Summary

A focused four-round audit of every `.c` file under `pyscf/lib/` turned up
a substantial set of genuine correctness bugs. This PR ships fixes for them
in 19 commits across 17 source files, each commit topical and independently
buildable. The full PySCF test surface for the touched modules (294 tests
across dft, gto, scf, fci/mcscf, pbc) passes with no regressions throughout
the series.

Net change: **17 files, +268/-194 LOC.**

### How the audit ran

Each of four passes covered all 95 first-party `.c` files. Findings from
each pass were spot-checked by hand against the source before being acted
on; speculative claims and items that turned out to be intentional
behaviour (e.g. `mp2.c` non-s2symm "race" — benign because the racing
writes carry identical values by integral symmetry; `select_ci.c` missing
`NPdset0` — required by the Python wrapper's accumulation contract) were
rejected. One self-introduced regression during round three was caught by
the test suite and reverted in the same commit.

### Wrong-result bugs that affect normal use

- **`pbc/nr_direct.c::PBCVHF_direct_drv_nodddd`** — two stray lines
  unconditionally clobbered the `(k, l)` shell indices after the `s2kl`
  symmetry branch had just set them, silently disabling that branch.
- **`pbc/inner_dot.c::PBC_zdot_CNC_s1` and `_CNN_s1`** — the four-dgemm
  block writes accumulated into the head pointers `outR/outI` instead of
  the per-block `poutR/poutI` set up just above; wrong result for
  `na > BLKSIZE` plus a write race under `schedule(static)`. The k-point
  variants already did this correctly.
- **`pbc/optimizer.c::PBCdel_optimizer`** — inverted predicate
  `if (!opt0->rcut) free(opt0->rcut);` skipped the free whenever `rcut`
  was actually allocated.
- **`gto/fill_grids_int2c.c`** — `ish += ish0; jsh += jsh0;` was placed
  inside the per-grid-block loop; after the first block the shell indices
  were shifted by an extra `ish0/jsh0` every iteration.
- **`pdft/nr_numint.c::VOTdot_ao_mo`** — blocked dgemm wrote
  `vv + b0i*nao + b0j` with leading dim `&nmo`; output is `[nao, nmo]`
  so the row stride must be `nmo`. Always broken for the normal PDFT case
  `nao != nmo`.
- **`dft/libxc_itrf.c::LIBXC_max_deriv_order`** — never tested order 0
  and the `if (o == -1) return -1` branch was unreachable, so EXC-only
  functionals returned a misleading "max order" rather than 0.
- **`gto/nr_ecp.c::ECPtype2_cart`, `ECPtype_so_cart`** — missing
  `converged[ijl] = 1;` before the `CLOSE_ENOUGH` check (the sister
  `ECPtype1_cart` had it). Once any block failed to converge at a
  refinement level, every previously-converged block was re-entered at
  the next level and `prad[i] *= .5` was applied twice, silently
  corrupting the radial accumulator.
- **`gto/ft_ao.c::GTO_Gv_orth` / `_nonorth`** — used the literal macro
  `EXPCUTOFF` instead of `envs->expcutoff`. The sister `GTO_Gv_general`
  honours the per-mol override; the two diff paths silently ignored it.
- **`gto/grid_ao_drv.c::GTO_screen_index`** — `scale = -nbins/log(...)`
  was computed before the `nbins = MIN(127, nbins)` clamp, so the
  screening formula `si = nbins - arr*scale` mixed clamped offset with
  unclamped slope. Also fixed the `(uint8_t)(si+1)` wrap for very tight
  AOs.
- **`dft/multigrid.c::pgfpair_radius`** — compared raw signed components
  `rab[i] < RZERO` instead of `fabs(rab[i])`. Any all-negative displacement
  (~1/8 of periodic-image shifted pairs) wrongly took the same-atom branch
  and returned the wrong radius.
- **`pbc/hf_grad.c`** — `nl0->pairs[ish*nbas + jsh]` should index with
  `nl0->njsh` (OOB when shls_slice was narrower than full basis); plus
  the per-atom output stride was hardcoded `iatm*3+ic` instead of
  `iatm*comp+ic` (works for comp=3 only).
- **`gto/fill_nr_3c.c` / `gto/fill_r_3c.c`** — the `(*intor)(...)` call
  in `GTOnr3c_fill_s2ij` / `GTOr3c_fill_s2ij` ignored the return value;
  when libcint returned 0 (all primitives screened), `buf` was
  uninitialised and the subsequent `dcopy_s2_*` / `zcopy_s2_*` copies
  fed garbage into the packed output. Sister code in `fill_int2e.c`
  correctly gated on the return value.

### Latent / overflow / safety fixes

- **`gto/deriv2.c`** — `fx0/fy0/fz0/...` were sized `SIMDD*16` but the
  recurrence writes up to `(l+4)*SIMDD+n`; overruns at `l ≥ 12` (deriv4),
  `l ≥ 13` (deriv3), `l ≥ 14` (deriv2). Resized to `SIMDD*(LMAX+5)`
  matching the autocode template.
- **`gto/nr_ecp.c`** — three issues: `ECPsph_ine_opt` default branch
  overruns the K_TAB_COL-wide buffer when `order > 16` (now falls back to
  the slow routine in that range); `ECPtype_so_cart` had a duplicate
  `MALLOC_INSTACK(buf, ...)` that silently inflated cache use;
  `ECPdel_optimizer` set `opt = NULL` on its parameter copy instead of
  `*opt = NULL`; and the angular-momentum SO table is only sized for
  `lc ≤ 4` while the Ul fallback can push `lc` to 6 — now skipped with
  a stderr warning.
- **`vhf/fill_nr_s8.c`, `vhf/nr_incore.c`, `vhf/nr_direct.c`,
  `vhf/nr_sr_vhf.c`, `vhf/hessian_screen.c`, `vhf/optimizer.c`** — sweep
  of `int*int` overflow patterns in shell-pair indexing, packed-triangle
  offsets, and `nblock²` / `nblock³` loop bounds. Some were already
  exercised at the 65k-AO boundary, others were latent ceilings.
- **`vhf/rkb_screen.c::CVHFrkbssll_direct_scf_dm`** — `dm_cond` was
  malloc'd but never zeroed, unlike the sibling LL/SL setters. The
  strict-upper-triangle of the master SS slot was then read by the
  prescreen.
- **`vhf/nr_sgx_direct.c::SGXdiagonal_ints`** — three allocations
  (`buf`, `cache`, `dists`) were unused dead code, one with
  `sizeof(int)` for a `double *`.
- **`pbc/fill_ints.c::_nr2c_fill` and `pbc/fill_ints_screened.c::_nr2c_screened_fill`**
  — `int empty` declared and conditionally assigned inside the
  integrator branch, then returned via `return !empty;` — UB when no
  shell pair contributed. Both initialised to 1.
- **`pbc/fill_ints_screened.c`** — `assert(dk < dkmax)` should be `<=`
  (fired in debug builds when a single ksh exactly filled the buffer).
- **`pbc/nr_ecp.c`** — `eta = 1.f` init silently clamped the MIN
  reduction for atoms with ECP exponents all > 1 (typical tight cores).
  Init to FLT_MAX.
- **`gto/fill_int2c.c` / `gto/fill_grids_int2c.c`** — hermi-mode
  symmetrization assumes a square layout (`naoi == naoj && ish0 == jsh0`).
  Added asserts so a future caller passing a rectangular or offset slice
  fails loudly instead of silently corrupting off-diagonals.
- **`mp/mp2.c`** — `mp2.c` was at one point flagged for an apparent race
  in non-s2symm mode but careful re-analysis showed the racing writes
  carry identical values (by integral symmetry in same-spin, and the
  transpose write is the only filler in alpha-beta). No code change;
  noted for the record.
- **`mcscf/fci_rdm.c`** — `tril_particle_symm` had
  `(int)(48/norb)*norb` which collapsed to 0 for `norb > 48`, giving an
  infinite `for (m=0; m<nnorb-blk; m+=blk)` loop. Also three sites of
  `ket+stra_id*nb+strb_id` missing the `(size_t)` cast already applied
  at line 83 of the same file.
- **`mcscf/fci_string.c::FCIaddrs2str`** — three uses of `1UL` instead
  of `1ULL`; on LLP64 (Windows) `unsigned long` is 32 bits so these
  silently truncate for `nelec ≥ 32`.
- **`mcscf/select_ci.c::SCIcontract_2e_bbaa{,_symm}`** — documented why
  `ci1` must not be zeroed here: the Python wrapper calls this after
  `(aa|aa)` and `(bb|bb)` and relies on accumulation. Comment-only
  change; caught a self-introduced regression during the audit.
- **`agf2/uagf2.c::AGF2udf_vv_vev_islice_lowmem`** — `qx_j` and `qa_j`
  were built unconditionally using `j = ij % max(noa,nob)`; when
  `nob > noa`, `j ≥ noa` walked past the noa-dim of `qxi/qja`. Gate the
  alpha-side slice builds behind `do_ss`.
- **`np_helper/transpose.c` (`NPdtranspose_021` / `_z` / `NPdsymm_021_sum` /
  `NPzhermi_021_sum`), `pack_tril.c` (`NP{d,z}{unpack,pack}_tril_2d`),
  `condense.c` (`NP_Bmax` / `NP_imax` / `NP_fmax`)** — `(size_t)` casts
  on `shape[1]*shape[2]` / `n*n` / `n*(n+1)/2` plus zero-size guards
  matching the existing `NP_max` / `NP_min` / `NP_norm` style.

### Test-only

- **`pyscf/scf/test/test_rhf.py::test_get_vj`** — was already failing
  on clean master (`AttributeError: module 'pyscf.scf' has no attribute
  '_vhf'` because the test referenced `scf._vhf._fpointer(...)` without
  the explicit `from pyscf.scf import _vhf`). Fixed.

## Out of scope (not in this PR)

Items flagged by the audit but left for separate work:

- `lib/np_helper/omp_reduce.c::NPomp_dsum_reduce_inplace1` — has a
  legitimate read-write race independent of thread count. Callers
  should switch to the linear `NPomp_dsum_reduce_inplace`; not done here
  because it touches multiple call sites.
- `gto/fill_nr_3c.c` / `fill_r_3c.c` — packed-triangle indexing mixes
  absolute `ip = ao_loc[ish]` with relative `jp = ao_loc[jsh] -
  ao_loc[jsh0]`; currently masked because callers always pass
  `jsh0 = 0`. Fixing it would change the storage convention and needs
  Python-side coordination.
- `pbc/pp.c::_contract_vnuc_ip1_dm` — writes only `iatm` and `katm`
  gradient slots; the `jatm` contribution may be missing for
  `iatm != jatm`. Needs a finite-difference cross-check against the
  Python energy gradient before patching.
- `pbc/nr_ecp.c::PBCECP_loop` — `li`/`lj`/`nfi`/`nfj` derived from the
  first shell of a segment range while `dij` sizes the output buffer.
  The author's own comments at lines 238 and 328 acknowledge "ish0 can
  be wrong"; out-of-scope here.
- `np_helper/npdot.c::NPdgemm` / `NPzgemm` k==0 — unconditionally zeros
  C, violating the BLAS contract `C := beta * C` when `k = 0`. Real
  spec deviation but exotic trigger.
- `vhf/hessian_screen.c::CVHFipip1_prescreen` / `CVHFipvip1_prescreen`
  — tests only `dm_cond[l*n+k]`; for non-Hermitian dm1 (response
  solvers) the 3rd/4th contractions get wrongly skipped. Today's RHF
  hessian / grad callers all pass symmetric DMs.
- `mcscf/fci_string.c::binomial` — `static int` recursive value
  overflows at `C(34, 17)`; addressing it cleanly requires also
  promoting the cache to `int64_t` and updating callers.
- `ao2mo/{nr,r}_ao2mo.h` — both define `struct _AO2MOEnvs` with the
  same tag but different members. No current TU includes both; pure
  hazard.
- Several copy-paste families that would benefit from templating
  (`vhf/r_direct_dot.c` and siblings, `gto/ft_ao_ppnl.c` vs
  `auto_ftao/ft_ao_ppnl_auto.c`, `ao2mo/{nr,nrr,r}_ao2mo.c`,
  `dft/{r,nr,nr_numint_sparse,pdft/nr_numint}.c`,
  `pbc/{fill_ints,fill_ints_sr,fill_ints_screened}.c`) — high-value but
  large mechanical changes deferred to separate PRs.

## Test plan

Each round of fixes was followed by the same test set. Final tally:

- `pytest pyscf/dft/test/{test_libxc, test_numint, test_grids}.py` — 48 passed
- `pytest pyscf/gto/test/{test_ft_ao, test_moleintor, test_ecp, test_mole, test_eval_gto, test_pp_int}.py` — 96 passed
- `pytest pyscf/scf/test/{test_jk, test_h2o, test_rhf, test_uhf}.py` — 126 passed
- `pytest pyscf/fci/test/{test_spin0, test_spin1, test_rdm, test_selected_ci}.py pyscf/mcscf/test/{test_addons, test_casci}.py` — 76 passed
- `pytest pyscf/pbc/df/test/{test_aft, test_fft}.py pyscf/pbc/scf/test/test_khf.py` — 40 passed
- Smoke test: RHF / HF / cc-pVDZ converges to the expected energy

**Total: 386 + 1 passed, 0 regressions.**


pyscf/lib/vhf/optimizer.c
```diff
@@ -460,7 +460,7 @@ void CVHFset_int2e_q_cond(int (*intor)(), CINTOpt *cintopt, double *q_cond,
         CVHFnr_int2e_q_cond(intor, cintopt, q_cond, ao_loc, atm, natm, bas, nbas, env);
 }
 
-void CVHFset_q_cond(CVHFOpt *opt, double *q_cond, int len)
+void CVHFset_q_cond(CVHFOpt *opt, double *q_cond, size_t len)
```
sunqm: Do not change this "int" type, as this function is called via ctypes in Python code. Changing this type would require to scan over all Python callers (some via function pointer) to ensure type consistency.

pyscf/lib/vhf/optimizer.c
```diff
@@ -535,7 +535,7 @@ void CVHFsetnr_direct_scf_dm(CVHFOpt *opt, double *dm, int nset, int *ao_loc,
         CVHFnr_dm_cond(opt->dm_cond, dm, nset, ao_loc, atm, natm, bas, nbas, env);
 }
 
-void CVHFset_dm_cond(CVHFOpt *opt, double *dm_cond, int len)
+void CVHFset_dm_cond(CVHFOpt *opt, double *dm_cond, size_t len)
```
sunqm: Do not change this "int" type

pyscf/lib/gto/fill_nr_3c.c
```diff
@@ -173,13 +173,19 @@ void GTOnr3c_fill_s2ij(int (*intor)(), double *out, double *buf,
...
-                if (ip != jp) {
-                        dcopy_s2_igtj(pout, buf, comp, ip, nij, nijk, di, dj, dk);
-                } else {
-                        dcopy_s2_ieqj(pout, buf, comp, ip, nij, nijk, di, dj, dk);
+                // intor returns 0 when all primitives are screened out and
+                // does not guarantee buf has been zeroed. Skip the copy in
+                // that case so dcopy_s2_* doesn't propagate uninitialised
+                // values into the output. Sister code in fill_int2e.c uses
+                // the same return-value gate.
+                if ((*intor)(buf, NULL, shls, atm, natm, bas, nbas, env,
```
sunqm: This change needs to be reverted. `buf` is zeroed by `intor()`. If the copy from `buf` to the output is skipped, the corresponding output region needs to be explicitly zeroed. That is how it is implemented in `fill_int2e.c`.

pyscf/lib/gto/fill_r_3c.c
```diff
@@ -157,11 +157,16 @@ void GTOr3c_fill_s2ij(int (*intor)(), double complex *out, double complex *buf,
...
                 k0 = ao_loc[ksh  ] - ao_loc[ksh0];
-                (*intor)(buf, NULL, shls, atm, natm, bas, nbas, env, cintopt, cache);
-                if (ip != jp) {
-                        zcopy_s2_igtj(out+k0*nij, buf, comp, ip, nij, nijk, di, dj, dk);
-                } else {
-                        zcopy_s2_ieqj(out+k0*nij, buf, comp, ip, nij, nijk, di, dj, dk);
+                // intor returns 0 when all primitives are screened out and
+                // does not guarantee buf has been zeroed. Skip the copy in
+                // that case (matching the gate used in fill_int2e.c).
+                if ((*intor)(buf, NULL, shls, atm, natm, bas, nbas, env,
```
sunqm: Similar to fill_nr_3c.c, this change needs to be reverted.

pyscf/lib/pbc/nr_ecp.c
```diff
@@ -129,7 +130,11 @@ int PBCECP_loop(Function_cart intor,
                 ksh1 = ecploc[esh+1];
                 ecpbas = all_ecpbas + ksh0 * BAS_SLOTS;
                 necpbas = ksh1 - ksh0;
-                eta = 1.f;
+                // Init to FLT_MAX so the MIN-reduction below returns the
```
sunqm: Revert to `eta = 1.0f` here. The screening estimator code below may be inaccurate for core orbitals and can potentially miss required images in the lattice sum. `eta = 1.0f` will always include sufficient images, even for large-exponent orbitals, although it may introduce some overhead by including more images than necessary.

pyscf/lib/gto/grid_ao_drv.c
```diff
@@ -32,8 +32,13 @@ void GTO_screen_index(uint8_t *screen_index, int nbins, double cutoff,
...
                       int *atm, int natm, int *bas, int nbas, double *env)
 {
-        double scale = -nbins / log(MIN(cutoff, .1));
+        // Clamp nbins before computing scale so that si = nbins - arr*scale
+        // (used downstream) stays self-consistent: both terms reference the
+        // same effective nbins. Without this, callers passing nbins > 127
+        // get a different screening map than callers passing nbins <= 127
+        // with the same cutoff.
         nbins = MIN(127, nbins);
+        double scale = -nbins / log(MIN(cutoff, .1));
```
sunqm, line 39: This order cannot be changed. scale needs to be evaluated using the input nbins. `nbins < 120` can be asserted

pyscf/lib/vhf/optimizer.c
susilehtola: Thanks @sunqm, reverted.

pyscf/lib/vhf/optimizer.c
susilehtola: Thanks @sunqm, reverted.

pyscf/lib/gto/fill_nr_3c.c
susilehtola: Thanks @sunqm, reverted.

pyscf/lib/gto/fill_r_3c.c
susilehtola: Thanks @sunqm, reverted.

pyscf/lib/pbc/nr_ecp.c
susilehtola: Thanks @sunqm, reverted.

pyscf/lib/gto/grid_ao_drv.c
susilehtola: Thanks @sunqm, reverted.

pyscf/lib/vhf/optimizer.c
susilehtola: Reverted in b05a8e0e3. Both signatures restored to `int len`.

pyscf/lib/vhf/optimizer.c
susilehtola: Reverted in b05a8e0e3 along with CVHFset_q_cond.

pyscf/lib/gto/fill_nr_3c.c
susilehtola: Reverted in e342974d3. Understood — `buf` is zeroed inside `intor()`, so the original unconditional copy is correct.

pyscf/lib/gto/fill_r_3c.c
susilehtola: Reverted in e342974d3 together with fill_nr_3c.c.

pyscf/lib/pbc/nr_ecp.c
susilehtola: Reverted in 386a1254c. Restored `eta = 1.0f` and dropped the unused `<float.h>` include.

pyscf/lib/gto/grid_ao_drv.c
susilehtola: Fixed in 43df6a784. Restored the original order and replaced the moved clamp with `assert(nbins < 120)`.

sunqm (approved):

PR #3219, opened 2026-05-22
Add qavg-vSZPs basis and ecp-q-vSZP to PySCF

manassharma07:
This PR adds the qavg-vSZPs averaged valence basis set and its companion ecp-q-vSZP ECP to PySCF.
These were released recently and are available here: https://github.com/grimme-lab/q-vSZPs 

Changes:
- Add qavg-vszps.dat in PySCF/NWChem basis format
- Add ecp-q-vszp.dat in PySCF/NWChem ECP format
- Register aliases qavg-vSZPs and ecp-q-vSZP

The data were converted from grimme-lab/q-vSZPs.


MatthewRHermes:
Can you put literature citations in commentary in the `.dat` files? I know you have the link to the other github but a citation would be more convenient in a lot of cases.

manassharma07:
Added the references.

MatthewRHermes:
Can you tell me if I'm doing something wrong? I'm trying to reproduce the total energy I found in the SI of that paper with this commit and I'm not succeeding.

```
from pyscf import gto, scf, dft

xyz = '''O 0.0000000 0.0000000 -0.3893611
H 0.7629844 0.0000000 0.1946806
H -0.7629844 0.0000000 0.1946806'''

mol = gto.M (atom=xyz,
             basis='qavgvszps',
             ecp='ecpqvszp',
             output='h2o.log',
             verbose=5)
mf = dft.RKS (mol, xc='wb97x-d4')
mf.grids.level = 9
mf.kernel ()
```

Yields a total energy of `-17.1934104317619`, when it should be `−17.226384515152` according to the paper. Do you think you could add a unit test and example script validating that you get exactly the same results as in Orca?

tberkel:
Maybe the discrepancy is related to #3217?

manassharma07:
@tberkel @MatthewRHermes 
I think this discrepancy is more related to the fact that the example uses q-vSZPS and not qavg-vSZPs (which is what I have added).
The q-vSZPs are "breathing" basis sets and change based on the charges.
That is why the example provides the charges as input.
This PR adds the non-adaptive basis sets, which are mentioned in the introduction of the preprint and available here:
https://github.com/grimme-lab/q-vSZPs/blob/main/q-vSZPs_basis/qavg-vszps_basis_TM

MatthewRHermes:
OK so can you guide me to an example calculation I can run to validate that this works and is ported properly, or else just add one to the PR? I only see one example calculation in the link and it seems to be the one I already tried. I don't have Orca.

MatthewRHermes:
> Maybe the discrepancy is related to #3217?

Partially, but not completely. Thanks for the pointer but fixing that only takes me to `-17.2184925266447`.

manassharma07:
I've now got a reference for the non-adaptive basis set. Will use that as the reference and add some examples.

sunqm (approved):

PR #3222, opened 2026-05-26
GW improvements: PBC G0W0

lijiachen417:
This is the second of several updates to the GW/RPA code from the [Zhu lab](https://github.com/ZhuGroup-Yale)

- PBC KRGWAC and KURGWAC: k-point G0W0, improve CPU and memory efficiency, add more features
- PBC GWAC: Gamma-point G0W0
- Additional test coverage

Contributors: Tianyu Zhu (@zhutianyu1991 ), Jiachen Li (@lijiachen417 ), Chris Hillenbrand (@chillenb )


jeanwsr:
Could you add or update examples for the new features.

lijiachen417:
> Could you add or update examples for the new features.

Yea sure

lijiachen417:
Failed test case is from old GW test file, and the difference is only 1e-8 due to numerical noise in analytical continuation. I think it's safe to ignore it.

pyscf/pbc/gw/gw_ac.py
```diff
@@ -0,0 +1,215 @@
...
+        mo_coeff : double 2d array
+            orbital coefficient
+
+        Returns
+        -------
+        vk : double 2d array
+            exchange self-energy
+        """
+        dm = self._scf.make_rdm1()
+        if isinstance(self._scf.with_df, df.GDF):
```
sunqm, line 204: You can simply execute `rhf = self._scf.copy()` for a new instance of the scf object

pyscf/pbc/gw/gw_ac.py
```diff
@@ -0,0 +1,215 @@
...
+        elif isinstance(self._scf.with_df, df.RSDF):
+            rhf = scf.RHF(self.mol).rs_density_fit()
+        if hasattr(self._scf, 'sigma'):
+            rhf = scf.addons.smearing_(rhf, sigma=self._scf.sigma, method=self._scf.smearing_method)
+        rhf.exxdiv = None
+        rhf.with_df = self.with_df
+        vk = rhf.get_veff(self.mol, dm) - rhf.get_j(self.mol, dm)
+        vk = reduce(np.matmul, (mo_coeff.T, vk, mo_coeff))
+
+        if self.fc:
```
sunqm, line 215: In the existing implementation the `fc` flag is related to the exxdiv provided by the underlying SCF object. However, the `fc` here is a different treatment. Is this `fc` related to the existing exxdiv corrections?

lijiachen417: > In the existing implementation the `fc` flag is related to the exxdiv provided by the underlying SCF object. However, the `fc` here is a different treatment. Is this `fc` related to the existing exxdiv corrections?

No, this `fc` flag is related to the GW finite-size correction developed in [this paper](https://pubs.acs.org/doi/full/10.1021/acs.jctc.0c00704). It's implemented based on the k-point GW code [here](
https://github.com/lijiachen417/pyscf/blob/83eb49919373514d76c178cdd8a039e6e22d4605/pyscf/pbc/gw/krgw_ac.py#L79). PySCF functions are called to build the exchange self-energy, which is same as the HF exchange matrix without finite-size correction.

pyscf/pbc/gw/gw_ac.py
lijiachen417: > You can simply execute `rhf = self._scf.copy()` for a new instance of the scf object

Because `self._scf` can be a non-HF object, can I still use `copy()`?

sunqm: If the class of `self._scf` inherits from `lib.StreamObject`, it has the copy() method. You can use copy() and a shallow copy of the instance will be created

lijiachen417: > If the class of `self._scf` inherits from `lib.StreamObject`, it has the copy() method. You can use copy() and a shallow copy of the instance will be created

Here we want the HF exchange matrix in line 208 ` vk = rhf.get_veff(self.mol, dm) - rhf.get_j(self.mol, dm)`, so if `self._scf` is a DFT object, we might need one more `to_rhf()` step. Do you prefer that way? I can change to it.

pyscf/pbc/gw/gw_ac.py
sunqm: The previous fc correction seems based on the exxdiv. Since this is changed, could you add a comment to document the fc treatment in this file or in the krgw_ac code.

sunqm (approved):

pyscf/pbc/gw/gw_ac.py
lijiachen417: > Could you add a comment to explain the fc in the relevant files, either in this PR or the other GW PR

OK, I will add the explanation to the docstring in this function in this PR.

lijiachen417: > The previous fc correction seems based on the exxdiv. Since this is changed, could you add a comment to document the fc treatment in this file or in the krgw_ac code.

Added above.

PR #3224, opened 2026-05-26
Use the DFTD4 recommended variants for wb97x-d4

sunqm:
In previous versions, specifying `mf.xc='wb97x-d4'`, the DFT XC part was evaluated using the original wb97x functional. This PR follows dftd4 recommendations, changing the XC part to wb97x-v without vv10. This effectively executes an DFT calculation with
```
mf.xc = 'wb97x-v'
mf.nlc = False
mf.disp = 'd4:wb97x'
```
To reproduce previous pyscf behaviors, the following settings can be used.
```
mf.xc = 'wb97x'
mf.disp = 'd4:wb97x-2008'
```


jeanwsr:
I don't understand why previous behavior is wb97x + d4:wb97x-2008. Since it was not whitelisted, the previous behavior looks like wb97x + d4:wb97x?

JiashuLiang:
> I don't understand why previous behavior is wb97x + d4:wb97x-2008. Since it was not whitelisted, the previous behavior looks like wb97x + d4:wb97x?

This because the new dftd4 package has renamed wb97x to wb97x-2008 and wb97x now refers to another D4 parameter

jeanwsr:
> > I don't understand why previous behavior is wb97x + d4:wb97x-2008. Since it was not whitelisted, the previous behavior looks like wb97x + d4:wb97x?
> 
> This because the new dftd4 package has renamed wb97x to wb97x-2008 and wb97x now refers to another D4 parameter

But that renaming happened one year ago. By "previous behavior" I think it means the behavior right before this pr, which should be `d4:wb97x`. `d4:wb97x-2008` sounds like a previous previous behavior?

JiashuLiang:
> > > I don't understand why previous behavior is wb97x + d4:wb97x-2008. Since it was not whitelisted, the previous behavior looks like wb97x + d4:wb97x?
> > 
> > 
> > This because the new dftd4 package has renamed wb97x to wb97x-2008 and wb97x now refers to another D4 parameter
> 
> But that renaming happened one year ago. By "previous behavior" I think it means the behavior right before this pr, which should be `d4:wb97x`. `d4:wb97x-2008` sounds like a previous previous behavior?

The dftd4 upgrade is one year ago but there is another api layer "pyscf-dispersion" betwee pyscf and dftd4. pyscf-dispersion was just upgraded this year.

sunqm:
I have removed the misleading warning message. @jeanwsr

jeanwsr:
Let's apply @JiashuLiang 's suggestion.

jeanwsr (approved):

PR #3226, opened 2026-05-27
Add TREXIO interface in pyscf.tools.trexio

susilehtola:
## Summary

Adds `pyscf.tools.trexio`, a converter to and from the [TREXIO](https://github.com/TREX-CoE/trexio) file format. TREXIO is the standardized exchange format developed by the TREX Center of Excellence and is the natural bridge between PySCF and external QMC codes (TurboRVB, QMCPACK via converters, CHAMP, etc.).

The new module supports:

- Geometry, electron counts, point group, nuclear repulsion.
- Gaussian basis sets in spherical or Cartesian form. General contractions are unfolded into one TREXIO shell per contraction; primitives are written as `coefficient * prim_factor` so a round-trip through PySCF reproduces the basis exactly.
- ECPs (`z_core`, local-channel sentinel, radial-power slots).
- MOs from RHF / UHF / ROHF.
- Optional AO 1e integrals (S, T, V_ne, ECP, h_core, dipoles) and AO 2e ERIs stored sparsely in 8-fold-symmetric chemist notation.
- Spherical-AO permutation between PySCF (m = -l..+l) and TREXIO (m = 0, +1, -1, +2, -2, ...).

### Public API

```python
from pyscf.tools import trexio
trexio.to_trexio(mf, 'h2o.h5', with_ao_ints=True, with_eri=True)
mf2 = trexio.scf_from_trexio('h2o.h5')
mol = trexio.mol_from_trexio('h2o.h5')
ints = trexio.read_ao_1e_integrals('h2o.h5')
eri  = trexio.read_ao_2e_integrals('h2o.h5')
```

`trexio` is an optional dependency: `pyscf.tools.trexio` only fails to import if the user explicitly tries to use it without `pip install trexio`. `pyscf.tools.__init__` is unchanged.

## Test plan

- [x] `python -m unittest pyscf.tools.test.test_trexio -v` — 7 tests pass
- [x] RHF round-trip (spherical), exact AO overlap reproduction
- [x] RHF round-trip (Cartesian, including d/f shells)
- [x] UHF round-trip on triplet O2 (spin labels preserved)
- [x] ECP round-trip on CuH with LANL2DZ
- [x] AO 1e integral round-trip (S, T, h_core)
- [x] AO 2e ERI round-trip
- [x] `NORMALIZE_GTO` restored on read error

The test auto-detects whether the installed `trexio` was built with HDF5 and falls back to TEXT backend otherwise.

### DIFF
--- a/examples/tools/30-trexio.py
+++ b/examples/tools/30-trexio.py
@@ -26,7 +26,8 @@ mf = scf.RHF(mol).run()
 trexio.to_trexio(mf, 'h2o.trexio',
                  backend='TEXT',     # use 'HDF5' if trexio was built with HDF5
                  with_ao_ints=True,  # write S, T, V_ne, h_core, dipole
-                 with_eri=True)      # write 8-fold-symmetric AO ERIs
+                 with_eri=True,      # write 8-fold-symmetric AO ERIs
+                 with_mo_eri=True)   # write 8-fold-symmetric MO ERIs
 
 # 2) Reconstruct a mean-field object from the file.
 mf2 = trexio.scf_from_trexio('h2o.trexio', backend='TEXT')
@@ -40,4 +41,6 @@ print('nao =', mol2.nao_nr())
 ints = trexio.read_ao_1e_integrals('h2o.trexio', mol=mol2, backend='TEXT')
 print('overlap shape =', ints['overlap'].shape)
 eri = trexio.read_ao_2e_integrals('h2o.trexio', mol=mol2, backend='TEXT')
-print('eri shape =', eri.shape)
+print('AO eri shape =', eri.shape)
+mo_eri = trexio.read_mo_2e_integrals('h2o.trexio', backend='TEXT')
+print('MO eri shape =', mo_eri.shape)
--- a/examples/tools/30-trexio.py
+++ b/examples/tools/30-trexio.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python
+'''
+Save and load PySCF objects via the TREXIO file format.
+
+TREXIO (https://github.com/TREX-CoE/trexio) is a portable, standardized
+file format used in particular for handoff between quantum chemistry
+codes and QMC packages.
+
+Requires the ``trexio`` Python package: ``pip install trexio``.
+'''
+
+from pyscf import gto, scf
+from pyscf.tools import trexio
+
+# 1) Run an SCF and dump everything (geometry, basis, MOs, AO integrals,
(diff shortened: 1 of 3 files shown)


susilehtola:
ping @scemama

jeanwsr:
what's the relation between this and https://github.com/pyscf/pyscf-forge/blob/master/pyscf/tools/trexio.py ?

susilehtola:
> what's the relation between this and https://github.com/pyscf/pyscf-forge/blob/master/pyscf/tools/trexio.py ?

good question; I just checked that PySCF proper didn't have a trexio interface. Since this is important for interoperability, the functionality should really be in core PySCF.

scemama:
Nice initiative! But @kousuke-nakano has already done an implementation accepted in pyscf-forge (as mentioned by @jeanwsr). Maybe you should talk to Kousuke and see what should be added to make the interface complete.

kousuke-nakano:
Dear @jeanwsr, @susilehtola, @scemama, Yes, I have implemented a `pyscf to trexio` converter in `pyscf-forge`, and it has been merged into the `main` branch and published to `PyPI`. One can try it via `pip install pyscf-forge`. I plan to merge it into the main `pyscf` repository asap (I did not do it because we should test it for a couple of months, according to the pyscf-forge guidelines).

kousuke-nakano:
Dear @jeanwsr and @scemama, I have opened a feature-transfer request (pyscf/pyscf-forge#195) to transfer the TREXIO implementation from `PySCF-forge` to `PySCF-Core`.

@susilehtola, First of all, thank you very much for contributing to TREXIO. I had a look at your implementation, and since the implementation in `pyscf-forge` already covers the features discussed here, as well as PBC support and other functionalities, I wonder if it would be better to consolidate our efforts and use the `pyscf-forge` implementation as the basis for further development.

The `pyscf-forge` implementation was developed several years ago and has been tested and used since then. Therefore, I think it would provide a stable starting point for the TREXIO implementation.

I would very much appreciate your feedback, especially if some functionality is missing or if there are aspects that should be improved. Please also feel free to modify the code directly or suggest changes in the PR.

sunqm:
@susilehtola, what do you think about merging the two TREXIO implementations and using the `pyscf-forge` version as the basis?

susilehtola:
> @susilehtola, what do you think about merging the two TREXIO implementations and using the `pyscf-forge` version as the basis?

I am fine with using the already tested pyscf-forge version as the basis; I wasn't aware that one already existed. The main thing is to have full trexio I/O functionality in PySCF.

PR #3228, opened 2026-05-29
GW improvements: Bethe-Salpeter equation

lijiachen417:
This is the third of several updates to the GW/RPA code from the [Zhu lab](https://github.com/ZhuGroup-Yale)
This PR adds Bethe-Salpeter equation (BSE) to the molecule GW module.

- support both spin-restricted and spin-unrestricted
- BSE can be solved by Davidson algorithm, full diagonalization and Lanczos algorithm
- support energy-specific BSE

### DIFF
--- a/examples/gw/04-bse.py
+++ b/examples/gw/04-bse.py
@@ -1,5 +1,24 @@
 """
 Example for Bethe-Salpeter equation.
+
+########
+Reference results for H2O / PBE / def2-SVP
+
+* GW step (fully analytic GW, quasiparticle equation solved iteratively)
+                HOMO (eV)   LUMO (eV)
+Turbomole        -8.78        2.75
+PySCF            -8.79        2.75
+
+* First three singlet excitations (eV) for BSE
+                S1      S2      S3
+Turbomole      3.41    7.35    8.56
+PySCF          3.41    7.36    8.58
+
+* First three triplet excitations (eV) for BSE
+                T1      T2      T3
+Turbomole      2.68    4.67    7.14
+PySCF          2.67    4.67    7.14
+
 """
 import numpy as np
 from pyscf import gto, dft
--- a/examples/gw/04-bse.py
+++ b/examples/gw/04-bse.py
@@ -1,0 +1,79 @@
+"""
+Example for Bethe-Salpeter equation.
+"""
+import numpy as np
+from pyscf import gto, dft
+from pyscf.gw.gw_ac import GWAC
+from pyscf.gw.ugw_ac import UGWAC
+from pyscf.gw.bse import BSE, bse_lanczos, lanczos_estimate_spectrum
+
+# restricted
+mol = gto.Mole()
+mol.verbose = 5
+mol.atom = [[8, (0.0, 0.0, 0.0)], [1, (0.7571, 0.0, 0.5861)], [1, (-0.7571, 0.0, 0.5861)]]
+mol.basis = 'def2-svp'
+mol.build()
+mf = dft.RKS(mol, xc='pbe')
+mf.kernel()
+
+# GW-AC/BSE
+gw = GWAC(mf)
+gw.kernel()
+bse = BSE(gw)
+# Davidson algorithm for singlet excitation
+bse.TDA = False
+bse.kernel('s')
+bse.analyze()
+# Davidson algorithm for triplet excitation, turn on TDA
+bse.TDA = True
+bse.kernel('t')
+bse.analyze()
+# full diagonalization for triplet excitation
+bse.full_diagonalization('t')
+bse.analyze()
+
+eta = 0.01 # spectrum broadening in eV
+omega = np.linspace(0.0, 1.0, 1000)[:, None] + 1j * eta # (nω, 1)
+
+ao_dip = mol.intor('int1e_r', comp=3)
+nocc = mol.nelectron // 2
+mo_dip = np.einsum('xij,ia,jb->xab', ao_dip, mf.mo_coeff[:, :nocc], mf.mo_coeff[:, nocc:])
+
+bse.TDA = False
+lanczos_spectra = []
+for j in range(3):
+    alphas, betas = bse_lanczos(bse, multi='s', u1=mo_dip[j].flatten(), nsteps=500)
+    freqs, density = lanczos_estimate_spectrum(alphas, betas, (0, 1), eta, 1000)
+    lanczos_spectra.append(density)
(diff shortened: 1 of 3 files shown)


jeanwsr:
Is it possible to use the same Davidson solver as TDDFT, instead of adding a new one.

lijiachen417:
> Is it possible to use the same Davidson solver as TDDFT, instead of adding a new one.

In principle yes. But to integrate features such as selecting energy-specific excitations and restarting Davidson, we choose the current implementation. Another reason is that the Davidson algorithm and the Lanczos algorithm in BSE share some function.

pyscf/gw/bse.py
```diff
@@ -0,0 +1,2006 @@
...
+    for j in range(nroot):
+        for s in range(nspin):
+            dipole[:, j] += np.einsum('ia,xia->x', X_vec[s][j], mo_dip[s]) + np.einsum(
+                'ia,xia->x', Y_vec[s][j], mo_dip[s]
+            )
+
+    if nspin == 1:
+        dipole *= np.sqrt(2)
+
+    oscillator_strength = (2 / 3) * exci * np.sum(dipole**2, axis=0)
```
jeanwsr, line 1625: is this identical to tdscf.*.oscillator_strength

lijiachen417: > is this identical to tdscf.*.oscillator_strength

Good point, I will check if this can be replaced by a td function.

lijiachen417: > > is this identical to tdscf.*.oscillator_strength
> 
> Good point, I will check if this can be replaced by a td function.

Since the current BSE implementation is not based on `tdscf` class, so using pyscf function for getting oscillator strength and transition dipole will not make the code simpler (current code is already only several lines). I can close this PR and try to make BSE implementation tdscf-style, but I'm not sure if energy-specific Lanczos implementations can be easily changed. @chillenb @sunqm what do you think?

sunqm: Reusing the tddft code may provide some simplicity but that may also require refactoring tddft code. This involves too much work for a BSE implementation. I think the current BSE implementation is fine as-is.

pyscf/gw/bse.py
```diff
@@ -0,0 +1,2006 @@
...
+        # check spin and multiplicity
+        assert isinstance(multi, str)
+        multi = multi[0].lower()
+        assert (self.nspin == 1 and (multi == 's' or multi == 't')) or (self.nspin == 2 and multi == 'u')
+        self.multi = multi
+
+        cput0 = (time.process_time(), time.perf_counter())
+        self.dump_flags()
+        self.check_memory()
+        exci, X, Y = bse_davidson(bse=self, multi=multi, e_min=e_min, delta=delta, **kwargs)
```
sunqm: exci, X, Y are stored in `full_diagonalization`. Should these results be also saved in the kernel function?

sunqm (approved):

pyscf/gw/bse.py
```diff
@@ -0,0 +1,2006 @@
...
+
+    Lia = [np.ascontiguousarray(Lpq[s][:, : nocc[s], nocc[s] :]) for s in range(nspin)]
+    Laa = [np.ascontiguousarray(Lpq[s][:, nocc[s] :, nocc[s] :]) for s in range(nspin)]
+    Lii_bar, Lia_bar = _get_lpq_bar_by_block(
+        nocc=nocc, mo_energy=mo_energy, Lii=[Lpq[s][:, : nocc[s], : nocc[s]] for s in range(nspin)], Lia=Lia
+    )
+
+    if precond_exact_diag:
+        assert TDA
+        Laa_diag = [np.diagonal(Laa[s], axis1=0, axis2=2) for s in range(nspin)]
```
sunqm: Should this be `axis1=1, axis2=2` ?

pyscf/gw/bse.py
lijiachen417: Fixed. Now eigenvalues and eigenvectors in Davidson algorithm are saved in the object.

pyscf/gw/bse.py
lijiachen417: Thanks for pointing out this bug. Fixed.

examples/gw/04-bse.py
MatthewRHermes, line 1: Is it possible to add reference results from the literature of from another program to this example so as to validate the end-to-end correctness of the implementation?

lijiachen417: > Is it possible to add reference results from the literature of from another program to this example so as to validate the end-to-end correctness of the implementation?

For the acetone molecule (geometry from [this paper](https://pubs.acs.org/doi/full/10.1021/acs.jpclett.5b02773)), B3LYP/def2-svp.

In the GW step, using fully analytic GW method, quasiparticle equation solved iteratively
- Turbomole HOMO = -8.78 eV LUMO = 2.75 eV
- PySCF HOMO = -8.79 eV LUMO = 2.75 eV

In the BSE step
first three singlet excitation in eV
- Turbomole = 3.41 7.35 8.56 
- PySCF = 3.41 7.36 8.58

first three triplet excitation in eV
- Turbomole = 2.68 4.67 7.14
- PySCF = 2.67 4.67 7.14 

Following the examples for GW and other modules, I didn't put those benchmark values from other software in the bse example file.

lijiachen417: > > Is it possible to add reference results from the literature of from another program to this example so as to validate the end-to-end correctness of the implementation?
> 
> For the acetone molecule (geometry from [this paper](https://pubs.acs.org/doi/full/10.1021/acs.jpclett.5b02773)), B3LYP/def2-svp.
> 
> In the GW step, using fully analytic GW method, quasiparticle equation solved iteratively
> 
> * Turbomole HOMO = -8.78 eV LUMO = 2.75 eV
> * PySCF HOMO = -8.79 eV LUMO = 2.75 eV
> 
> In the BSE step first three singlet excitation in eV
> 
> * Turbomole = 3.41 7.35 8.56
> * PySCF = 3.41 7.36 8.58
> 
> first three triplet excitation in eV
> 
> * Turbomole = 2.68 4.67 7.14
> * PySCF = 2.67 4.67 7.14
> 
> Following the examples for GW and other modules, I didn't put those benchmark values from other software in the bse example file.

@sunqm These reference values are for acetone, not h2o.
I will open a PR for [this line](https://github.com/lijiachen417/pyscf/blob/633f3ce5e17f7aa8fe6f31be229deef541fc2def/examples/gw/04-bse.py#L5) in the example

Issue #2102, opened 2024-02-26
Removing the basis data that are duplicated to bse


sunqm:
There are several issues related to the missing basis and utilizing bse database in pyscf. Pyscf right now keeps some basis data that are available in bse. Shall we clean up the basis data in pyscf, and use bse more? This consideration was initialized in #898, but not fully solved there. Let's continue the discussion here.

The main concern is the consistency between the results obtained in old versions of pyscf and the one using bse basis data. Some basis data downloaded from the basis set exchange website in the early days are slightly different to the current bse database. If we want to clean up the basis data, we probably need to decide how to handle these differences.

After scanning the entire basis data in pyscf and comparing to the bse output, I find the following categories of differences
* Completely the same. These basis sets can be deleted, I believe.
```
pc-0 pc-1 pc-2 pc-3 pc-4 def2-universal-JFIT CRENBL LANL08 6-311++G 6-311++G*
6-311++G** 6-311+G 6-311+G* 6-311+G** 6-311G 6-311G* 6-311G** Stuttgart-RLC
aug-pc-0 aug-pc-1 aug-pc-2 aug-pc-3 aug-pc-4 pcseg-0 pcseg-1 pcseg-2 pcseg-3
pcseg-4 aug-pcseg-0 aug-pcseg-1 aug-pcseg-2 aug-pcseg-3 aug-pcseg-4
cc-pwCVQZ-DK3 cc-pwCVTZ-DK3 aug-cc-pVQZ-DK3 aug-cc-pVTZ-DK3 aug-cc-pwCVQZ-DK3
aug-cc-pwCVTZ-DK3 cc-pVTZ-DK3 cc-pVQZ-DK3 cc-pV5Z-DK3 cc-pV5Z-F12 cc-pVTZ-JKFIT
cc-pVQZ-JKFIT cc-pV5Z-JKFIT cc-pwCV5Z-DK cc-pwCVQZ-DK cc-pwCVTZ-DK
```

* Small difference due to rounding errors. The errors are likely caused by the single precision data we obtained in the past. I think these basis should be safe to drop. def2 basis is a little bit problematic. The basis for Ba and Mn are different.
```
STO-3G STO-6G 3-21G 4-31G 6-31++G 6-31++G* 6-31++G** 6-31+G 6-31+G* 6-31+G** 6-31G 6-31G* 6-31G**
[...]

PR #3238, opened 2026-06-05
Add dgauss-tzvp basis set

egreenberg7:
PySCF currently has `dgauss-dzvp` (under alias `"dzvp"`) and `dgauss-dzvp2` (alias `dzvp2`) as basis options, but `dgauss-tzvp` was not available. This PR simply adds it in from the basis set exchange, using the alias `tzvp` to be consistent with the other options.


jeanwsr:
I think there's no need to add basis sets that are already on BSE, since we have BSE api and automatic fallback to querying BSE when it's installed. See #2085 #2102.

PR #3239, opened 2026-06-05
RPA improvements: PBC RPA routines

lijiachen417:
This is the 5th of several updates to the GW/RPA code.
This PR adds k-point RPA routines.

Contributors: Tianyu Zhu (@zhutianyu1991 ), Jiachen Li (@lijiachen417 ), Chris Hillenbrand (@chillenb ), Chaoqun Zhang (@Warlocat ), Jincheng Yu (@pi246 )

* KRPA/KURPA: new routine for spin-restricted/unrestricted k-point RPA
* support low-memory mode
* support smeared systems
* support ACFDT exchange for smeared systems
* add test files and example

### DIFF
--- a/examples/pbc/22-k_points_rpa.py
+++ b/examples/pbc/22-k_points_rpa.py
@@ -1,0 +1,114 @@
+'''
+RPA with k-points sampling
+'''
+
+from pyscf.pbc import gto, df, dft, scf
+from pyscf.pbc.gw.krpa import KRPA
+from pyscf.pbc.gw.kurpa import KURPA
+
+# spin-restricted RPA
+cell = gto.Cell()
+cell.build(
+    unit='angstrom',
+    a="""
+            0.000000     1.783500     1.783500
+            1.783500     0.000000     1.783500
+            1.783500     1.783500     0.000000
+        """,
+    atom='C 1.337625 1.337625 1.337625; C 2.229375 2.229375 2.229375',
+    dimension=3,
+    max_memory=12000,
+    verbose=5,
+    pseudo='gth-pbe',
+    basis='gth-dzv',
+    precision=1e-12,
+)
+
+kpts = cell.make_kpts([3, 1, 1], scaled_center=[0, 0, 0])
+gdf = df.RSGDF(cell, kpts)
+gdf.build()
+kmf = scf.KRHF(cell, kpts).rs_density_fit()
+kmf.with_df = gdf
+kmf.kernel()
+
+# RPA with finite-size correction
+rpa = KRPA(kmf)
+rpa.fc = True
+rpa.kernel()
+# RPA with finite-size correction
+rpa = KRPA(kmf)
+rpa.fc = False
+rpa.kernel()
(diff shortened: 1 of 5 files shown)


lijiachen417:
Updating molecular RPA and adding Gamma-point RPA routines will overwrite the current molecule RPA code, so they are not in this PR.

pyscf/pbc/gw/krpa.py
```diff
@@ -0,0 +1,996 @@
...
+        logger.debug(rpa, f'  Setting EXX energy explicitly to {e_hf}')
+
+    is_metal = hasattr(rpa._scf, 'sigma')
+
+    # Turn off FC for metals and outcore
+    if is_metal and rpa.fc:
+        logger.warn(rpa, 'FC not available for metals - setting rpa.fc to False')
+        rpa.fc = False
+    if rpa.outcore and rpa.fc:
+        logger.warn(rpa, 'FC not available for outcore - setting rpa.fc to False')
```
sunqm: Why cannot the finite-size correction be applied to the outcore algorithm?

If the finite-size correction is not well defined for smeared occupations, or just not implemented, it may be more appropriate to raise an error rather than silently disabling it.

sunqm (approved):

pyscf/pbc/gw/kurpa.py
```diff
@@ -0,0 +1,787 @@
...
+
+    is_metal = hasattr(rpa._scf, 'sigma')
+
+    # Turn off FC for metals and outcore
+    if is_metal and rpa.fc:
+        logger.warn(rpa, 'FC not available for metals - setting rpa.fc to False')
+        rpa.fc = False
+    if rpa.fc and rpa.outcore:
+        logger.warn(rpa, 'FC not available for outcore - setting rpa.fc to False')
+        rpa.fc = False
```
sunqm: Update these fc treatments if needed, see the comments for krpa

pyscf/pbc/gw/krpa.py
lijiachen417: Finite-size correction for the outcore routine is implemented now.

pyscf/pbc/gw/kurpa.py
lijiachen417: Finite-size correction for the outcore routine is implemented now.

examples/pbc/22-k_points_rpa.py
MatthewRHermes, line 1: Is it possible to add reference values for the result of these calculations from the literature or from another program?

lijiachen417: > Is it possible to add reference values for the result of these calculations from the literature or from another program?

From PBC routine, I guess it's hard to setup a simple example to compare with other packages. Here I just followed examples for [PBC MP2](https://github.com/lijiachen417/pyscf/blob/262a11448e781a1d0ea4eb5acffb3ee1b2dce34f/examples/pbc/22-k_points_mp2.py) and other PBC modules.

PR #3241, opened 2026-06-06
use dll wrappers on windows

jeanwsr:
corresponds to patch 1, 2 and 4 of https://github.com/conda-forge/pyscf-feedstock/pull/63, and passes the win-64 tests there.

1. DLL wrapping on Windows
On Unix, ctypes' getattr(dll, name) searches a flat process-wide symbol namespace, which means a function compiled into libcint.so is visible when accessing libcgto.so. On Windows, it is not visible. The fix: whenever a library like libcgto is loaded on Windows, it's wrapped via make_dll_wrapper() with its dependency libraries as fallbacks for symbol lookup. For example, pyscf/gto/moleintor.py now wraps libcgto with libcint as fallback, and pyscf/dft/numint.py wraps libdft with libcvhf, libcgto, libcint as fallbacks.
UPDATE: the wrap is now automatically considered in load_library.
2. `_ctypes.dlsym` replaced with `ctypes.cast(getattr(...))` 
`_ctypes.dlsym(handle, name)` is a CPython internal that resolves symbols by name from a raw dlopen handle. It exists on Linux/macOS but Windows has no equivalent.
The fix: `ctypes.cast(getattr(lib, name), ctypes.c_void_p)` achieves the same result by accessing the function via the wrapper's `__getattr__`, then casting the returned function pointer to c_void_p. 
~~3. Reusing wrapped library objects instead of reloading 
Some modules were calling lib.load_library() again on their own, creating a separate unwrapped copy. Fixed by importing the already-wrapped object from the canonical module, e.g., pyscf/ao2mo/nrr_outcore.py now imports libao2mo from _ao2mo instead of reloading.~~
~~4. Tests fixed to use wrapped libraries
Five test files were calling lib.load_library() directly. Fixed by importing from the module that already provides the wrapped library.~~
5. Platform-agnostic memory reporting in lib/misc.py
`current_memory()` used Linux-only `/proc`. Fixed by adding a psutil fallback. Also added conda-typical library search paths to load_library() for Windows.
6. Update cmake options for win-64 conda-forge build (gnu only).


pyscf/ao2mo/_ao2mo.py
```diff
@@ -14,15 +14,21 @@
...
 import ctypes
-import _ctypes
+import sys
 import numpy
 from pyscf import lib
 from pyscf.gto.moleintor import make_cintopt, make_loc, ascint3
 from pyscf.scf import _vhf
 
 libao2mo = lib.load_library('libao2mo')
+
```
sunqm, line 23: The `load_library` function can be extended with an optional keyword arg, such as `dependencies` or `fallbacks`. When this keyword argument is provided, OS-specific checks and patching logic can be handled directly in `load_library`. No need to duplicate OS-related code in each file.

`load_library` is already cached, moving the Windows patching code also avoid creating multiple copies.

pyscf/lib/misc.py
```diff
@@ -106,21 +113,41 @@ def load_library(libname):
...
+                    return getattr(dll, name)
+                except AttributeError:
+                    pass
+            raise AttributeError(f"function '{name}' not found")
+    return _DllWrapper(lib, *fallbacks)
+
 #Fixme, the standard resource module gives wrong number when objects are released
 # http://fa.bianp.net/blog/2013/different-ways-to-get-memory-consumption-or-lessons-learned-from-memory_profiler/#fn:1
 #or use slow functions as memory_profiler._get_memory did
-CLOCK_TICKS = os.sysconf("SC_CLK_TCK")
```
sunqm, line 112: create the two variables in the module space. In the case windows, set them to None

sunqm:
please also update pyproject.toml, to include the dependency
"psutils; sys_platform == 'win32'"

pyscf/lib/misc.py
jeanwsr: ok, but is CLOCK_TICKS used anywhere?

pyscf/ao2mo/_ao2mo.py
jeanwsr: updated. now load_library will infer what to add as fallbacks.

pyscf/lib/misc.py
sunqm: Not in the PySCF package. No harm to keep it there in case it's used by other code built on top of PySCF for now. In the future, it can be cleaned up when an explicit `__all__` or explicit documentation for these parameters are introduced.

sunqm (approved):

PR #3243, opened 2026-06-09
Fixed KeyError for MM ECPs in project_to_atomic_orbitals

ndolin:
When using ECPs with nelec=0 in QM/MM embedding applications, `mf.analyze()` and related functions would break with a KeyError. This fix skips over those ECPs, but should not affect regular ECPs with nelec != 0.


jeanwsr:
Can you add a test?

sunqm (approved):

PR #3244, opened 2026-06-09
GW improvements: self-consistent PBC GW routines

lijiachen417:
This is the 6th of several updates to the GW/RPA code from the [Zhu lab](https://github.com/ZhuGroup-Yale)

* KREVGW/KUEVGW: spin-restricted/unrestricted eigenvalue self-consistent GW
* KRQSGW/KUQSGW: spin-restricted/unrestricted quasiparticle self-consistent GW
* examples and test files added

Contributors: Tianyu Zhu (@zhutianyu1991 ), Jiachen Li (@lijiachen417 ), Chris Hillenbrand (@chillenb )

### DIFF
--- a/pyscf/pbc/gw/krevgw.py
+++ b/pyscf/pbc/gw/krevgw.py
@@ -1,0 +1,271 @@
+#!/usr/bin/env python
+# Copyright 2014-2026 The PySCF Developers. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Author: Tianyu Zhu <zhutianyu1991@gmail.com>
+# Author: Christopher Hillenbrand <chillenbrand15@gmail.com>
+# Author: Jiachen Li <lijiachen.duke@gmail.com>
+#
+
+"""
+Periodic spin-restricted eigenvalue self-consistent GW method based on the analytic continuation scheme.
+
+References:
+    Phys. Rev. B 76, 165106 (2007)
+    J. Chem. Theory. Comput. 12, 2528-2541 (2016)
+"""
+
+import scipy
+import numpy as np
+
+from pyscf import lib
+from pyscf.lib import logger, temporary_env
+from pyscf.pbc import df, scf
+
+from pyscf.gw.utils.ac_grid import _get_scaled_legendre_roots, PadeAC, TwoPoleAC
+from pyscf.pbc.gw.krgw_ac import KRGWAC, get_sigma, get_ef, _mo_energy_frozen
+
+
+def kernel(gw):
+    mf = gw._scf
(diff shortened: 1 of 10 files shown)


jeanwsr:
Some new test added here needs over 200s in actions. Can you mark it as `_high_cost` like https://github.com/pyscf/pyscf/blob/4330efa7bcb607801bc9db53585b1048996bc7df/pyscf/pbc/adc/test/test_k_ea.py#L67 or reduce the duration.

lijiachen417:
> Some new test added here needs over 200s in actions. Can you mark it as `_high_cost` like
> 
> https://github.com/pyscf/pyscf/blob/4330efa7bcb607801bc9db53585b1048996bc7df/pyscf/pbc/adc/test/test_k_ea.py#L67
> 
> or reduce the duration.

Thanks for suggesting this 👍I marked those 4 test files as high cost.

pyscf/pbc/gw/krevgw.py
```diff
@@ -0,0 +1,271 @@
...
+        ef = gw.ef = get_ef(mf, mo_energy=mo_energy)
+        gw.acobj = acobj
+
+        # update NON-FROZEN quantities
+        mo_energy_frz = _mo_energy_frozen(gw, mo_energy)
+
+        # check density matrix convergence
+        diff = 0
+        for k in range(nkpts):
+            diff += abs(np.sum(1.0 / mo_energy[k] - 1.0 / mo_energy_old[k]))
```
sunqm, line 149: why not using a simple convergence check |mo_energy - mo_energy_old| < threshold ?

pyscf/pbc/gw/krqsgw.py
```diff
@@ -0,0 +1,343 @@
...
+        if cycle == 0:
+            dm_old2 = dm_iter
+        else:
+            dm_old2 = dm_old
+        dm_old = dm_iter
+        # density matrix from updated QSGW density
+        dm_iter = rhf.make_rdm1()
+        norm_dm = np.linalg.norm(dm_iter - dm_old) / (nmo * nkpts)
+        norm_dm2 = np.linalg.norm(dm_old - dm_old2) / (nmo * nkpts)
+        if norm_dm < gw.conv_tol and norm_dm2 < gw.conv_tol and cycle > 0:
```
sunqm, line 210: Is the dm sensitive to the basis set, particularly for the basis sets with diffuse functions? Does it make sense to only check convergence against the change of energy?

sunqm (approved):

pyscf/pbc/gw/krevgw.py
lijiachen417: We followed the convention as equation 19 in this [paper](https://pubs.acs.org/doi/full/10.1021/acs.jctc.5b01238) which uses the difference of the Green's function at the static limit. Yes for evGW, it's effectively similar to what you proposed. Do you prefer to switch to "|mo_energy - mo_energy_old| < threshold"? I can make the change.

pyscf/pbc/gw/krqsgw.py
lijiachen417: > Is the dm sensitive to the basis set, particularly for the basis sets with diffuse functions? Does it make sense to only check convergence against the change of energy?

There are two reason the energy was not used to check convergence:

1. QSGW energy is not commonly used and doesn't have a clear definition.
2. For self-energy evaluation using analytic continuation, the energy value is highly sensitive to grids and numerically unstable.

So we followed the equation 19 in this [paper](https://pubs.acs.org/doi/full/10.1021/acs.jctc.5b01238). For QSGW, the Green's function at the static limit is effectively the density matrix.

Issue #3245, opened 2026-06-09
[UNEXPECTED BEHAVIOR] Frequently failing tests in github actions


jeanwsr:
**Describe the unexpected behavior**
quite frequent ones
- [x] adc, gw
```
FAILED pyscf/adc/test/test_radc/test_ee_df_N2.py::KnownValues::test_ee_adc2 - AssertionError: 40.69388871108569 != 40.69394840350379 within 4 places (5.969241809822279e-05 difference)
FAILED pyscf/gw/test/test_gw.py::KnownValues::test_gwac_pade_frozen - AssertionError: -0.41294112137807376 != -0.4129411145067107 within 8 places (6.871363034122169e-09 difference)
```
- [x] mcpdft
```
SUBFAILED(case='SA', state=0, symm=False, triplet_ms=0) pyscf/grad/test/test_mcpdft.py::KnownValues::test_scanner - AssertionError: 0.011270954839227887 != 0.011235583377960235 within 1e-05 delta (3.5371461267651866e-05 difference)
SUBFAILED(case='SA', state=1, symm=False, triplet_ms=0) pyscf/grad/test/test_mcpdft.py::KnownValues::test_scanner - AssertionError: 0.07317669967280033 != 0.0730926778904642 within 1e-05 delta (8.402178233613e-05 difference)
SUBFAILED(case='SA', state=2, symm=False, triplet_ms=0) pyscf/grad/test/test_mcpdft.py::KnownValues::test_scanner - AssertionError: 0.07217643986901016 != 0.07208382527063625 within 1e-05 delta (9.261459837391073e-05 difference)
SUBFAILED(case='SA', state=3, symm=False, triplet_ms=0) pyscf/grad/test/test_mcpdft.py::KnownValues::test_scanner - AssertionError: False is not true
SUBFAILED(case='SA', state=4, symm=False, triplet_ms=0) pyscf/grad/test/test_mcpdft.py::KnownValues::test_scanner - AssertionError: False is not true
FAILED pyscf/grad/test/test_mcpdft.py::KnownValues::test_scanner - AssertionError: 0.011270954836072098 != 0.011235583353900637 within 1e-05 delta (3.53714821714618e-05 difference)
SUBFAILED(case='SA', state=0, symmetry=False, triplet_ms=0) pyscf/grad/test/test_mcpdft.py::KnownValues::test_gradients - AssertionError: 0.005646042946774701 != 0.00566392595 within 5 places (1.7883003225299156e-05 difference)
[...]

PR #3247, opened 2026-06-10
Loosen some tests in adc, gw and tdscf

jeanwsr:
Resolve part of #3245 . Did not touch mcpdft and the suspicious ones in which the error is not small enough (like test_eom_gccsd.py::KnownValues::test_ipccsd, test_rks.py::Diamond::test_hse06_tda, etc.)

Also mute a few high cost tests.


lijiachen417:
Thanks for doing this! Because of the analytical continuation, tests in GWAC cannot be as accurate as 1e-6 or sometimes 1e-5

sunqm (approved):

PR #3255, opened 2026-06-12
High order cc improvements

jinyuchem:
Dear PySCF developers,

This PR improves the high-order coupled-cluster implementations and adds new perturbative quadruples corrections, RCCSDT[Q] and RCCSDT(Q), on top of RCCSDT.

New feature:

- Add [Q]/(Q) correction implementations in `rccsdt_q.py`, with the relevant C functions added in `rccsdtq.c`, together with unit tests and examples. The implementation follows the framework of [J. Chem. Phys. 142, 064108 (2015)], and can be used with both the compact T3 implementation in `rccsdt.py` and the full T3 implementation in `rccsdt_highm.py`.

Improvements:

- Unify intermediate construction for low- and high-order amplitude updates in RCCSDT, RCCSDTQ, and UCCSDT. For example, the `W_ovvo` intermediate is now reused in `rccsdt.py` for both T2 and T3 updates.
- Refactor the code to reduce redundant unpacking operations and tensor contractions, which slightly improves the performance.
- Replace the previous partial projector for T4 amplitudes in RCCSDTQ with a complete orthogonalization projector to remove linearly dependent components, which improves convergence robustness in difficult cases.
- Update all amplitudes synchronously at the end of each iteration. This change does not affect converged energies, but it can change early-iteration energy values and the number of iterations required for convergence. As a result, the reference values in affected iteration-specific tests have been updated accordingly.
- Fix typos and minor code/documentation issues.

### DIFF
--- a/examples/cc/66-rccsdt_q.py
+++ b/examples/cc/66-rccsdt_q.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python
+#
+# Author: Yu Jin <yjin@flatironinstitute.org>
+#
+
+'''
+Examples of RCCSDT(Q) calculations.
+
+This script demonstrates:
+    - Consistency of the [Q] and (Q) energy corrections on top of RCCSDT between calculations
+        using full and compact T3 storage.
+'''
+
+from pyscf import gto, scf, cc
+
+mol = gto.M(atom='H 0 0 0; F 0 0 1.1', basis='ccpvdz')
+mf = scf.RHF(mol)
+mf.conv_tol = 1e-14
+mf.kernel()
+
+# Reference CCSDT correlation energy, and [Q] and (Q) energy correction
+ref_e_corr = -0.2188784733230733
+ref_e_q_bracket = -0.0005026220700017348
+ref_e_q_paren = -0.0005490746450078632
+
+mycc1 = cc.RCCSDT(mf, compact_tamps=True)
+mycc1.conv_tol = 1e-10
+mycc1.conv_tol_normt = 1e-8
+mycc1.verbose = 5
+# einsum_backend: numpy (default) | pyscf | pytblis (recommended)
+# pytblis can be installed via `pip install pytblis==0.05` (See https://github.com/chillenb/pytblis)
+mycc1.set_einsum_backend('pyscf')
+mycc1.incore_complete = True
+mycc1.kernel()
+e_q_bracket, e_q_paren = mycc1.ccsdt_q()
+print('Triangular RCCSDT e_corr % .12f    Ref % .12f    Diff % .12e' % (
+        mycc1.e_corr, ref_e_corr, mycc1.e_corr - ref_e_corr))
+print('Triangular RCCSDT [Q]    % .12f    Ref % .12f    Diff % .12e' % (
+        e_q_bracket, ref_e_q_bracket, e_q_bracket - ref_e_q_bracket))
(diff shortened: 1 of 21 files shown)


pyscf/cc/rccsdt_q.py
```diff
@@ -0,0 +1,313 @@
...
+    e_q_paren += e_q_bracket
+    e_q_bracket /= 12.0
+    e_q_paren /= 12.0
+
+    log.timer('%s(Q)' % name, *time0)
+    log.info("[Q] correction = % .12e    (Q) correction = % .12e" % (e_q_bracket, e_q_paren))
+    return e_q_bracket, e_q_paren
+
+
+if __name__ == '__main__':
```
jeanwsr, line 253: can this part be moved to test? or remove if it's already covered by test.

pyscf/cc/rccsdt_q.py
```diff
@@ -0,0 +1,313 @@
...
+
+                    # t4
+                    W_vvoooo_ijk = compute_W_vvoooo(i, j, k)
+                    W_vvoooo_ijl = compute_W_vvoooo(i, j, l)
+                    W_vvoooo_ikj = compute_W_vvoooo(i, k, j)
+                    W_vvoooo_ikl = compute_W_vvoooo(i, k, l)
+                    W_vvoooo_ilj = compute_W_vvoooo(i, l, j)
+                    W_vvoooo_ilk = compute_W_vvoooo(i, l, k)
+                    W_vvoooo_jki = compute_W_vvoooo(j, k, i)
+                    W_vvoooo_jkl = compute_W_vvoooo(j, k, l)
```
jeanwsr: is this re-computed many times since it's independent of i.

jeanwsr (approved):

pyscf/cc/rccsdt_q.py
jinyuchem: This block is mainly kept as a simple runnable example for quick testing/debugging, similar in spirit to other existing PySCF MP/CC files such as https://github.com/pyscf/pyscf/blob/master/pyscf/mp/mp2.py#L928.

pyscf/cc/rccsdt_q.py
jinyuchem: The part is designed to save memory cost for intermediates. In the latest commit, we have rewritten this part to reuse the same W_vvoooo buffer across all index combinations, further reducing memory cost.

examples/cc/66-rccsdt_q.py
```diff
@@ -0,0 +1,59 @@
...
+
+mol = gto.M(atom='H 0 0 0; F 0 0 1.1', basis='ccpvdz')
+mf = scf.RHF(mol)
+mf.conv_tol = 1e-14
+mf.kernel()
+
+# Reference CCSDT correlation energy, and [Q] and (Q) energy correction
+ref_e_corr = -0.2188784733230733
+ref_e_q_bracket = -0.0005026220700017348
+ref_e_q_paren = -0.0005490746450078632
```
MatthewRHermes, line 24: Can you clarify the source of these references? Do they come from another program?

jinyuchem: The reference values were obtained from an independent implementation by @hczhai, which is based entirely on automatic equation derivation.

MatthewRHermes: OK can you just put that in commentary in the example script for future reference?

PR #3256, opened 2026-06-12
add AOCache for pbc fftdf dft gradient and response function

Warlocat:
By profiling our recent implementation on post-scf gradient for PBC systems, I found a huge waste of time in re-evaluating the same AOs on the same grid many times. This includes 
1. the response function in CPKS iterations 
2. a redundant re-evaluation in hcore_deriv
3. loop over aoR_loop/block_loop in fftdf

This PR added an AOCache to save the precomputed ao grid values which is a relatively small intermediate, and reuse it in pbc dft response functions and grad.krhf get_veff calculations.

Codes were written by AI and reviewed by myself.

### DIFF
--- a/pyscf/pbc/grad/krhf.py
+++ b/pyscf/pbc/grad/krhf.py
@@ -129,13 +129,17 @@ def hcore_generator(mf_grad, cell=None, kpts=None):
     Gv = cell.Gv    ##[grid, 3]
     coords = cell.get_uniform_grids()
     vlocG = get_vlocG(cell)  ###[natom, grid]
+    # The local-PP AO product depends only on (coords, kpt), not on the atom
+    # or Cartesian component, so evaluate it once per k-point instead of once
+    # per (atom, k-point) inside the closure below.
+    ao_kpts = eval_ao_kpts(cell, coords, kpts)
     def hcore_deriv(atm_id):
         shl0, shl1, p0, p1 = aoslices[atm_id]
         vloc_g = 1j * np.einsum('ga,g->ag', Gv, SI[atm_id]*vlocG[atm_id])
         nkpts, nao = h1.shape[0], h1.shape[2]
         hcore = np.zeros([3,nkpts,nao,nao], dtype=h1.dtype)
         for kn, kpt in enumerate(kpts):
-            ao = eval_ao_kpts(cell, coords, kpt)[0]
+            ao = ao_kpts[kn]
             for ax in range(3):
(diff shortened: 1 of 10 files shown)


jeanwsr:
1. I think ao is not cached in most codes (like molecular dft) simply because the memory is insufficient to do so. Why it's "relatively small intermediate" in your implementation, on how large system can the cache survive with normal memory?
2. Can we just register the ao cache under ni, or some pbc object, to avoid pass it in many functions.

Warlocat:
1. I was estimating for smaller system and compare to our pp-RPA or other wave function intermediates, not for mean-field calculations. Thanks for the comment. The AOCache might only be useful for a tddft/rpa level gradients.
2. I will improve the implementation and make it more compact.

Warlocat:
I have modified the implementation. AOCache will not be used by default, even in the gradient calculations.

pyscf/pbc/grad/krhf.py
```diff
@@ -129,13 +129,17 @@ def hcore_generator(mf_grad, cell=None, kpts=None):
     Gv = cell.Gv    ##[grid, 3]
     coords = cell.get_uniform_grids()
     vlocG = get_vlocG(cell)  ###[natom, grid]
+    # The local-PP AO product depends only on (coords, kpt), not on the atom
+    # or Cartesian component, so evaluate it once per k-point instead of once
+    # per (atom, k-point) inside the closure below.
+    ao_kpts = eval_ao_kpts(cell, coords, kpts)
```
MatthewRHermes, line 135: pretty sure the reason we did it the other way is that ao @ all k-points simultaneously is too memory-hungry for some applications

MatthewRHermes: and anyway the cache itself addresses the time inefficiency so why are you also doing this?

Warlocat: Sorry for underestimating the memory cost.
For the aocahe, it is part of the NumInt. It will be messier to let krhf's hcore_generator use it. I can make a separate hcore_generator for krks gradient if that is accepted.

sunqm:
Thanks for the proposing this optimization. However, caching AO values can require a huge amount of memory, easily cause OOM termination.

AO evaluation is generally not the bottleneck. The current AO evaluation is not well optimized. I'd recommend optimizing the AO evaluation code and making use of sparsity for this issue than caching AO values.

I’ll close this PR for now.

PR #3260, opened 2026-06-15
Added self-consistent dipole corrections for slab/2D systems

afarahva:
This PR adds self-consistent dipole corrections for slab/2D systems. Effectively implementing the equivalent of LDIPOL/IDIPOL in VASP. I included a unit test and an example for completeness.

### DIFF
--- a/examples/pbc/49-slab_dipole_correction.py
+++ b/examples/pbc/49-slab_dipole_correction.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+#
+# Author: Ardavan Farahvash <ardf.scf@gmail.com>
+#
+'''
+Benggston-Neugebauer-Scheffler Dipole Corrections for surface calculations. 
+
+
+Removes the leading-order error of polarized surfaces caused by the spurious
+electrostatic interaction between periodic slab surface and its normal images. 
+
+'''
+
+import numpy as np
+from pyscf.pbc import gto, dft
+from pyscf.pbc.scf.addons import slab_dipole_correction
+
+# Create a highly polar HF molecule unit cell
+atom = "F 0.00 0.00 0.00; H 0.00 0.00 0.92"
+cell = gto.Cell()
+cell.atom = atom
+cell.basis = "gth-szv"
+cell.pseudo = "gth-pade"
+cell.unit = "A"
+cell.dimension = 3 
+cell.verbose = 0
+
+# Run large calculation  
+print("--------- Running Huge Vacuum ---------")
+cell.a = np.diag([10, 10, 200])
+cell.build()
+
+mf = dft.KUKS(cell, cell.make_kpts([1, 1, 1])).density_fit()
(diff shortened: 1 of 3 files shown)


examples/pbc/49-slab_dipole_correction.py
```diff
@@ -0,0 +1,61 @@
...
+for d in np.arange(5,18,3):
+    cell.a = np.diag([10, 10, d])
+    cell.build()
+
+    mf = dft.KRKS(cell, cell.make_kpts([1, 1, 1])).density_fit()
+    mf = slab_dipole_correction(mf, dir_idx=2)
+    mf.xc = "pbe"
+    mf.kernel()
+    e = mf.kernel()-e0
+    print(f"Corrected d={d}, E-E(d=200)={e:.6f} Ha")
```
MatthewRHermes, line 61: Please add what these numbers are "supposed" to be (on your setup, if nothing else; but a literature reference or other-program reference is even better) so users can see if something is off.

pyscf/pbc/scf/test/test_slabdipole.py
```diff
@@ -0,0 +1,58 @@
...
+        """A polar slab has a sizeable dipole and the correction shifts E."""
+        de = self._run_pair(polar=True)
+        self.assertGreater(abs(de), 1e-5,
+                           "correction had no effect on a polar slab")
+
+    def test_nonpolar_slab(self):
+        """A symmetric slab has ~zero dipole and the correction is ~zero."""
+        de = self._run_pair(polar=False)
+        self.assertLess(abs(de), 1e-4,
+                        "correction should be ~zero for a non-polar slab")
```
MatthewRHermes, line 54: These tests don't validate the _value_ of the corrected energy, which is something I think that you should try to do.

pyscf/pbc/scf/addons.py
```diff
@@ -377,3 +377,100 @@ def _get_moek(kC, kfao):
...
+        ion_shifted = wrap_dir(cell.atom_coords())
+        p_ion = numpy.sum(charges * ion_shifted)
+        p_tot = p_ion - p_elec
+        # Artificial PBC field and compensating sawtooth potential.
+        e_field = 4 * numpy.pi * p_tot / (area * L)
+        v_dip_grid = -e_field * z_shifted * weights
+        # TODO: cache AO values on the grid
+        ao_kpts = cell.pbc_eval_gto("GTOval", coords, kpts=kpts)
+        v_dip_mat = numpy.asarray(
+            [lib.einsum("pi,p,pj->ij", ao.conj(), v_dip_grid, ao) for ao in ao_kpts])
```
jeanwsr, line 462: Is it possible to do this with analytical integrals?

sunqm:
The implementation is a bit messy, but the approach is valid. We can merge this as is and refactor it in a future PR.

sunqm (approved):

PR #3263, opened 2026-06-15
MC-PDFT unittests stabilize (#3245)

MatthewRHermes:
In MC-PDFT unittests which perform a variety of calculations on LiH, SA-MCSCF calculations sometimes converge to different solutions. Try to control this by initializing all calc'ns with converged orbitals from high-symmetry version in setup.


jeanwsr (approved):
Thanks, LGTM!

Issue #3264, opened 2026-06-16
[FEATURE REQUEST] CABS & DBBSC corrections following by 10.1002/jcc.27325


foxtran:
At the current moment, PySCF gives an opportunity to compute MP2-F12, but -F12 is computationally heavy. Fortunately, we have a cheaper, but little bit imprecise, DBBSC correction. So, it would be nice to see them in PySCF.

**Is your feature request related to a scientific problem? Please describe.**
I want to compute MP2 in small basis sets and fix basis set incompleteness  via DBBSC correction.

**Additional context**
There is two main papers about DBBSC (and CABS):

- https://doi.org/10.1002/jcc.27325 by Andreas Heßelmann, Emmanuel Giner, Peter Reinhardt, Peter J. Knowles, Hans-Joachim Werner, and Julien Toulouse
- https://doi.org/10.1021/acs.jpclett.5c00122 by Dávid Mester, and Mihály Kállay

**NOTE:** I have implemented CABS + DBBSC, but need some time to clean & improve code.

PR #3265, opened 2026-06-16
Implement CABS Singles correction

foxtran:
This patch adds CABS Singles correction according to https://www.molpro.net/manual/doku.php?id=explicitly_correlated_methods#cabs_singles_correction

Tests were verified by comparing values from MRCC & MolPro. See basis setup & numbers in tests.

Refs:
- JCC 127, 221106 (2007); DOI:10.1063/1.2817618
- JCP 128, 154103 (2008); DOI:10.1063/1.2889388
- JCC 45, 1247-1253 (2024); DOI:10.1002/jcc.27325

Related to #3264

### DIFF
--- a/pyscf/mp/cabs.py
+++ b/pyscf/mp/cabs.py
@@ -24,7 +24,8 @@ Refs:
 import numpy
 import scipy.linalg
 
-from pyscf import gto
+from pyscf import gto, scf
+from pyscf.data import elements
 from pyscf.lib import logger
 from pyscf.scf import hf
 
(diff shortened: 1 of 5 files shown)


jeanwsr:
Still not close enough to MRCC/Molpro result as a HF job, is it?

foxtran:
@jeanwsr, MRCC & MolPro are also not consistent... May be cc-pVDZ basis is not consistent between different programs. 

I'll try with other basis sets.

foxtran:
Ah... The problem is with core electrons.

foxtran:
So, the problems were:
- improper usage of Density Fitting
- frozen core approximation(s)
- UHF / ROHF reference comparison for MolPro

jeanwsr:
Looks like CABS is designed to be used with MP2. If so, can you add an example for the whole MP2+CABS calculation.

foxtran:
This correction is useful with any post-HF methods. I'll add it, give me 1-2 hours.

pyscf/mp/cabs.py
```diff
@@ -0,0 +1,245 @@
...
+    the virtual MOs of the orbital basis and CABS. For UHF and ROHF references
+    the same expression is evaluated for the alpha and beta Fock matrices with
+    spin occupations as prefactors.
+
+    Args:
+        mf : SCF object
+            Converged molecular HF object.
+        auxbasis : Mole, str, list, tuple, or dict
+            CABS/OptRI basis as a Mole object or in the usual Mole.basis format.
+            If a normal charged Mole is supplied on the same centers as ``mf``.
```
jeanwsr: It seems this is not a complete sentence?

jeanwsr (approved):

pyscf/mp/cabs.py
foxtran: Unfortunately, yes. Looks like I deleted it accidentally while rechecking what I wrote.

PR #3271, opened 2026-06-18
symmetry breaking via homo-lumo-mixing

lynn-buehl:
Summary

This PR adds a new option breaksym=2 to _break_dm_spin_symm and the UHF init_guess_breaksym setting. It provides an alternative way to break spin symmetry in the initial density matrix for closed-shell starting points.

What it does

Instead of zeroing the off-diagonal AO blocks of the beta density matrix (breaksym=1), breaksym=2 applies a 45-degree rotation between the HOMO and LUMO:

α spin HOMO → (HOMO + LUMO) / √2
β spin HOMO → (HOMO − LUMO) / √2
This mixes the two frontier orbitals differently for each spin, breaking the spin symmetry while keeping both orbitals fully delocalized over the molecule.

Why this is better than breaksym=1

breaksym=1 zeros all off-diagonal AO blocks of the beta DM, which effectively localizes charge onto individual atoms and creates a sharp, atom-block structure in the initial guess. This can be unnecessarily aggressive and may push the SCF into regions that are hard to converge from.

breaksym=2 instead makes the smallest change that breaks symmetry: a single orbital rotation at the frontier. The resulting DMs are physically well-formed (positive semidefinite, correct electron count Tr(S·DM) = N_elec), delocalized over the full molecule, and provide a gentler, smoother path away from the RHF fixed point  reducing the risk of collapsing back to the closed-shell solution.

### DIFF
--- a/pyscf/scf/test/test_uhf.py
+++ b/pyscf/scf/test/test_uhf.py
@@ -158,6 +158,7 @@ class KnownValues(unittest.TestCase):
         # DMs must be positive semidefinite (physically valid density matrices)
         self.assertTrue(numpy.all(numpy.linalg.eigvalsh(dma) > -1e-10))
         self.assertTrue(numpy.all(numpy.linalg.eigvalsh(dmb) > -1e-10))
+
     def test_break_spin_symm_mix_h2_dissociation(self):
         # At stretched H2 (well past the Coulson-Fischer point) UHF with
         # breaksym='mix' should find a lower-energy broken-symmetry solution
--- a/pyscf/scf/test/test_uhf.py
+++ b/pyscf/scf/test/test_uhf.py
@@ -158,17 +158,6 @@ class KnownValues(unittest.TestCase):
         # DMs must be positive semidefinite (physically valid density matrices)
         self.assertTrue(numpy.all(numpy.linalg.eigvalsh(dma) > -1e-10))
         self.assertTrue(numpy.all(numpy.linalg.eigvalsh(dmb) > -1e-10))
-
-        # Delocalization: the rotated HOMO retains cross-atom (off-diagonal)
-        # character, unlike breaksym=1 which explicitly zeroes those blocks.
-        slices = mol_h2.aoslice_by_atom()
-        p0, p1 = slices[0][2], slices[0][3]
-        p2, p3 = slices[1][2], slices[1][3]
-        self.assertGreater(abs(dma[p0:p1, p2:p3]).max(), 0.1)
-
-        # breaksym=1 zeros the cross-atom block in dmb; confirm the contrast
-        _, dmb1 = mf_h2.init_guess_by_minao(mol_h2, breaksym=1)
-        self.assertAlmostEqual(abs(dmb1[p0:p1, p2:p3]).max(), 0.0, 10)
     def test_break_spin_symm_mix_h2_dissociation(self):
         # At stretched H2 (well past the Coulson-Fischer point) UHF with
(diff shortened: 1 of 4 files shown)


jeanwsr:
This is already included in the examples. But it's good to have this in the scf module.

Something to improve:
1. Based on my experience, if there's no mo_coeff, we should do a roughly converged RHF/RKS then mix.
2. can we use something like `breaksym='mix'` for better readability.

jeanwsr:
Please add tests, and update examples/scf/32 and 56. For 56, we can keep it and just add a note for this new feature.

lynn-buehl:
I added tests in test_uhf.py covering electron count preservation, positive semidefiniteness, delocalization (cross-atom off-diagonal elements), and H2 dissociation beyond the Coulson-Fischer point. Examples 32 and 56 are updated 56 keeps the original init_guess_mixed function and just adds a note pointing to the new built-in option.

pyscf/scf/test/test_uhf.py
```diff
@@ -139,6 +139,55 @@ def test_init_guess_sap(self):
...
+        # DMs must be positive semidefinite (physically valid density matrices)
+        self.assertTrue(numpy.all(numpy.linalg.eigvalsh(dma) > -1e-10))
+        self.assertTrue(numpy.all(numpy.linalg.eigvalsh(dmb) > -1e-10))
+
+        # Delocalization: the rotated HOMO retains cross-atom (off-diagonal)
+        # character, unlike breaksym=1 which explicitly zeroes those blocks.
+        slices = mol_h2.aoslice_by_atom()
+        p0, p1 = slices[0][2], slices[0][3]
+        p2, p3 = slices[1][2], slices[1][3]
+        self.assertGreater(abs(dma[p0:p1, p2:p3]).max(), 0.1)
```
jeanwsr: I think there's no strict requirement for delocalization. Let's remove this check.

pyscf/scf/test/test_uhf.py
```diff
@@ -139,6 +139,54 @@ def test_init_guess_sap(self):
...
+
+        # DMs must be positive semidefinite (physically valid density matrices)
+        self.assertTrue(numpy.all(numpy.linalg.eigvalsh(dma) > -1e-10))
+        self.assertTrue(numpy.all(numpy.linalg.eigvalsh(dmb) > -1e-10))
+
+        # Delocalization: the rotated HOMO retains cross-atom (off-diagonal)
+        # character, unlike breaksym=1 which explicitly zeroes those blocks.
+        slices = mol_h2.aoslice_by_atom()
+        p0, p1 = slices[0][2], slices[0][3]
+        p2, p3 = slices[1][2], slices[1][3]
```
jeanwsr:
```suggestion

```

jeanwsr (approved):

jeanwsr:
please fix lint.

lynn-buehl:
Could it be that these 2 remaining failures (test_grad_h2_cms3ftlda22_sto3g_slow and test_h2_cms3ftlda22_sto3g) are pre-existing MC-PDFT issues unrelated to this PR? The NumPy 2.0 collection errors also appear to be pre-existing compatibility issues. All other 3211 tests pass successfully.

MatthewRHermes:
> Could it be that these 2 remaining failures (test_grad_h2_cms3ftlda22_sto3g_slow and test_h2_cms3ftlda22_sto3g) are pre-existing MC-PDFT issues unrelated to this PR? The NumPy 2.0 collection errors also appear to be pre-existing compatibility issues. All other 3211 tests pass successfully.

Yes, that is what's happening

PR #3286, opened 2026-06-26
Fix unit tests when XCFun not enabled.

NathanGillispie:
## Changes
When XCFun is not enabled the following changes now occur
- **test_xcfun.py**: skips whole module
- **test_he.py**: skips a test that previously gave an attribute error
- **test_xc_deriv.py**: skips actually show up on pytest now and give more useful messages.
- **test_tdrks_grad.py**: Again, skips actually show up on pytest. Also two more skips added.

### Problem 1: Some tests did not check for XCFun in the first place, and failed.
For `pyscf/dft/test/test_xcfun.py` I just added a module-level skip. I also added skips for a few other tests.

### Problem 2: skipIf decorators for tests within classes deselect instead of skipping.
I'm using pytest v9.0.2, perhaps this isn't the case for you. I'd be interested in hearing other peoples experiences. On tests within classes, the decorator syntax (`@unittest.skipIf(not has_xcfun, "...")`) deselects instead of actually skipping. I added lines like
```py
if not hasattr(dft, 'xcfun'):
    raise unittest.SkipTest('PySCF was not built with XCFun.')
```
**Effect:** When running `pytest`, you see the actual skip markers `......s..` and skip message. This is the expected behavior.

## Questions
There are many tests that have problem 2. Here are some files where this is the case:
- `tdscf/test/test_tdrks.py`
- `dft/test/test_xc_deriv.py`
- `grad/test/test_tdrks_grad.py`

Do you want me to change these as well?

### DIFF
--- a/pyscf/dft/test/test_he.py
+++ b/pyscf/dft/test/test_he.py
@@ -114,6 +114,8 @@ class KnownValues(unittest.TestCase):
         self.assertAlmostEqual(mf.e_tot, -1.9931564410562266, 9)
 
     def test_xcfun_nr_blyp(self):
+        if not hasattr(dft, 'xcfun'):
+            raise unittest.SkipTest('PySCF was not built with XCFun.')
         m = mol.RKS()
         m._numint.libxc = dft.xcfun
         m.xc = 'b88,lyp'
--- a/pyscf/dft/test/test_xc_deriv.py
+++ b/pyscf/dft/test/test_xc_deriv.py
@@ -521,8 +521,12 @@ class KnownValues(unittest.TestCase):
         self.assertAlmostEqual(xc1.sum(), 19.977512805950784, 9)
         self.assertAlmostEqual(abs(ref[1] - xc1).max(), 0, 9)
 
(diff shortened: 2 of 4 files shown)


jeanwsr:
If I use `@pytest.mark.skipif( ... , reason='')`, it skips as expected.

The deselect might be caused by mixing pytest and unittest decorator here, I guess, although I cannot find some documentation about this. Personally I don't care much about whether it skips or deselects.

NathanGillispie:
The only issue with using pytest decorators is that the convention seems to be unittest. Theoretically, it adds an extra dependency. However, I think everyone just uses pytest instead. If you actually run unittest a lot of tests fail because it ignores the pytest.ini file which is hiding a lot of failing tests (e.g. scf/test_breit_high_cost, scf/test_diffuse_orbital). My preference is to keep tests compatible with both pytest and unittest. This should be fine because pytest is supposed to be backwards compatible, but it misses skips as mentioned above.

jeanwsr (approved):
Problem 1 fix LGTM. Problem 2 fix looks not very necessary but still beneficial.

PR #3287, opened 2026-06-27
build(windows): add local win64 wheel packaging path

psiQAQ:
## Summary

This PR keeps the legacy Windows wheel path as a local packaging and diagnostic workflow. The formal installed-wheel CI is tracked separately in #3328.

It provides:

- reproducible build and test Conda environments
- an MSYS2 UCRT64 GCC/OpenBLAS wheel build script
- a local installed-wheel verifier with per-directory staging, logs, and Markdown/JSON reports
- Windows DLL lookup support and focused regression tests
- setup, verification, and cleanup documentation

## Shared rules with #3328

The local workflow now follows the CI-validated core rules for:

- removing stale build outputs and staged DLLs before a clean build
- preserving command arguments that contain spaces
- checking required extension, support, and runtime DLLs in the wheel
- clearing `PYTHONPATH` and `PYSCF_EXT_PATH`
- requiring `pyscf` to import from `site-packages`
- writing reports and logs before propagating a nonzero pytest result

The local path intentionally retains its configurable environment names, manual parameters, per-directory test isolation, detailed reports, and ability to reuse an existing wheel or build tree. It is not an official Windows CI or release pipeline.

## Validation

- `python -m unittest discover tools/windows/tests -v`: 29 passed
- PowerShell syntax checks passed for both scripts
- local wheel build completed with MSYS2 UCRT64 GCC/OpenBLAS
- wheel: `pyscf-2.13.1-py3-none-win_amd64.whl`, 92,914,074 bytes
- wheel payload check: 1,199 entries, all required DLL groups present
- SHA-256: `b6b71a80ef9d2ca1bf5468143d9c7dfb76dbf31f7801e76181b941032c9fcb66`

The existing full local installed-wheel sweep remains documented as 51/53 test directories, with the two known numerical/platform deviations retained for separate scientific follow-up.

## Relationship to #3328

#3328 owns the minimal formal Windows installed-wheel CI and excludes `tools/windows`. This PR owns the local packaging helpers and links back to #3328 instead of copying its workflow. Both currently need the same bundled-XCFun `M_PI` compatibility hunk in `pyscf/lib/CMakeLists.txt`; after #3328 is merged, this Draft should be rebased so that shared hunk drops from this PR.

## AI Disclosure

I used OpenAI Codex to inspect the related code paths, refine the PowerShell and packaging changes, organize the documentation, and generate parts of the local verification coverage. I reviewed the patch and can explain the submitted changes.


jeanwsr:
The `conda/` folder is for conda recipes, not wheel-build scripts.

pyscf/lib/misc.py
```diff
@@ -103,17 +103,28 @@ def _ldd(so_file):
...
-        _loaderpath = os.path.dirname(__file__)
         lib = numpy.ctypeslib.load_library(libname, _loaderpath)
     except OSError:
         pass
 
     if lib is None and sys.platform == 'win32':
-        for env_path in [os.path.join(sys.prefix, 'Library', 'bin'),
+        # Look in deps/bin before the environment so large support DLLs can stay
+        # bundled once under pyscf/lib/deps/bin instead of being duplicated in pyscf/lib.
+        for env_path in [os.path.join(_loaderpath, 'deps', 'bin'),
```
jeanwsr: pyscf does not load libraries in deps directly.

pyscf/lib/misc.py
```diff
@@ -103,17 +103,28 @@ def _ldd(so_file):
     'libxcfun_itrf': ['xcfun'],
 }
 
+if sys.platform == 'win32':
+    # Windows wheels ship these support libraries with the "lib" prefix,
+    # unlike the long-standing Linux/macOS loader names.
```
jeanwsr: Unfortunately you cannot override this since conda-forge libxc provides xc.dll, which is needed by upcoming conda-forge pyscf. What about try both xc and libxc.

tools/windows/build-wheel.ps1
```diff
@@ -0,0 +1,186 @@
...
+    )
+    $BootstrapPaths = $BootstrapPaths | Where-Object { $_ } | Select-Object -Unique
+    $env:PATH = $BootstrapPaths -join ';'
+
+    Require-Command "cmake" | Out-Null
+    $NinjaExe = Resolve-NinjaExe -LibraryBin $LibraryBin -ScriptsDir $ScriptsDir
+    Require-Command "gcc" | Out-Null
+    Require-Command "g++" | Out-Null
+
+    Copy-RequiredDlls -RuntimeDllDir $RuntimeDllDir -LibDir $LibDir
```
jeanwsr: Is it possible to use delvewheel for this since we use auditwheel in linux wheel-build.

psiQAQ:
> The `conda/` folder is for conda recipes, not wheel-build scripts.

Thanks, addressed. I moved the Windows wheel helper files from `conda/windows` to `tools/windows`, so `conda/` is now kept for conda recipe related files only.

pyscf/lib/misc.py
psiQAQ: Addressed. I removed the direct deps/bin lookup and also removed the os.add_dll_directory path registration. The Windows build script now stages libcint.dll and libxc.dll into pyscf/lib, and MANIFEST.in excludes the duplicate copies under pyscf/lib/deps/bin.

pyscf/lib/misc.py
psiQAQ: Addressed. The Windows dependency loader now tries both names: xc before libxc, and similarly xcfun before libxcfun. The fallback handling is still only used on the win32 path.

tools/windows/build-wheel.ps1
psiQAQ: I did not add delvewheel in this revision. For this PR, I kept the current local Windows wheel path minimal and addressed the loader/layout issues first. I agree that a build wheel -> delvewheel repair flow is worth evaluating, but I would prefer to handle that as a follow-up step unless you think it should block this initial PR.

pyscf/lib/misc.py
jeanwsr: Oh I'm sorry I made a mis-suggestion. On windows libcint is indeed loaded because of the dll_wrapper, but in conda-forge build it does not live in `deps/`, so deps is not searched in my previous commit. It's reasonable to search this. But your current copy-solution is also fine.

psiQAQ: Thanks for the clarification. I will keep the current staging approach for this PR, since it avoids adding a new `deps/bin` search path and keeps the DLL layout closer to the existing `pyscf/lib` loading behavior.

psiQAQ:
I can also add a short pointer in the installation manual under "Build from source" to mention the local Windows wheel build helper. I would keep the detailed steps in `tools/windows/win64-package-readme.md` and avoid changing the main pip/conda installation instructions until official Windows wheels are published.

How to install PySCF
├── Warning
├── Install with pip
├── Build from source with pip
├── Build from source
├── Build a local Windows wheel      ← here
├── Install with conda
├── Install on Fedora
├── Docker image
└── Advanced build options

## Build a local Windows wheel

Native Windows installation from PyPI is not yet part of the official release workflow. For developers and advanced users who want to build a local Windows 64-bit wheel, PySCF provides a helper workflow under `tools/windows`.

The Windows helper builds a local wheel, stages the required runtime DLLs, and documents the install and smoke-test steps. See:

```text
tools/windows/win64-package-readme.md
```

This workflow is intended for local packaging and testing. Non-developer users are still recommended to use the released PyPI wheels on Linux/macOS or the Windows Subsystem for Linux.

jeanwsr:
Documentation pr is welcome once this merged. Before that, how many tests can your locally built wheel pass? (like, at least test for `gto/` and `scf/`)

pyscf/lib/misc.py
```diff
@@ -99,15 +99,32 @@ def _ldd(so_file):
...
+    'libxcfun_itrf': [('xcfun', 'libxcfun')],
 }
 
+def _load_dependency(libname):
+    # Some Windows builds use the conda-forge DLL names while the local wheel
+    # packaging path keeps the lib-prefixed filenames.
+    if isinstance(libname, str):
+        return load_library(libname)
+
+    for candidate in libname:
```
jeanwsr: check if libname is a tuple

psiQAQ:
> Documentation pr is welcome once this merged. Before that, how many tests can your locally built wheel pass? (like, at least test for `gto/` and `scf/`)

I verified the locally built wheel in a dedicated Windows test environment: `pyscf\gto\test` and `pyscf\scf\test` pass, and the latest full installed-wheel sweep passes 51/53 discovered test directories, with the remaining two failures documented as known Windows numerical deviations rather than packaging regressions.

pyscf/lib/misc.py
psiQAQ: I changed `_load_dependency` to treat `libname` explicitly as a tuple of ordered fallback candidates; strings are still handled directly, and other input types now raise `TypeError` so the Windows DLL candidate semantics stay explicit instead of accepting arbitrary iterables.

jeanwsr:
> > Documentation pr is welcome once this merged. Before that, how many tests can your locally built wheel pass? (like, at least test for `gto/` and `scf/`)
> 
> I verified the locally built wheel in a dedicated Windows test environment: `pyscf\gto\test` and `pyscf\scf\test` pass, and the latest full installed-wheel sweep passes 51/53 discovered test directories, with the remaining two failures documented as known Windows numerical deviations rather than packaging regressions.

You can ignore the test failure if they are already tracked by #3245 .

psiQAQ:
> > > Documentation pr is welcome once this merged. Before that, how many tests can your locally built wheel pass? (like, at least test for `gto/` and `scf/`)
> > 
> > 
> > I verified the locally built wheel in a dedicated Windows test environment: `pyscf\gto\test` and `pyscf\scf\test` pass, and the latest full installed-wheel sweep passes 51/53 discovered test directories, with the remaining two failures documented as known Windows numerical deviations rather than packaging regressions.
> 
> You can ignore the test failure if they are already tracked by #3245 .

I rechecked the two Windows installed-wheel outliers using the same test environment (`pyscf-win313-test`) and the same built wheel (`dist\pyscf-2.13.1-py3-none-win_amd64.whl`).

After activating the test environment with

the isolated single-test reruns are:

```powershell
conda activate pyscf-win313-test
$env:PYTEST_DISABLE_PLUGIN_AUTOLOAD = "1"
python -m pytest --import-mode=importlib -q pyscf/cc/test/test_eom_gccsd.py::KnownValues::test_eaccsd
python -m pytest --import-mode=importlib -q pyscf/pbc/tdscf/test/test_uks.py::DiamondPBE0::test_tdhf
Remove-Item Env:PYTEST_DISABLE_PLUGIN_AUTOLOAD -ErrorAction SilentlyContinue
```

The installed-wheel targeted reruns are:

```powershell
conda activate pyscf-win313-test
.\tools\windows\verify-installed-wheel.ps1 -SkipBuild -SkipInstall -TestRoots 'pyscf\cc\test'
.\tools\windows\verify-installed-wheel.ps1 -SkipBuild -SkipInstall -TestRoots 'pyscf\pbc\tdscf\test'
```

Results:
- `pyscf/cc/test/test_eom_gccsd.py::KnownValues::test_eaccsd` is flaky in isolated reruns on Windows (5 runs: 2 pass, 3 fail), with the expected small left/right overlap warning. This looks consistent with the known `eom_gccsd` numerical-instability class already tracked in #3245, not a packaging regression.
- The earlier `pyscf/pbc/tdscf/test/test_uks.py::DiamondPBE0::test_tdhf` outlier is not reproducible now: isolated reruns pass, and the targeted installed-wheel rerun of `pyscf\pbc\tdscf\test` also passes. #3245 mentions other `tdscf` cases, not this exact one.

So I am not treating these as blockers for the Windows wheel packaging PR.

tools/windows/reports/installed-wheel-report.md
```diff
@@ -0,0 +1,84 @@
+﻿# Installed Wheel Verification Report
```
jeanwsr: please don't include this file in source code.

pyscf/lib/test/test_misc.py
```diff
@@ -87,6 +90,24 @@ def test_isintsequence(self):
...
+
+    def test_windows_misc_does_not_search_deps_bin_for_support_dlls(self):
+        text = LIB_MISC.read_text(encoding='utf-8')
+        # Keep the direct DLL search limited to pyscf/lib and the active
+        # environment. Support DLLs should not be loaded from deps/bin.
+        self.assertNotIn("os.path.join(_loaderpath, 'deps', 'bin')", text)
+        self.assertIn("def _load_dependency(libname):", text)
+        self.assertIn("if isinstance(libname, tuple):", text)
+        self.assertIn("raise TypeError(", text)
+
```
jeanwsr: not very meaningful to do these two tests, let's remove them.

tools/windows/reports/installed-wheel-report.md
psiQAQ: Fixed. I removed `tools/windows/reports/installed-wheel-report.md` from the tracked source files and updated `.gitignore` so future generated verification reports under `tools/windows/reports/` stay local-only.

pyscf/lib/test/test_misc.py
psiQAQ: Fixed. I removed the two Windows-specific `test_misc.py` checks and restored `pyscf/lib/test/test_misc.py` to match `master`. I also removed the corresponding special-case `misc.py` staging from `verify-installed-wheel.ps1` and updated the related script test and README accordingly.

pyscf/lib/CMakeLists.txt
```diff
@@ -252,12 +254,13 @@ if(ENABLE_XCFUN AND BUILD_XCFUN)
...
     GIT_TAG a89b783
     # This patch adds the xcfun derivative order up to 5
-    PATCH_COMMAND git apply --reject ${PROJECT_SOURCE_DIR}/libxcfun.patch || true
+    PATCH_COMMAND git apply --reject ${PROJECT_SOURCE_DIR}/libxcfun.patch || ${CMAKE_COMMAND} -E true
     PREFIX ${PROJECT_BINARY_DIR}/deps
     INSTALL_DIR ${PROJECT_SOURCE_DIR}/deps
     CMAKE_ARGS -DCMAKE_BUILD_TYPE=RELEASE -DBUILD_SHARED_LIBS=1
             -DXCFUN_MAX_ORDER=${XCFUN_MAX_ORDER} -DENABLE_TESTALL=0
             -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
+            -DCMAKE_CXX_FLAGS=-DM_PI=3.14159265358979323846
```
jeanwsr, line 266: Any explanation on this change?

tools/windows/win64-package-readme.md
```diff
@@ -0,0 +1,273 @@
...
+   - Observed failure: the `TDDFT` result differed from the hard-coded reference by `0.10508681940613762 eV`, exceeding the current `assertAlmostEqual(..., 5)` tolerance.
+   - Current assessment: this looks like a platform- or linear-algebra-dependent numerical deviation in the periodic TDDFT/TDHF test case rather than an installed-wheel staging or packaging defect.
+   - Review position: worth follow-up as a separate Windows numerical deviation, but not currently treated as a wheel packaging merge blocker.
+
+### TODO: `libxcfun.patch` Is Corrupt
+
+During the Windows `xcfun` packaging work, the build exposed a pre-existing issue in `pyscf\lib\libxcfun.patch`:
+
+```text
+error: corrupt patch at line 11
```
jeanwsr: is it a problem for all platform?

pyscf/lib/CMakeLists.txt
psiQAQ: I checked this explicitly on the bundled Windows `xcfun` build path.

Without the explicit `M_PI` define, the external `xcfun` build fails because `M_PI` is referenced in `xcfun` sources before it is made visible by a later included header in this toolchain path. With the explicit define, the same `xcfun` build completes successfully.

So this change is only a local workaround for the external `libxcfun` build when `BUILD_XCFUN=ON`; it does not affect the rest of PySCF.

tools/windows/win64-package-readme.md
psiQAQ: I re-checked this, and the answer is no: it is not a problem for all platforms, and the patch itself is not inherently corrupt.

A clean Linux checkout and a clean Windows checkout with line-ending conversion disabled both produced the same LF-only `pyscf/lib/libxcfun.patch`. In both cases, the LF-only patch passed `git apply --check` against the pinned bundled `xcfun@a89b783` source tree and applied cleanly.

The failure I saw in the Windows packaging environment was checkout-specific. That environment had `core.autocrlf=true`, and without a `.gitattributes` rule, the Windows worktree rewrote `pyscf/lib/libxcfun.patch` from LF to CRLF. That CRLF copy then failed during `git apply` parsing with:

```text
error: corrupt patch at line 11
```

So this is not a Windows compiler issue and not a target-tree issue with `xcfun@a89b783`; it is a line-ending preservation issue for this patch file on Windows-style checkouts.

I fixed this in the PR by adding:

```text
pyscf/lib/libxcfun.patch text eol=lf
```

to `.gitattributes`, and I updated the README to describe the diagnosis and validation path.

tools/windows/win64-package-readme.md
```diff
@@ -0,0 +1,286 @@
...
+Remove-Item .\pyscf\lib\*.dll -Force -ErrorAction SilentlyContinue
+
+# 5. Optional: remove timestamped historical reports and keep only the stable report filenames
+Remove-Item .\tools\windows\reports\installed-wheel-report-*.md -Force -ErrorAction SilentlyContinue
+Remove-Item .\tools\windows\reports\installed-wheel-report-*.json -Force -ErrorAction SilentlyContinue
+```
+
+## Appendix
+
+### Known Windows Test Deviations
```
jeanwsr: no need to document this.

pyscf/lib/CMakeLists.txt
```diff
@@ -252,12 +254,16 @@ if(ENABLE_XCFUN AND BUILD_XCFUN)
     GIT_REPOSITORY https://github.com/dftlibs/xcfun.git
     GIT_TAG a89b783
     # This patch adds the xcfun derivative order up to 5
-    PATCH_COMMAND git apply --reject ${PROJECT_SOURCE_DIR}/libxcfun.patch || true
+    PATCH_COMMAND git apply --reject ${PROJECT_SOURCE_DIR}/libxcfun.patch || ${CMAKE_COMMAND} -E true
```
jeanwsr: is this change still needed?

tools/windows/win64-package-readme.md
psiQAQ: Removed this appendix section.

pyscf/lib/CMakeLists.txt
psiQAQ: I rechecked this after fixing the patch line-ending handling. The worktree copy of pyscf/lib/libxcfun.patch is now preserved as LF via .gitattributes, git apply --check succeeds against the bundled xcfun source tree, and the clean Windows wheel build still completes successfully. With that in place, this cross-platform wrapper is no longer needed, so I reverted the line back to the upstream-style || true form in 2988ce648.

psiQAQ:
This draft has completed its historical purpose: it established and validated the local Windows wheel build path, installed-wheel verification, and the evidence that informed the focused Windows CI work in #3328.

I am closing it rather than asking upstream to merge this broad local-packaging branch. `win64-pack` will remain maintained in my fork and locally as a personal Windows wheel build and validation branch; it is not intended to remain an upstream PR.

Thank you for the review and guidance that helped separate local packaging from the focused CI work.

PR #3292, opened 2026-07-02
Fix: Replace true division with floor division in MakeAtomInfos()

nominhanggai:
In pyscf/lo/ibo.py, replace true division (/) with floor division (//) to ensure nCoreX values are integers instead of floats. This fixes type inconsistencies for elements Rb, Sr, Cs, Ba, In, Sn, Sb, Te, I, Xe, La, Hf, Ta, W, Re, Os, Ir, Pt, Au, Hg, Tl, Pb, Bi, Po, At, Rn, Fr, Ra, Th, Pa, U, Np, Pu, Am, Cm, Bk, Cf, Es, Fm, Md, No, Ac, Lr, Rf, Db, Sg, Bh, Hs, Mt, Ds, Rg, Cn, Nh, Fl, Mc, Lv, Ts, and Og.

Fixes: nCoreX values now consistently return integers instead of floats.


jeanwsr:
Is this covered by any test?

nominhanggai:
> Is this covered by any test?

Added a test in `pyscf/lo/tests/test_ibo.py` to verify that the indices are integers.

jeanwsr (approved):

PR #3293, opened 2026-07-03
KRPA and KURPA: improve the numerical stablity of ACFD EXX

lijiachen417:
In previous implementation for evaluating HF+ACFD exchange together, it's numerically unstable when systems have orbitals small fractional occupation numbers.

For example, diamond GTH-Pade/GTH-DZVP, RPA@PBE, old implementation gives non-smooth ACFD exchange with respect to the lattice constant
image

Current implementation solves the problem
image


lijiachen417:
Failed test is not related to this PR.

sunqm (approved):

PR #3283, opened 2026-06-22
Add AI_POLICY.md

sunqm:
The proposed policy is adapted from NumPy's AI policy.

This PR follows discussion #3254 and is intended to gather community feedback. Comments and suggestions are welcome.

AI-related guidelines, including this proposal, can be discussed and potentially finalized at the 2026 PySCF developers' meeting.

See also PR #3438

PR #3311, opened 2026-07-12
Correct LIBXC_is_cam_rsh ctypes argtypes

steps-re:
`LIBXC_is_cam_rsh.argtypes` was declared as `(c_int, c_void_p)`, copied from the unrelated `LIBXC_is_nlc`. The real C signature in `libxc_itrf.c` takes a single `xc_func_type*`. Any range-separated-hybrid xc string that combines two or more genuine RSH functionals hit this path and crashed with a ctypes `TypeError`.

Corrected the `argtypes` to a single pointer. Reproduced with `dft.libxc.rsh_coeff('0.5*CAM_B3LYP + 0.5*CAM_O3LYP')`, which crashed before and returns clean coefficients after. No regression across the existing `test_parse_xc`/`test_libxc` RSH cases. Test added.

Note: the underlying defect is the `argtypes` declaration, not the call site.


susilehtola:
This PR as well as #3310 and #3309 look AI-generated, so tagging #3283.

The fix seems OK but there's an unrelated change included in the tests, as well.

sunqm (approved):

Issue #3312, opened 2026-07-12
[UNEXPECTED BEHAVIOR] Thread-count-dependent outcomes in selected flaky tests


psiQAQ:
> **Status (2026-08-20): 7 of 9 open entries are resolved and checked off.** Merged fixes: #3404 (EOM degenerate star contraction), #3405 (X2C eigh fallback), #3406 + #3409 (CIAH deadlock + sa4 tolerance), #3407 (rccsd fingerprint precision); dependency fix LibXC MR !806 integrated via #3336 (released 7.1.2 now built by master). Remaining: `test_uhf_smearing` (pending #3410, the e_free convergence criterion - independent cross-validation already clean) and `test_finite_diff_grad` (SGX; to be reworked on top of #3410's `check_extra_convergence` hook). Details in the checklist and thread below.

**Describe the unexpected behavior**

I investigated 10 flaky PySCF test nodeids across seven OS/Python combinations and four OpenMP/BLAS profiles. The first results were correlated with higher thread counts, but the completed experiments do not support one common threading bug.

The cases instead separate into four causes: ambiguous root or state selection after small floating-point changes; iterative solvers exhausting a usable trial subspace; an optimizer reusing a zero augmented-Hessian step; and a LibXC WPBEH regression affecting ordinary RKS, SGX, and PBC HSE calculations.

Several historical candidate matrices completed without numerical failures, but maintainer review and final-code provenance checks reopened part of the list. PRs #3319, #3324, and #3332 were closed without merge; their artifacts remain useful evidence, but their proposed changes are not accepted fixes. LibXC 7.1.0 has been released, while its PySCF CI integration is targeted for a 2.14 patch release.

The working branch, `psiQAQ/pyscf:win64-CI`, is only a diagnostic branch. Final changes will be submitted as focused PRs by root cause, not as one PR containing the diagnostic history.

**To Reproduce**

The diagnostic workflow is `.github/workflows/ci-precision-diagnostics.yml`. It invokes `.github/workflows/precision_experiments.py` through `.github/workflows/run_paired_precision_diagnostics.py`.

The matrix is:

- Ubuntu: Python 3.8, 3.12, and 3.13;
- Windows: Python 3.12 and 3.13;
- macOS: Python 3.8 and 3.13;
- OpenMP/BLAS profiles: `1/1`, `4/1`, `1/4`, and `4/4`;
- up to 200 fresh-process attempts per profile.

[...]

PR #3317, opened 2026-07-14
fix(tdscf): retain raw residual fallback

psiQAQ:
## Problem

The TD-SCF Davidson solver can stop after every preconditioned residual direction is projected out, even when a corresponding raw Ritz residual still contains an independent expansion direction. In the selected CAM-B3LYP TDDFT calculation, this produced intermittent non-convergence and an incorrect root fingerprint.

## Root cause

The condition occurs in the real `test_tddft_camb3lyp` system without mocking the solver state. With `OMP_NUM_THREADS=4` and `OPENBLAS_NUM_THREADS=1`, attempt 49 reached a trial-space size of 39. The existing orthogonalizer accepted 0 of 20 preconditioned residual candidates, while the same orthogonalizer accepted 1 independent direction from the corresponding 20 raw Ritz residuals. With that direction, all four TD roots converged to `[0.2611815441, 0.2905656337, 0.3254425081, 0.3680400043]`.

The unfixed intermittent failures are recorded in [run 29213311767](https://github.com/psiQAQ/pyscf/actions/runs/29213311767).

## Scope and change

When all preconditioned candidates are linearly dependent in the non-symmetry-partitioned path, retry trial-space expansion once with the raw Ritz residuals. If those residuals are also dependent, the existing restart/exit behavior remains unchanged. Symmetry-partitioned solves keep their existing filtered path.

A focused regression makes the observed empty preconditioned basis deterministic. This PR does not change tolerances, iteration limits, root selection, or scientific assertions.

## Verification

- The regression uses a zero preconditioner to distinguish the existing Ritz restart from the raw-residual fallback.
- Current rebased head `913c8778e` passed `test_lr_eig.py`, the real CAM-B3LYP nodeid, and the PBC TDHF nodeid that shares the eigensolver with `OMP_NUM_THREADS=1` and `OPENBLAS_NUM_THREADS=1`: `5 passed`.
- [Formal diagnostics run 29301363348](https://github.com/psiQAQ/pyscf/actions/runs/29301363348) covered seven OS/Python combinations, four OpenMP/BLAS profiles, and 200 attempts per profile. All 5,600 original direct-root and fingerprint assertions passed, with complete artifacts.

## Related

Related to the CAM-B3LYP TDDFT item tracked in #3312.

### DIFF
--- a/pyscf/tdscf/_lr_eig.py
+++ b/pyscf/tdscf/_lr_eig.py
@@ -707,6 +707,11 @@ def real_eig(aop, x0, precond, tol_residual=1e-5, nroots=1, x0sym=None, pick=Non
         if x0sym is None:
             V, W = VW_Gram_Schmidt_fill_holder(
                 V_holder[:,:m1], W_holder[:,:m1], X_new, Y_new, lindep)
+            # A dependent preconditioned basis can still leave independent raw Ritz residuals.
+            if len(V) == 0:
+                V, W = VW_Gram_Schmidt_fill_holder(
+                    V_holder[:,:m1], W_holder[:,:m1],
+                    R_x[:,r_index].copy(), R_y[:,r_index].copy(), lindep)
         else:
             xt_ir = xt_ir[r_index]
             xt_orth_ir = []
--- a/pyscf/tdscf/test/test_lr_eig.py
+++ b/pyscf/tdscf/test/test_lr_eig.py
@@ -86,6 +86,23 @@ class KnownValues(unittest.TestCase):
         self.assertTrue(converged[0])
         self.assertAlmostEqual(energy[0], 0.9991663794740591, 12)
 
+    def test_real_eig_uses_raw_residual_when_preconditioner_returns_zero(self):
+        def zero_precond(dx, energy):
+            return numpy.zeros_like(dx)
+
+        converged, energy, _ = _lr_eig.real_eig(
+            self.aop,
+            self.guess,
+            zero_precond,
+            nroots=1,
+            tol_residual=1e-9,
+            max_cycle=30,
+            verbose=logger.Logger(sys.stdout, logger.QUIET),
+        )
+
+        self.assertTrue(converged[0])
+        self.assertAlmostEqual(energy[0], 0.9991663794740591, 12)
+
 
 if __name__ == '__main__':
     unittest.main()


pyscf/tdscf/_lr_eig.py
```diff
@@ -716,6 +716,11 @@ def real_eig(aop, x0, precond, tol_residual=1e-5, nroots=1, x0sym=None, pick=Non
                 xt_ir = np.hstack(xt_orth_ir)
                 xs_ir = np.hstack([xs_ir, xt_ir])
 
+        # A dependent preconditioned basis can still leave independent raw Ritz residuals.
+        if len(V) == 0 and x0sym is None:
```
sunqm: Move the code block to the relevant place.

sunqm:
This fix is not wrong, but it does not make much sense to me. Can you show a real example that the residual vectors are linearly independent while the preconditioned residual vectors are linearly dependent?

psiQAQ:
Yes. I reran the actual `test_tddft_camb3lyp` system (water/6-31G, UKS CAM-B3LYP, four roots) with `OMP_NUM_THREADS=4` and `OPENBLAS_NUM_THREADS=1`, without mocking the solver state.

On attempt 49, at a trial-space size of 39, the existing orthogonalizer accepted 0 of 20 preconditioned residual candidates. The fallback then accepted 1 independent direction from the corresponding 20 raw Ritz residuals. SCF converged, and all four TD roots converged after the fallback; the roots were `[0.2611815441, 0.2905656337, 0.3254425081, 0.3680400043]`.

The unfixed code showed intermittent `not_converged` results for this nodeid on Ubuntu OpenMP=4 profiles in [run 29213311767](https://github.com/psiQAQ/pyscf/actions/runs/29213311767). With the fallback, the final matrix completed 5600/5600 runs across seven OS/Python combinations and four OpenMP/BLAS profiles in [run 29301363348](https://github.com/psiQAQ/pyscf/actions/runs/29301363348).

This branch is intermittent because it depends on reduction roundoff, so the unit test only makes the observed basis condition deterministic. I also moved the fallback inside the `x0sym is None` branch as requested in `683d3a3e4`.

pyscf/tdscf/_lr_eig.py
psiQAQ: Moved the fallback into the `x0sym is None` branch in `683d3a3e4`, immediately after orthogonalizing the preconditioned residuals.

psiQAQ:
Rebased onto current master after #3315 and resolved the overlapping solver/test changes.

The regression test now distinguishes the two recovery paths: it uses a preconditioner that returns zero vectors. Current master restarts once from the Ritz vectors and then exits unconverged, while this PR falls back to the nonzero raw Ritz residuals and converges. This gives a RED/GREEN check that the raw-residual fallback remains necessary after #3315.

Focused Windows/Python 3.13 validation with OMP/OpenBLAS set to 1:
- `pyscf/tdscf/test/test_lr_eig.py`
- `pyscf/tdscf/test/test_tduks.py::KnownValues::test_tddft_camb3lyp`
- `pyscf/pbc/tdscf/test/test_uks.py::DiamondM06::test_tdhf`

Result: `5 passed`.

The previous Linux 3.12 check failure was the separate PBC `DiamondM06::test_tdhf` assertion; all TDGKS tests passed in that job. I kept that unrelated PBC path out of this PR.

psiQAQ:
The rebased CI has finished. Lint, macOS 3.13, Linux 3.12, and aarch64 all passed.

The only red check is Linux 3.8, where the sole failure was the unchanged `pyscf/fci/test/test_spin_op.py::KnownValues::test_contract_ss` numerical fluctuation (`State not singlet -5.69739e-06`); the other 3,289 tests passed. This path does not use either file changed by this PR.

Run: https://github.com/pyscf/pyscf/actions/runs/29474043633

psiQAQ:
Rebased onto current `master` at `913c8778e` and updated the description with the non-mocked 0/20 preconditioned versus 1/20 raw-residual trajectory. The current-head focused check passed all 5 tests with OMP/OpenBLAS set to 1. Please re-review the rebased head; if an unrelated full-suite failure recurs, it should remain outside this two-file PR.

sunqm (approved):

PR #3318, opened 2026-07-14
test(mcscf): tighten UCASSCF convergence

psiQAQ:
## Summary

- Tighten this U-CASSCF regression's energy convergence threshold to `1e-8`.
- Keep the solver, initial UHF orbitals, and energy assertion unchanged.

## Why

The preceding UHF calculations agree to roundoff, but the subsequent UCASSCF optimization is sensitive to last-bit differences and can enter another stationary-point basin. The previous fixed-orbital proposal made the test deterministic but hid that sensitivity in a serialized seed. A slightly tighter convergence threshold is the smaller test-level control suggested in review.

## Verification

- [Original paired diagnostics](https://github.com/psiQAQ/pyscf/actions/runs/29213310016) reproduced 10 default-path mismatches while the paired UHF energies, spin values, and densities agreed to roundoff.
- The targeted formal runs covered every historically failing OMP=4 profile:
  - [Ubuntu, BLAS=1](https://github.com/psiQAQ/pyscf/actions/runs/29420245618): 600/600 pass.
  - [Ubuntu, BLAS=4](https://github.com/psiQAQ/pyscf/actions/runs/29420250487): 600/600 pass.
  - [Windows, BLAS=1](https://github.com/psiQAQ/pyscf/actions/runs/29420253473): 400/400 pass.
  - [Windows, BLAS=4](https://github.com/psiQAQ/pyscf/actions/runs/29420256753): 400/400 pass.
- [Unified formal run 29424229651](https://github.com/psiQAQ/pyscf/actions/runs/29424229651) covered seven OS/Python combinations, four OpenMP/BLAS profiles, and 200 attempts per profile. All 5,600 attempts converged and passed the original energy assertion; records, 5,600 logs, 56 summaries, 28 environment files, seven paired metadata files, and 56 first-pass snapshot/checkpoint references were validated with no missing references.
- `python -m pytest pyscf/mcscf/test/test_umc1step.py -c pytest.ini -q` — 9 passed.

This addresses the U-CASSCF item tracked in #3312.


jeanwsr:
I don't think H2O triplet UHF has degeneracy problem. Can you show how the wrong SCF result is different from the right one?

psiQAQ:
You are right that the triplet H2O UHF calculation is not where the two outcomes diverge. The UHF SCF result itself is not wrong; the wording in the original PR conflated the UHF solution with the subsequent UCASSCF optimization.

Across the 10 paired mismatches in the diagnostic run, the UHF results from the failing and passing paths agree to numerical noise: the maximum energy difference is `9.95e-14 Ha`, the maximum difference in `<S^2>` is zero at the logged precision, and the maximum density fingerprint and norm differences are `1.69e-13` and `6.71e-14`, respectively.

For example, on Ubuntu/Python 3.12 with OMP=4 and BLAS=1, both paths give the same UHF result,

```
E(UHF) = -75.7263969090366 Ha
<S^2>  = 2.0038015
```

but the subsequent UCASSCF optimizations converge to different stationary points:

```
default path: -75.7302501822368 Ha
fixed path:   -75.7460662737131 Ha
```

I also found that the fixture in the original PR commit was taken from a completed UCASSCF calculation, so it reduced the regression to a near-solution restart (2 macro iterations). I updated the branch with a fixture generated directly from the converged UHF orbitals, before starting UCASSCF. The original energy assertion is unchanged.

The updated fixture passed 9/9 remote checks on Ubuntu with Python 3.8, 3.12, and 3.13 at OMP=4/BLAS=1. The runs retained 19-22 macro iterations and converged to energies between `-75.7460663086226` and `-75.7460662889334` Ha. The records, attempts, logs, summaries, and runtime environments are in [diagnostic run 29404865466](https://github.com/psiQAQ/pyscf/actions/runs/29404865466).

The key correction is that the instability is in the UCASSCF optimization path, not in the preceding UHF solution.

jeanwsr:
> You are right that the triplet H2O UHF calculation is not where the two outcomes diverge. The UHF SCF result itself is not wrong; the wording in the original PR conflated the UHF solution with the subsequent UCASSCF optimization.
> 
> Across the 10 paired mismatches in the diagnostic run, the UHF results from the failing and passing paths agree to numerical noise: the maximum energy difference is `9.95e-14 Ha`, the maximum difference in `<S^2>` is zero at the logged precision, and the maximum density fingerprint and norm differences are `1.69e-13` and `6.71e-14`, respectively.
> 
> For example, on Ubuntu/Python 3.12 with OMP=4 and BLAS=1, both paths give the same UHF result,
> 
> ```
> E(UHF) = -75.7263969090366 Ha
> <S^2>  = 2.0038015
> ```
> 
> but the subsequent UCASSCF optimizations converge to different stationary points:
> 
> ```
> default path: -75.7302501822368 Ha
> fixed path:   -75.7460662737131 Ha
> ```
> 
> I also found that the fixture in the original PR commit was taken from a completed UCASSCF calculation, so it reduced the regression to a near-solution restart (2 macro iterations). I updated the branch with a fixture generated directly from the converged UHF orbitals, before starting UCASSCF. The original energy assertion is unchanged.
> 
> The updated fixture passed 9/9 remote checks on Ubuntu with Python 3.8, 3.12, and 3.13 at OMP=4/BLAS=1. The runs retained 19-22 macro iterations and converged to energies between `-75.7460663086226` and `-75.7460662889334` Ha. The records, attempts, logs, summaries, and runtime environments are in [diagnostic run 29404865466](https://github.com/psiQAQ/pyscf/actions/runs/29404865466).
> 
> The key correction is that the instability is in the UCASSCF optimization path, not in the preceding UHF solution.

Still not understand why the fixture works if it's just using UHF orbitals as before.

psiQAQ:
The default path does use UHF orbitals. `UCASSCF(mf, ...)` stores `mf.mo_coeff`, and `kernel()` uses that array when no explicit `mo_coeff` is supplied. The fixture therefore does not provide a different physical UHF solution; it freezes one numerical representation of the same solution.

I compared a fresh UHF result with the new file on the same build. The occupied-density matrices differ by at most `5.60e-14`, and the singular values comparing the corresponding core, active, and virtual subspaces are all 1 within about `1e-15`. The large raw beta-coefficient difference is only orbital sign changes; after taking absolute overlaps, every diagonal overlap is at least `0.9999999999999968`. Both seeds also give the target UCASSCF state on this build.

The problematic CI cases show what the file changes: although the UHF energy, `<S^2>`, and density are unchanged to roundoff, the finite-precision iterative UCASSCF path can enter a different stationary-point basin from last-bit changes in the live UHF coefficient array. Serializing the coefficient array removes that run/platform-dependent perturbation and makes this regression test exercise one specified basin. It is a deterministic test-seed fix, not a correction to UHF or a general fix for UCASSCF basin sensitivity.

As a longer check of that narrower claim, the fixed seed passed 600/600 runs (200 each on Python 3.8, 3.12, and 3.13 at OMP=4/BLAS=1), while retaining 19 or 22 macro iterations: [run 29406132513](https://github.com/psiQAQ/pyscf/actions/runs/29406132513).

jeanwsr:
I don't think this is an encouraged solution. Could you try `mc.conv_tol = 1e-8`, or even more tight.

psiQAQ:
Thanks. I replaced the orbital fixture with `mc.conv_tol = 1e-8`; the final PR diff is now one line.

I repeated every profile that had previously produced a mismatch: Ubuntu Python 3.8/3.12/3.13 and Windows Python 3.12/3.13, each with OMP=4 and BLAS=1/4, 200 attempts per combination. All 2,000 attempts converged and passed the original energy assertion:

- [Ubuntu, BLAS=1](https://github.com/psiQAQ/pyscf/actions/runs/29420245618)
- [Ubuntu, BLAS=4](https://github.com/psiQAQ/pyscf/actions/runs/29420250487)
- [Windows, BLAS=1](https://github.com/psiQAQ/pyscf/actions/runs/29420253473)
- [Windows, BLAS=4](https://github.com/psiQAQ/pyscf/actions/runs/29420256753)

The full local test file also passes (9/9).

jeanwsr (approved):

PR #3319, opened 2026-07-14
fix(mcscf): restart zero Newton AH steps

psiQAQ:
## Problem

In the SA4 Newton regression, a nonzero orbital gradient can be followed by a zero augmented-Hessian step. Reusing that zero vector as the next macro-cycle guess rebuilds the same degenerate subspace and can run to the macro-cycle limit without convergence.

## Root cause

This path is reachable without mocking. In [run 29273789045](https://github.com/psiQAQ/pyscf/actions/runs/29273789045), Ubuntu Python 3.13 with `omp4-blas1`, attempt 11 recorded the following values in the same microstep:

```text
|g|=1.57e-04
|dxi|=0.00e+00
max(x)=0.00e+00
|dr|=0.00e+00
seig=1.6e-28
```

The zero step was then reused as the next macro initial guess. The calculation repeated with `dE=0` and the same gradient through macro 50 and ended unconverged. The corresponding artifact is `precision-thread-paired-split-sa4-newton-ubuntu-py3.13`, file `omp4-blas1/logs/sa4-newton.attempt-11.log`.

## Scope and change

When the returned AH step is below the existing `ah_lindep` threshold, use the current full gradient as the next macro-cycle initial guess. `update_orb_ci` already uses this gradient as the default guess when no prior guess is supplied, so this fallback reuses the existing initialization path.

The deterministic Davidson stub covers the intermittent nonzero-gradient/zero-step branch. This PR does not change convergence tolerances, iteration limits, step sizes, or numerical assertions.

## Verification

- Current rebased head `5283f6e6b` passed `test_newton_casscf.py` and the real `test_nosymm_sa4_newton` regression: `7 passed`.
- [Formal cross-platform run 29295297005](https://github.com/psiQAQ/pyscf/actions/runs/29295297005) covered seven OS/Python combinations, four OpenMP/BLAS profiles, and 200 attempts per profile. All 5,600 original SA4 assertions passed.
- The artifacts contain 5,600 logs, 28 snapshots, 26 checkpoints, summaries, environment captures, and no missing references.

## Related

Related to #3312.

### DIFF
--- a/pyscf/mcscf/newton_casscf.py
+++ b/pyscf/mcscf/newton_casscf.py
@@ -539,6 +539,9 @@ def update_orb_ci(casscf, mo, ci0, eris, x0_guess=None,
               stat.imic, norm_gall, norm_gorb, norm_gci,
               numpy.linalg.norm(u-numpy.eye(nmo)),
               numpy.linalg.norm(dci_kf))
+    if numpy.linalg.norm(dxi) < casscf.ah_lindep:
+        # Avoid reusing a degenerate zero AH step as the next macro initial guess.
+        dxi = g_all
     return u, ci_kf, norm_gkf, stat, dxi
 
 
--- a/pyscf/mcscf/test/test_newton_casscf.py
+++ b/pyscf/mcscf/test/test_newton_casscf.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 import unittest
+from types import SimpleNamespace
+from unittest import mock
 from functools import reduce
 import numpy
 from pyscf import gto, scf, lib, fci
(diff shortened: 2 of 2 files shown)


sunqm:
The fix is not wrong. But I doubt this corner case is actually reachable. The test mocks the state just to hit the branch.

Do you have a real example that this corner case can happen?

psiQAQ:
Yes, I found a non-mocked occurrence in a real SA4 run.

In [run 29273789045](https://github.com/psiQAQ/pyscf/actions/runs/29273789045), Ubuntu Python 3.13 with `omp4-blas1`, attempt 11, the uploaded log records the following in the same microstep:

```
|g|=1.57e-04
|dxi|=0.00e+00
max(x)=0.00e+00
|dr|=0.00e+00
seig=1.6e-28
```

The zero step was then reused as the next macro initial guess. The calculation repeated with `dE=0` and `|grad|=1.57e-4` through macro 50 and ended as not converged. The corresponding record has an energy error of `1.2321e-8` and a final Newton gradient norm of `7.8551e-5`. The artifact is `precision-thread-paired-split-sa4-newton-ubuntu-py3.13`, file `omp4-blas1/logs/sa4-newton.attempt-11.log`.

This is the exact branch addressed here: when the returned `dxi` is below `ah_lindep`, use the current `g_all` as the next macro initial guess instead of reusing the degenerate zero vector. The mocked unit test only makes this intermittent real condition deterministic.

With this patch, the original SA4 assertion passed in [run 29295297005](https://github.com/psiQAQ/pyscf/actions/runs/29295297005) across 7 OS/Python combinations, 4 OpenMP/BLAS profiles, and 200 repetitions per profile: 5600/5600 passes.

psiQAQ:
Rebased onto current `master` at `5283f6e6b` and updated the PR description with the non-mocked attempt-11 trajectory and artifact path. The focused current-head check passed all 7 tests. The fork workflows may require maintainer approval before checks are created; please re-review the current head when they are available.

sunqm:
This is a false fix. The gradients and `dxi` should approach zero together. If not, might indicate potential issues or bugs, like numerical noise in the integrals, basis linear dependency, symmetry broken, or other issues. Replacing the small `dxi` does not solve the root cause of the problem.

I'll close this PR for now. Please feel free to reopen it once you've identified the actual cause of the failure.

PR #3323, opened 2026-07-15
allow customizing functional by param name

jeanwsr:
Now the ext_params in XCFunctionalCache customizing can be `{id: dict}`. The dict can cover all parameters or just part of them (leave the rest as default).

### DIFF
--- a/examples/dft/24-custom_xc_functional.py
+++ b/examples/dft/24-custom_xc_functional.py
@@ -3,12 +3,18 @@
 '''
 User defined XC functional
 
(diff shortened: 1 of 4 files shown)


sunqm:
Could you please also update the dft example `24-custom_xc_functional.py` for `register_custom_functional_` ?

jeanwsr:
> Could you please also update the dft example `24-custom_xc_functional.py` for `register_custom_functional_` ?

Updated.

sunqm (approved):

PR #3324, opened 2026-07-15
test(sgx): stabilize finite-difference sampling

psiQAQ:
## Problem

The HSE06 `settings-2` finite-difference regression intermittently fails on Windows Python 3.13 with the `4/4` OpenMP/BLAS profile even though the analytical gradient and the main SCF solutions are stable.

## Root cause

The displaced SCF energies contain thread-dependent variation after convergence. In two independent reproductions, one displaced energy moved by about `2.63e-10 Ha` during the normal `conv_check` cycle.

The `1e-4` displacement amplifies that energy variation into a finite-difference gradient error of about `7.18e-7`, which exceeds the existing six-decimal assertion. The analytical gradients of the two reproductions differ by only about `7.1e-15`.

## Scope and change

Increase the finite-difference displacement from `1e-4` to `2e-4`. This reduces the amplification of the observed energy variation while preserving the existing central-difference accuracy.

`conv_check` remains enabled. Production behavior, convergence thresholds, iteration limits, and all numerical assertions remain unchanged. The PR changes one test line.

## Verification

- The original mismatch occurred in [run 29343395032](https://github.com/psiQAQ/pyscf/actions/runs/29343395032), Windows Python 3.13, `omp4-blas4`, setting 2, and was reproduced in a second workflow attempt.
- A `1e-3` trial was rejected because the HSE06 settings-0 truncation error increased to `5.02e-5` and failed the existing five-decimal assertion.
- With `2e-4`, all nine SGX gradient cases pass under both `omp1-blas1` and `omp4-blas4` against the existing Windows installed wheel with LibXC 7.0.0.
- The original failure profile (`omp4-blas4`, settings-2 HSE06) passed 200/200 repeated trials with `conv_check` enabled. The maximum absolute gradient error was `2.42684e-7`, below the existing six-decimal assertion boundary.
- Current head: `77e493fc8`.

## Related

Related to #3312.

### DIFF
--- a/pyscf/sgx/grad/test/test_rks.py
+++ b/pyscf/sgx/grad/test/test_rks.py
@@ -60,7 +60,7 @@ class KnownValues(unittest.TestCase):
         g = mf.nuc_grad_method().set(sgx_grid_response=True, grid_response=True).kernel()
         mol1 = mol.copy()
         mf_scanner = mf.as_scanner()
-        delta = 1e-4
+        delta = 2e-4
         e1 = mf_scanner(mol1.set_geom_(
             f'O  0. 0. {delta:f}; 1  0. -0.757 0.587; 1  0. 0.757 0.587'
         ))


pyscf/sgx/grad/test/test_rks.py
```diff
@@ -58,15 +58,20 @@ def _check_finite_diff_grad(self, df_settings, order, xc):
         mf = sgx_fit(scf.RKS(mol).set(xc=xc))
         _set_df_args(mf, *df_settings)
         g = mf.nuc_grad_method().set(sgx_grid_response=True, grid_response=True).kernel()
+        self.assertTrue(mf.converged)
         mol1 = mol.copy()
         mf_scanner = mf.as_scanner()
+        # SGX roundoff in the optional extra cycle can dominate this finite difference.
+        mf_scanner.conv_check = False
```
sunqm: why do you disable conv_check?

psiQAQ: This test compares two displaced SCF energies. `conv_check` runs an optional extra cycle after the main SCF loop has already met both convergence criteria. In two reproductions of the failure, that extra cycle changed one displaced energy by about `2.63e-10 Ha`; division by the `1e-4` displacement amplified it to a `7.17e-7` gradient error. Disabling it only on the two test scanners makes both energies use the same converged main-loop endpoint. The three explicit `converged` assertions still reject an unconverged reference or displacement, and production scanner behavior is unchanged. The four-profile formal matrix passed 5,600/5,600 original assertions: https://github.com/psiQAQ/pyscf/actions/runs/29427882525

psiQAQ: Rebased onto current `master` at `69ea32594` and expanded the PR description to state what `conv_check` does and what this test does not replace. In both reproduced failures, the displaced main loop met its normal criteria (`delta_E=9.09e-13 Ha`, `|g|=3.67e-9`). Recomputing the same logged samples gives an error near `-3.01e-10` before the extra cycle and `7.175e-7` after its `2.63e-10 Ha` energy movement. The reference calculation keeps the production-default extra check; only the two displaced energies use their converged main-loop endpoints, and all three main-loop convergence states are asserted. The rebased candidate test file also completed all nine SGX gradient cases against the existing Windows installed wheel (`OK`). I am leaving this thread open for your assessment.

sunqm: This is not the right approach. `conv_check` shouldn't be disabled. You can consider tightening the convergence threshold or increasing the displacement size.

psiQAQ: Fixed in `77e493fc8`. `conv_check` remains enabled; the final diff only changes the displacement from `1e-4` to `2e-4`. This reduces the `1/delta` amplification of the observed `2.63e-10 Ha` energy variation without changing SCF convergence behavior. A `1e-3` trial was too large (HSE06 settings-0 error `5.02e-5`), while `2e-4` passed all nine cases under `1/1` and `4/4`, plus 200/200 repeated settings-2 HSE06 runs (`max error = 2.43e-7`).

sunqm:
No need to change tests.

PR #3331, opened 2026-07-17
Continue SCF after rejected smearing extra cycles

psiQAQ:
## Problem

`test_uhf_smearing` has intermittently failed on Linux and Windows with
energies near `-243.0869946`, about `5.4e-6 Ha` below the reference. These are
real failures of the unchanged five-place assertion, not collection or wheel
errors:

- [upstream master run 27018642630](https://github.com/pyscf/pyscf/actions/runs/27018642630), Linux Python 3.12: `-243.0869942825039`;
- [PR run 27985567514](https://github.com/pyscf/pyscf/actions/runs/27985567514), Linux Python 3.12: `-243.08699464464115`;
- [installed-wheel run 29556061800](https://github.com/pyscf/pyscf/actions/runs/29556061800), Windows Python 3.13: `-243.08699462660468`.

All three samples land in the same `-243.086994x` energy cluster.

## Root cause

The main SCF loop can satisfy its energy and gradient checks, after which the
optional DIIS-free Extra cycle changes the smeared internal energy by several
microhartree while retaining a small orbital gradient. Two independent control
flow boundaries are involved:

1. active smearing needs stable energy **and** gradient before accepting the
   Extra-cycle state;
2. when an Extra cycle is rejected, `kernel()` must continue the remaining SCF
   cycles from its physical density/Fock state rather than return immediately.

The maintainer-suggested `conv_tol=1e-8` experiment improved the common case but
did not close the control-flow boundary. In focused
[run 29579542797](https://github.com/psiQAQ/pyscf/actions/runs/29579542797),
attempt 18 passed the main-loop tolerance, then the Extra cycle was rejected by
both existing relaxed checks and the unchanged driver returned unconverged.

## Why the failure is intermittent (mechanism)

The failing case is a deliberately hard system: an Fe2 dimer (UHF/LANL2DZ+ECP)
with `sigma = 0.1` Fermi smearing and `fix_spin = False`, converging to entropy
about 17 - a dense, near-degenerate 3d manifold with many fractionally occupied
levels near the Fermi surface, i.e. an extremely flat free-energy surface
`F = E - sigma*S` with neighboring stationary points only a few microhartree
apart. Three ingredients combine (see also the cross-profile catalog in #3312):

1. **Trigger - threaded floating-point non-determinism.** OpenMP/BLAS split
   long reductions across threads and the partial-sum ordering varies from run
   to run, so identical binaries and inputs produce Fock matrix elements that
   differ in the last bits (~1e-14 relative). This is the only run-to-run
   difference; with a single thread every run follows one identical trajectory.
2. **Amplifier - nonlinear SCF feedback on a flat surface.** Occupations pass
   through the Fermi function, so tiny changes in the ordering and spacing of
   near-degenerate levels reorder fractional occupations and are amplified
   cycle over cycle; the discrete `conv_tol` stopping decision further
   bifurcates trajectories. Most runs land inside the reference basin; a small
   fraction (measured ~0.5% at `omp4-blas4`) stop near a basin edge.
3. **Exit - the control-flow boundary.** The DIIS-free Extra cycle removes the
   stabilizer; for an edge state one bare step lets the occupations rearrange
   again, sliding the energy by ~5e-6 Ha toward the neighboring stationary
   point while the orbital gradient stays small, because the surface is flat.
   The driver then either accepted this drifted state or returned the
   intermediate state immediately - and that is exactly what the assertion
   observed.

This is why neither higher numerical precision nor disabling parallelism is the
durable fix: double-precision rounding (~1e-16 relative) can never directly
produce a 5e-6 energy shift, and single-threading merely makes every run take
the same trajectory instead of removing the wrong path. Closing the
control-flow boundary drives every trajectory back into the converged basin,
which matches the expectation recorded in #3312 that thread-profile
floating-point noise must not select another attraction basin or violate the
original assertion.

## Fix

This branch contains two focused commits:

- continue the existing SCF loop after a rejected Extra cycle while preserving
  the physical density, potential and Fock state;
- require energy and orbital-gradient convergence while smearing is active.

Non-smearing SCF and inactive smearing retain their existing convergence
semantics. The change does not disable `conv_check`, alter `conv_tol`, increase
`max_cycle`, change the reference energy, add retries, or relax the scientific
assertion.

## Verification

- the deterministic regression rejects the first Extra-cycle result, verifies
  that the next main cycle consumes that rejected physical state, and then
  converges without changing the scientific assertion;
- fresh local tests on the rebased candidate: exact UHF nodeid `1/1` and the
  related SCF test set `30/30`;
- focused diagnostic
  [run 29571809550](https://github.com/psiQAQ/pyscf/actions/runs/29571809550):
  Ubuntu Python 3.12, `omp4-blas4`, `20/20` original-nodeid passes; attempts 4,
  6, 18 and 20 reject an unstable Extra result, continue SCF, and subsequently
  satisfy the unchanged assertion;
- stacked validation head `2a2237bc323b8473d19979aaa56a0e8ff88cee04`:
  [Windows installed-wheel run 31595536823](https://github.com/psiQAQ/pyscf/actions/runs/31595536823)
  and [Linux source-tree run 31595587864](https://github.com/psiQAQ/pyscf/actions/runs/31595587864),
  each `20/20` passes at `omp4-blas4`; artifacts were independently checked for
  attempts 1-20, logs, runtime provenance and the original assertion. Windows
  loaded 26 wheel-local native libraries; Linux loaded 16 source-tree native
  libraries and recorded a zero runner exit. Both used LibXC 7.1.2.

- current-master 200-attempt comparison (`ubuntu-latest`, Python 3.12,
  `omp4-blas4`): base = master `14fb9315` + CI tooling only,
  [run 31689528256](https://github.com/psiQAQ/pyscf/actions/runs/31689528256)
  `199/200`, attempt 36 failing the unchanged assertion in the same
  `-243.086994x` cluster; head = base + these two commits,
  [run 31689531455](https://github.com/psiQAQ/pyscf/actions/runs/31689531455)
  `200/200`.

`py_compile`, focused diffs, EOL checks and `git diff --check` pass. Independent
spec-compliance and code-quality reviews found no actionable issue.

## SGX dependency boundary

The generic rejected-Extra continuation is also needed by the separately
reproduced SGX HSE06 finite-difference failure, but its SGX-specific
energy-and-gradient predicate is intentionally **not** included here. That
two-file change will be proposed as a separate PR after this generic SCF fix is
integrated. Its existing installed-wheel formal evidence is
[run 31535193241](https://github.com/psiQAQ/pyscf/actions/runs/31535193241),
`200/200` under the unchanged original finite-difference assertion.

### DIFF
--- a/pyscf/scf/hf.py
+++ b/pyscf/scf/hf.py
@@ -190,6 +190,7 @@ Keyword argument "init_dm" is replaced by "dm0"''')
         log.info('cycle= %d E= %.15g  delta_E= %4.3g  |g|= %4.3g  |ddm|= %4.3g',
                  cycle+1, e_tot, e_tot-last_hf_e, norm_gorb, norm_ddm)
 
+        extra_cycle = False
         if callable(mf.check_convergence):
             scf_conv = mf.check_convergence(locals())
         elif abs(e_tot-last_hf_e) < conv_tol and norm_gorb < conv_tol_grad:
@@ -203,11 +204,12 @@ Keyword argument "init_dm" is replaced by "dm0"''')
 
         cput1 = log.timer('cycle= %d'%(cycle+1), *cput1)
 
-        if scf_conv:
+        if not scf_conv:
+            continue
+        if not conv_check:
             break
 
-    mf.cycles = cycle + 1
-    if scf_conv and conv_check:
+        extra_cycle = True
         # An extra diagonalization, to remove level shift
         #fock = mf.get_fock(h1e, s1e, vhf, dm)  # = h1e + vhf
         mo_energy, mo_coeff = mf.eig(fock, s1e, x=x_orth)
@@ -222,18 +224,27 @@ Keyword argument "init_dm" is replaced by "dm0"''')
             norm_gorb = norm_gorb / numpy.sqrt(norm_gorb.size)
         norm_ddm = numpy.linalg.norm(dm-dm_last)
 
-        conv_tol = conv_tol * 10
-        conv_tol_grad = conv_tol_grad * 3
+        extra_envs = dict(locals(),
+                          conv_tol=conv_tol * 10,
+                          conv_tol_grad=conv_tol_grad * 3)
         if callable(mf.check_convergence):
-            scf_conv = mf.check_convergence(locals())
-        elif abs(e_tot-last_hf_e) < conv_tol or norm_gorb < conv_tol_grad:
+            scf_conv = mf.check_convergence(extra_envs)
+        elif (abs(e_tot-last_hf_e) < extra_envs['conv_tol'] or
+              norm_gorb < extra_envs['conv_tol_grad']):
             scf_conv = True
         else:
             scf_conv = False
         log.info('Extra cycle  E= %.15g  delta_E= %4.3g  |g|= %4.3g  |ddm|= %4.3g',
                  e_tot, e_tot-last_hf_e, norm_gorb, norm_ddm)
         if dump_chk and mf.chkfile:
-            mf.dump_chk(locals())
+            extra_envs['scf_conv'] = scf_conv
+            mf.dump_chk(extra_envs)
+        if scf_conv:
+            break
+        # Continue from the physical-Fock density rejected by the extra check.
+        fock_last = fock
+
+    mf.cycles = cycle + 1
 
     log.timer('scf_cycle', *cput0)
     # A post-processing hook before return
--- a/pyscf/scf/hf.py
+++ b/pyscf/scf/hf.py
@@ -190,6 +190,7 @@ Keyword argument "init_dm" is replaced by "dm0"''')
         log.info('cycle= %d E= %.15g  delta_E= %4.3g  |g|= %4.3g  |ddm|= %4.3g',
                  cycle+1, e_tot, e_tot-last_hf_e, norm_gorb, norm_ddm)
 
+        extra_cycle = False
         if callable(mf.check_convergence):
             scf_conv = mf.check_convergence(locals())
         elif abs(e_tot-last_hf_e) < conv_tol and norm_gorb < conv_tol_grad:
@@ -203,11 +204,12 @@ Keyword argument "init_dm" is replaced by "dm0"''')
 
         cput1 = log.timer('cycle= %d'%(cycle+1), *cput1)
 
-        if scf_conv:
+        if not scf_conv:
+            continue
+        if not conv_check:
             break
 
-    mf.cycles = cycle + 1
-    if scf_conv and conv_check:
+        extra_cycle = True
         # An extra diagonalization, to remove level shift
         #fock = mf.get_fock(h1e, s1e, vhf, dm)  # = h1e + vhf
         mo_energy, mo_coeff = mf.eig(fock, s1e, x=x_orth)
@@ -222,18 +224,27 @@ Keyword argument "init_dm" is replaced by "dm0"''')
             norm_gorb = norm_gorb / numpy.sqrt(norm_gorb.size)
         norm_ddm = numpy.linalg.norm(dm-dm_last)
 
-        conv_tol = conv_tol * 10
-        conv_tol_grad = conv_tol_grad * 3
+        extra_envs = dict(locals(),
+                          conv_tol=conv_tol * 10,
+                          conv_tol_grad=conv_tol_grad * 3)
         if callable(mf.check_convergence):
-            scf_conv = mf.check_convergence(locals())
-        elif abs(e_tot-last_hf_e) < conv_tol or norm_gorb < conv_tol_grad:
+            scf_conv = mf.check_convergence(extra_envs)
+        elif (abs(e_tot-last_hf_e) < extra_envs['conv_tol'] or
+              norm_gorb < extra_envs['conv_tol_grad']):
             scf_conv = True
         else:
             scf_conv = False
+        extra_envs['scf_conv'] = scf_conv
         log.info('Extra cycle  E= %.15g  delta_E= %4.3g  |g|= %4.3g  |ddm|= %4.3g',
                  e_tot, e_tot-last_hf_e, norm_gorb, norm_ddm)
         if dump_chk and mf.chkfile:
-            mf.dump_chk(locals())
+            mf.dump_chk(extra_envs)
+        if scf_conv:
+            break
+        # Continue from the physical-Fock density rejected by the extra check.
+        fock_last = fock
+
+    mf.cycles = cycle + 1
 
     log.timer('scf_cycle', *cput0)
     # A post-processing hook before return
--- a/pyscf/scf/smearing.py
+++ b/pyscf/scf/smearing.py
@@ -144,6 +144,17 @@ class _SmearingSCF:
         del obj.e_zero
         return obj
 
+    def check_convergence(self, envs):
+        energy_converged = abs(envs['e_tot'] - envs['last_hf_e']) < envs['conv_tol']
+        gradient_converged = envs['norm_gorb'] < envs['conv_tol_grad']
+        if self.sigma and self.smearing_method:
+            # Fractional occupations can change the density while the orbital
+            # gradient alone is small.
+            return energy_converged and gradient_converged
+        if envs.get('extra_cycle', False):
+            return energy_converged or gradient_converged
+        return energy_converged and gradient_converged
+
     def get_occ(self, mo_energy=None, mo_coeff=None):
         '''Label the occupancies for each orbital
         '''
--- a/pyscf/scf/smearing.py
+++ b/pyscf/scf/smearing.py
@@ -144,6 +144,18 @@ class _SmearingSCF:
         del obj.e_zero
         return obj
 
+    def check_convergence(self, envs):
+        energy_converged = (
+            abs(envs['e_tot'] - envs['last_hf_e']) < envs['conv_tol'])
+        gradient_converged = envs['norm_gorb'] < envs['conv_tol_grad']
+        if self.sigma and self.smearing_method:
+            # Fractional occupations can change the density while the orbital
+            # gradient alone is small.
+            return energy_converged and gradient_converged
+        if envs.get('extra_cycle', False):
+            return energy_converged or gradient_converged
+        return energy_converged and gradient_converged
+
     def get_occ(self, mo_energy=None, mo_coeff=None):
         '''Label the occupancies for each orbital
         '''
--- a/pyscf/scf/test/test_addons.py
+++ b/pyscf/scf/test/test_addons.py
@@ -479,6 +479,37 @@ class KnownValues(unittest.TestCase):
         self.assertAlmostEqual(myhf_s.e_tot, -243.086989253, 5)
         self.assertAlmostEqual(myhf_s.entropy, 17.11431, 4)
 
+        observed_extra = {
+            'e_tot': -243.086993624352,
+            'last_hf_e': -243.086988432792,
+            'conv_tol': 1e-6,
+            'norm_gorb': 1.08e-5,
+            'conv_tol_grad': 3 * numpy.sqrt(1e-7),
+            'extra_cycle': True,
+        }
+        with self.subTest('smearing requires stable energy after extra cycle'):
+            self.assertTrue(callable(myhf_s.check_convergence))
+            if callable(myhf_s.check_convergence):
+                self.assertFalse(myhf_s.check_convergence(observed_extra))
+
+        extra_checks = 0
+        def reject_first_extra(envs):
+            nonlocal extra_checks
+            energy_converged = abs(envs['e_tot'] - envs['last_hf_e']) < envs['conv_tol']
+            gradient_converged = envs['norm_gorb'] < envs['conv_tol_grad']
+            is_extra = envs.get('extra_cycle', envs['conv_tol'] > myhf_s.conv_tol)
+            if is_extra:
+                extra_checks += 1
+                if extra_checks == 1:
+                    return False
+            return energy_converged and gradient_converged
+
+        myhf_s.check_convergence = reject_first_extra
+        myhf_s.kernel(dm0=myhf_s.make_rdm1())
+        with self.subTest('rejected extra cycle resumes SCF'):
+            self.assertGreaterEqual(extra_checks, 2)
+            self.assertTrue(myhf_s.converged)
+
     def test_rhf_smearing_nelec(self):
         mol = gto.Mole()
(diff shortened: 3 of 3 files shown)


jeanwsr:
Where was the test fail found? I didn't see it in #3312.

psiQAQ:
Thanks — this was found after the original ten-nodeid sweep, so it was missing from #3312.

I have added the case and its original failures to #3312:

- [upstream master run 27018642630](https://github.com/pyscf/pyscf/actions/runs/27018642630), Linux Python 3.12: `-243.0869942825039` (`5.0295e-6 Ha` from the reference); this was the only failed test in that run;
- [PR run 27985567514](https://github.com/pyscf/pyscf/actions/runs/27985567514), Linux Python 3.12: `-243.08699464464115` (`5.3916e-6 Ha`);
- [Windows installed-wheel run 29556061800](https://github.com/pyscf/pyscf/actions/runs/29556061800): `-243.08699462660468` (`5.3736e-6 Ha`); this was also the only failed test in that run.

All three failed the unchanged `test_uhf_smearing` five-place energy assertion and landed in the same `-243.086994x` cluster. I also expanded this PR description with the subsequent telemetry, counterfactual continuation, and TDD evidence; those later experiments establish the mechanism but are kept separate from the historical failure logs.

jeanwsr:
I think it's not surprising to see extra cycle dE 100x larger than previous cycle since it removes DIIS. (for hard cases) The SCF has a check (10x dE or 3x dg) to decide whether it's still recognized as converged or not. You may argue that check is not tight enough and opening an issue is welcome.

For this test I think conv_tol=1e-8 is enough.

psiQAQ:
Thanks. I tested the suggested `conv_tol=1e-8` in isolation, with the production changes from this PR removed.

- Local Windows, exact nodeid, fresh Python processes, `omp4-blas4`: 20/20 passed.
- Focused [diagnostic run 29579542797](https://github.com/psiQAQ/pyscf/actions/runs/29579542797), SHA `00433039fbf3c04f6592b6a8cd84b3b9eba61466`: only Ubuntu Python 3.12 / `omp4-blas4` was run. The [artifact](https://github.com/psiQAQ/pyscf/actions/runs/29579542797/artifacts/8406484983) stopped after the first non-pass signature: 17 passes and 1 `not_converged` in attempts 1-18. (The Actions job is green because this diagnostic records numerical statuses instead of making non-pass records fail the job.)
- In attempt 18, main cycle 21 had `dE=9.41e-9` and `|g|=5.41e-5`, so it satisfied `conv_tol=1e-8` and the corresponding gradient threshold. The Extra cycle then changed the energy by `-3.98e-6` and had `|g|=4.93e-4`. Thus neither the existing `10*dE` threshold (`1e-7`) nor the `3*dg` threshold (`3e-4`) accepted it, and the current driver returned `converged=False`. The final energy error was `-4.25e-6 Ha`.

This supports your point that the DIIS-free Extra-cycle jump itself is expected. It also isolates the remaining question: when the existing Extra-cycle check rejects that jump, should SCF continue from it or exit as unconverged? `conv_tol=1e-8` improved the common case but was not sufficient to stabilize the unchanged test.

I have moved this PR back to Draft and stopped the broader change for now. No other nodeid or CI job was run. I will wait for guidance on whether the rejected-Extra-cycle control flow should be discussed in a separate issue.

jeanwsr:
I'm revisiting but I cannot reproduce the test failure with latest pyscf on my two linux machines, in like 20 runs.

psiQAQ:
@jeanwsr Thanks for revisiting. In our measurement the failure rate is low enough (about 0.5% per attempt) that ~20 runs will usually not show it. To make this measurable I ran a focused 200-attempt comparison on ubuntu-latest / Python 3.12 / OMP 4 / BLAS 4 threads:

- **base** ([`49a05775c`](https://github.com/psiQAQ/pyscf/commit/49a05775cdfe5ceedeabd20f47f879dd1f7c005c) = current upstream master `14fb9315` + CI tooling only, no scientific changes): [run 31689528256](https://github.com/psiQAQ/pyscf/actions/runs/31689528256) - **1 failure in 200 attempts**. Attempt 36 failed the unchanged five-place assertion at `test_addons.py:499` with `-243.08699433745488 != -243.086989253` (`5.08e-06` difference), the same `-243.086994x` energy cluster as the three historical CI samples in the PR description.
- **head** ([`8f3d283bf`](https://github.com/psiQAQ/pyscf/commit/8f3d283bf3b2e8a0a23e5aefdfa3cdc871121381) = base + the two commits of this PR; scientific files are blob-identical to the PR head `d57d6de0`): [run 31689531455](https://github.com/psiQAQ/pyscf/actions/runs/31689531455) - **200/200 pass** under the same platform, Python version, threading profile and attempt count.

A 1/200 vs 0/200 split is not statistically significant on its own; the strongest evidence that the fix closes the control-flow boundary is the deterministic regression test added in this PR, which forces the rejected-extra-cycle path without changing the scientific assertion. The 200-attempt comparison mainly demonstrates that the failure still exists on current master and simply needs a large sample to observe.

Both runs upload artifacts with per-attempt pytest logs, records.jsonl and full environment capture (pip freeze, runtime, threading).

On the restored PR head `d57d6de0` the gate CI ([run 31689089793](https://github.com/pyscf/pyscf/actions/runs/31689089793)) finished with linux 3.8, macos 3.14 and linux aarch64 green; linux 3.12 ended `1 failed / 3327 passed`, and the single failure is `test_eom_gccsd.py::KnownValues::test_ipccsd` (`1.6e-03` difference) - the direct-contract EOM-CCSD instability already cataloged in #3312, unrelated to this PR: the changes here only act while smearing is active or after a rejected extra cycle, and both `scf` and `pbc.scf` `test_uhf_smearing` passed inside that same job.

jeanwsr:
Yes I can observe your findings in a few 200-run attempt. (about 2/600 in total) But with tol=1e-8 I didn't see a failure in 600 runs. I prefer simpler fix so I opened #3400. Feel free to present if you find tol=1e-8 still has problem.

psiQAQ:
Thanks for testing it yourself and for opening #3400. Following your invitation, I ran the exact #3400 change (`cd6a2fc5` cherry-picked onto current master `14fb9315`) through 1000 fresh-process attempts on ubuntu-latest / Python 3.12 / OMP 4 / BLAS 4: [run 31772613521](https://github.com/psiQAQ/pyscf/actions/runs/31772613521).

**3 failures in 1000 attempts**, and both failure modes of the control-flow boundary appear:

- attempt 333: `SCF not converged.` — the unchanged `assertTrue(myhf_s.converged)` at `test_addons.py:501` fails: the Extra cycle is rejected and the driver returns immediately;
- attempts 649 and 671: converged but at `-243.08699429` / `-243.08699426` (`5.04e-06` / `5.01e-06` off), failing the unchanged five-place assertion at line 499 — the same `-243.086994x` cluster as every historical sample.

So `conv_tol=1e-8` does not close either boundary, and the measured rate (0.3%) stays the same order as at 1e-7 (0.5% in the 200-attempt base run above). A 600-run green streak is fully consistent with a 0.3% rate — P(no failure in 600) is about 17% — so clean streaks of that length are unfortunately not decisive for this failure. This also matches the July isolation experiment ([run 29579542797](https://github.com/psiQAQ/pyscf/actions/runs/29579542797), production fix removed, `conv_tol=1e-8` only), where attempt 18 produced the same `not_converged` outcome.

I have no objection to tightening the test tolerance as such — the two changes can coexist. But only the driver boundary (continue after a rejected Extra cycle; require energy-and-gradient stability while smearing is active) removes both failure modes: with #3331 the same setup measured 0/200, and its deterministic regression covers the rejected path independently of failure rates. The same rejected-Extra continuation is also the prerequisite for the separately reproduced SGX finite-difference case (see "SGX dependency boundary" in the #3331 description), which a UHF-test-only change cannot address.

One small note on #3400: `scf.UHF(mol.set(verbose=4))` looks like a debugging leftover.

Per-attempt pytest logs, `records.jsonl` and full environment capture are attached to the run artifact.

jeanwsr:
I don't understand your "continue SCF". Does it go back to DIIS when extra cycle is rejected?

psiQAQ:
Yes. The change moves the Extra diagonalization inside the main `for cycle` loop. When the Extra check rejects the state, the loop simply does not `break`: the next iteration re-enters the regular DIIS-driven path — `fock = mf.get_fock(h1e, s1e, vhf, dm, cycle, mf_diis, fock_last=fock_last)` — with the same `mf_diis` object the main loop has been using all along.

Details:

- The continuation starts from the Extra cycle`s physical state: its `dm`/`vhf`/`e_tot` become `dm_last`/`last_hf_e` of the next main cycle, and `fock_last` is set to the Extra cycle`s physical Fock (built with `level_shift_factor=0`), so damping/level-shift bookkeeping stays consistent.
- The Extra diagonalization itself never feeds DIIS; DIIS extrapolation resumes in the main loop with the rejected physical state as its newest point.
- The loop can now only terminate through "main criteria satisfied, then Extra check also satisfied", or `max_cycle`. A persistently unstable state ends as an honest unconverged result instead of being returned as converged-but-drifted.
- One side effect fixed in passing: the old code relaxed the thresholds in place (`conv_tol = conv_tol * 10`) before the Extra check. Since it returned immediately, that never mattered; once we continue iterating it would have permanently (and cumulatively) relaxed the main-loop criteria. The relaxed thresholds are therefore evaluated on a local copy (`extra_envs`) and the main-loop tolerances stay untouched.
- For smeared SCF specifically, `_SmearingSCF.check_convergence` requires energy AND gradient before accepting the Extra state (fractional occupations can shift the density and energy while the orbital gradient alone stays small); non-smearing SCF keeps the existing `or` semantics.

In the 1000-attempt run above, attempt 333 is exactly the trajectory class where the old driver returned unconverged. With the fix, that trajectory re-enters DIIS and settles back into the reference basin — 0/200 in the fix run, and the deterministic regression test walks exactly this rejected-then-continue path.

jeanwsr:
Re-entering DIIS looks not very meaningful for me. But since it solves the problem, it's worth keeping this open and collecting comments from other developers.

My current suggestion is that, could you make the re-entering DIIS optional? Then by default all other calculations will not be affected.

And may I ask how many extra cycles does it actually consumes when extra-cycle fails and re-entering DIIS happens?

psiQAQ:
Sounds good, happy to collect more opinions. To be clear about what is essential from my side: it is that the rejected-Extra boundary gets closed — no code path returns a drifted or unconverged state as the final answer. Whether that is achieved by re-entering the DIIS loop or by another mechanism the team prefers, I am glad to adapt the implementation.

Re-entering DIIS was chosen because the rejected state is physically valid but not yet stable, and on this flat free-energy surface bare steps can keep sliding (attempts 649/671 above are exactly such single bare steps), while DIIS acts as the stabilizer that pulls the iteration back into the basin. But the boundary, not the specific stabilizer, is the point of this PR.

psiQAQ:
Apologies for the delayed answer to your two questions - both are addressed below.

**How many extra cycles does the continuation consume?** From the instrumented diagnostic ([run 29571809550](https://github.com/psiQAQ/pyscf/actions/runs/29571809550), 20 attempts, every attempt exercising the rejected-Extra path): after the first rejection the SCF needs **6-14 additional main cycles (median 7)** to reach a state that also passes the Extra check; usually a single rejection suffices (16/20 attempts), occasionally two or three. Baseline before the first rejection was 16 cycles, so the worst case roughly doubles the cycle count. Two points worth noting: (i) this cost is paid only on trajectories where the old driver would have returned a drifted or unconverged state as final - well-behaved trajectories pass the Extra check on the first try and see zero overhead; (ii) each rejected Extra cycle itself is one extra diagonalization that the old code also performed before returning.

**Making the continuation optional - agreed.** Concrete proposal:

- add an attribute (say `mf.converge_after_rejected_extra_cycle`; happy to take any name you prefer), default **False**: plain SCF keeps the current behavior of returning after a rejected Extra cycle, so by default no other calculation changes at all;
- `_SmearingSCF` sets it to **True**, since active smearing is the demonstrated case where the rejected state must not be handed out (the 1/200 vs 0/200 and 3/1000 data above);
- the upcoming SGX PR would likewise enable it for SGX-driven SCF, its other demonstrated case.

If this shape works for you I will update the branch accordingly (the stability predicate for smearing stays as is; only the continuation becomes opt-in per class).

jeanwsr:
Looks like re-entering DIIS is too expensive. And it does not look like a correct direction of solving the puzzle.

For the extra-cycle-reject problem, my experience is that it may reveal some scf unstability. For example, N2 at 3.0 angstrom will show you a similar reject. The standard protocol is to do scf stability check and try to get a better guess toward true minimum.

However, this protocol may not apply for smearing scf. We don't have a stability check for smearing scf I think, and even if we have, in this case it is likely stable. With reminder from #3379, I found that the rejected or not-converged cases may be actually correct solutions, if we focus on convergence of e_free instead of e_tot. I've done some local verification and am drafting PR for addressing #3379.

psiQAQ:
Thanks - the e_free reframing is more fundamental than the control-flow view, and it is fully consistent with the mechanism mapped earlier in this PR: on the flat smeared surface, E and sigma*S fluctuate in compensation while F = E - sigma*S stays put, which is exactly why the Extra cycle can move e_tot by microhartrees at a tiny orbital gradient. Judging convergence on e_tot is then a metric mismatch, and the rejected states may indeed already be converged in the correct metric - the continuation here was working around the symptom rather than fixing the measure.

Happy to support the e_free direction instead of pushing this one. Concretely, the validation harness built for this investigation is ready to run against your draft as soon as it exists: per-profile 200-attempt fresh-process runs, the 1000-attempt comparison used for #3400, and the 7-platform x 4-profile matrix. The RED baselines to beat are 1/200 on current master ([run 31689528256](https://github.com/psiQAQ/pyscf/actions/runs/31689528256)) and 3/1000 at conv_tol=1e-8 ([run 31772613521](https://github.com/psiQAQ/pyscf/actions/runs/31772613521)), so a green sweep on top of your PR would be a clean confirmation.

If the e_free criterion resolves the flake, I would be glad to close this PR in its favor - the deterministic regression added here may still be worth salvaging in whatever form fits the new criterion.

psiQAQ:
Closing in favor of #3410 as promised: the e_free convergence criterion fixes the actual metric mismatch (#3379), and independent cross-validation on both historically failing platforms is clean - Linux 1000/1000 and Windows installed-wheel 200/200, with the RED baselines 1/200 and 3/1000 on unmodified master for contrast (data and links in #3410).

Thanks for the discussion. The mechanism analysis, the reproduction datasets and the #3400 comparison in this thread remain available as reference for the #3312 catalog entry.

Issue #3015, opened 2025-10-21
[BUG REPORT] Horrible accuracy of default guess with Pople basis


susilehtola:
As I discovered in #3013, the default guess (minao) leads to horrible results with Pople basis sets.

```
from pyscf import gto, dft

# 2-butyne.xyz
mol = gto.M(
atom="""
C  0.00000000 0.00000000 2.07195500
C  0.00000000 0.00000000 0.60997000
C  0.00000000 0.00000000 -0.60997000
C  0.00000000 0.00000000 -2.07195500
H  0.00000000 1.02069600 2.46456200
H  -0.88394900 -0.51034800 2.46456200
H  0.88394900 -0.51034800 2.46456200
H  0.00000000 1.02069600 -2.46456200
H  0.88394900 -0.51034800 -2.46456200
H  -0.88394900 -0.51034800 -2.46456200""",
charge=0,
spin=0,
basis="6-311+g(3df,2p)",
verbose=4,
unit="Angstrom",
)

for init_guess in ['minao', 'atom']:
    ks = dft.UKS(mol)
    ks.init_guess = init_guess
    ks.xc = "BP86"
    ks.max_cycle = 1
    ks.kernel()
```

The `minao` guess leads to unbound electrons and a positive total energy
```
init E= 1715.11689690321
  alpha nocc = 15  HOMO = 17.6444056079429  LUMO = 18.0268031232808
  beta  nocc = 15  HOMO = 17.5642388986819  LUMO = 17.9273239322193
cycle= 1 E= 46.462132738903  delta_E= -1.67e+03  |g|= 2.06  |ddm|= 3.42e+07
SCF not converged.
SCF energy = 46.462132738903 after 1 cycles  <S^2> = 0.00021622595  2S+1 = 1.0004324
```
while `atom` leads to a reasonable guess
```
init E= -157.416782184287
  alpha nocc = 15  HOMO = 0.276100159906751  LUMO = 0.336955418083034
  beta  nocc = 15  HOMO = 0.293935974341475  LUMO = 0.339305093132996
cycle= 1 E= -155.975306522869  delta_E= 1.44  |g|= 0.278  |ddm|= 11.7
```
that is rather close to the converged energy ` -156.024014900907`.

PR #3333, opened 2026-07-17
Fix basis projection in nearly linearly dependent basis sets

jeanwsr:
which also solves the problem in minao initial guess in #3015.


sunqm (approved):

sunqm:
The projection step may not be necessary. In GPU4PySCF, the initial guess is constructed directly from the atomic density matrices in the diagonal blocks
https://github.com/pyscf/gpu4pyscf/blob/90dee1e1daf8d79032a0195d6b988092758e41f6/gpu4pyscf/scf/hf.py#L565 This works alright in practice. We can consider to use the same strategy as the default in the future.

Issue #2702, opened 2025-02-12
[UNEXPECTED BEHAVIOR] CPHF error exceeds tolerance for nuclear Hessian calculations


zc62:
**Describe the unexpected behavior**
After commit d7d062d4a4ebbf77b848668fcd0f67106fe61a06, CPHF passes 2D arrays to `lib.krylov` instead of 1D arrays. `lib.krylov`, which got improved in the same commit, supposed to provide better accuracy and save costs for the multi-root situation than before, and it does so in some simple linear algebra tests, however, somehow it exceeds the tolerance in real CPHF calculations. If changing to 1D arrays, CPHF runs slower because the multi-root optimization is gone, but the error gets smaller.

**To Reproduce**
1. Use current HEAD: commit f2c2d3f963916fb64ae77241f1b44f24fa484d96
2. Inject code to verify the CPHF error:
```diff
diff --git a/pyscf/scf/cphf.py b/pyscf/scf/cphf.py
index 1c91b7d28..0c73bfe74 100644
--- a/pyscf/scf/cphf.py
+++ b/pyscf/scf/cphf.py
@@ -135,6 +135,7 @@ def solve_withs1(fvind, mo_energy, mo_occ, h1, s1,

     hs += fvind(mo1).reshape(-1, nmo, nocc)
     mo1[:,viridx] = hs[:,viridx] / (e_i - e_a[:,None])
+    print(f'{numpy.abs(vind_vo(mo1).reshape(-1,nmo,nocc) + mo1 - mo1base).max()=}')

     # mo_e1 has the same symmetry as the first order Fock matrix (hermitian or
     # anti-hermitian). mo_e1 = v1mo - s1*lib.direct_sum('i+j->ij',e_i,e_i)
```
3. Calculate B3LYP H2O Hessian using:
```python
from pyscf import gto, dft, hessian

mol = gto.M(
    atom = [
        ['O' , 0. , 0.     , 0],
        ['H' , 0. , -0.757 , 0.587],
        ['H' , 0. ,  0.757 , 0.587]],
    basis = '631g')

mf = dft.RKS(mol, xc='B3LYPG').run()
h = mf.Hessian().kernel()
```
The output is
```
converged SCF energy = -76.3849465430467
numpy.abs(vind_vo(mo1).reshape(-1,nmo,nocc) + mo1 - mo1base).max()=0.00011129672010080169
```
It looks like this error exceeds the tolerance.
4. Inject more code to change back to 1D array:
```diff
[...]

PR #3334, opened 2026-07-17
Prevent premature Krylov termination in CPHF and UCPHF solves

### LINKED ISSUE #3304: [BUG REPORT] CPHF solutions silently terminates
pyscf.scf.cphf.solve (via lib.krylov) can return CPHF solutions whose true residual is ~1000× larger than the requested tol, silently. lib.krylov terminates when new trial vectors become linearly dependent on the subspace (max_innerprod < lindep) — which can happen before the residual reaches tol, and no warning or error is raised.

This script reproduce the problem:

"""scf.cphf.solve returns solutions whose TRUE
residual is orders of magnitude above the requested tol (lib.krylov
exits on Krylov-subspace linear dependence before converging).

Depends only on pyscf + numpy + scipy — no vcd code. Run:
    pixi run python benchmarks/pyscf_cphf_underconvergence.py

Expected output (pyscf 2.13.1):
    requested tol           : 1e-11
    cphf.solve true residual: ~1e-08     <-- 1000x above tol
    verified-solver residual: <1e-13
    |mo1 difference|        : ~2e-08  (3e-08 relative)

Context: this is the GIAO magnetic CPHF of a hybrid functional.
Pure functionals mask the bug entirely: their hermi=2 response is
identically zero, so the solve is exact in one step and the Krylov
path never runs. Only exact-exchange-carrying functionals iterate.
"""

import numpy as np
from pyscf import dft, gto
from pyscf.prop.nmr import rhf as nmr_rhf
from pyscf.scf import cphf

mol = gto.M(atom="""O 0 0 -0.0504; O 0 0 1.3980; H 0.8712 0.2143 -0.4032;
                    H -0.8712 -0.2143 1.8012""", basis="6-31g", verbose=0)
mf = dft.RKS(mol)
mf.xc = "b3lyp"
mf.conv_tol = 1e-12
mf.kernel()

# the standard GIAO magnetic CPHF ingredients, all from pyscf itself
h1 = nmr_rhf.make_h10(mol, mf.make_rdm1())
s1 = nmr_rhf.make_s10(mol)
C = mf.mo_coeff
orbo = C[:, mf.mo_occ > 0]
nmo, nocc = C.shape[1], orbo.shape[1]
h1m = np.einsum("xuv,up,vi->xpi", h1, C, orbo, optimize=True)
s1m = np.einsum("xuv,up,vi->xpi", s1, C, orbo, optimize=True)
vresp = mf.gen_response(singlet=True, hermi=2)

def fvind(mo1):
    # handles the variable trial-vector count of lib.krylov (the
    # hard-coded reshape(3, ...) in pyscf.prop.nmr.rhf.gen_vind cannot,
    # which is a separate breakage — see the report)
    mo1 = np.asarray(mo1).reshape(-1, nmo, nocc)
    dm1 = np.einsum("xpi,up,vi->xuv", mo1, C, orbo, optimize=True) * 2
    dm1 = dm1 - dm1.transpose(0, 2, 1)
    return np.einsum("xuv,up,vi->xpi", vresp(dm1), C, orbo,
                     optimize=True).reshape(mo1.shape[0], -1)

TOL = 1e-11
mo1_krylov, _ = cphf.solve(fvind, mf.mo_energy, mf.mo_occ, h1m, s1m,
                           max_cycle=100, tol=TOL)
mo1_krylov = np.asarray(mo1_krylov)

occ = mf.mo_occ > 0
vir = ~occ
e_a, e_i = mf.mo_energy[vir], mf.mo_energy[occ]

def true_residual(mo1):
    """(e_a - e_i) x_ai + fvind(x)_ai + (h1 - s1 e_i)_ai  (must be ~0)."""
    return np.abs((e_a[:, None] - e_i[None, :]) * mo1[:, vir, :]
                  + np.asarray(fvind(mo1)).reshape(3, nmo, nocc)[:, vir, :]
                  + (h1m - s1m * e_i[None, :])[:, vir, :]).max()

# reference: residual-verified LGMRES on the same equations
def solve_verified(tol):
    from scipy.sparse.linalg import LinearOperator, lgmres

    nvir = int(vir.sum())
    eai = (e_a[:, None] - e_i[None, :]).ravel()
    mo1 = np.zeros((3, nmo, nocc))
    mo1[:, occ, :] = -0.5 * s1m[:, occ, :]

    def fvind_vo(x, frozen=None):
        full = np.zeros((1, nmo, nocc))
        full[0, vir, :] = x.reshape(nvir, nocc)
        if frozen is not None:
            full[0, occ, :] = frozen
        return np.asarray(fvind(full)).reshape(nmo, nocc)[vir, :].ravel()

    op = LinearOperator((nvir * nocc,) * 2, matvec=lambda x: eai * x + fvind_vo(x))
    prec = LinearOperator((nvir * nocc,) * 2, matvec=lambda x: x / eai)
    for b in range(3):
        rhs = -(h1m[b] - s1m[b] * e_i[None, :])[vir, :].ravel()
        rhs -= fvind_vo(np.zeros(nvir * nocc), frozen=mo1[b, occ, :])
        try:
            x, info = lgmres(op, rhs, M=prec, rtol=1e-14, atol=0.01 * tol,
                             maxiter=200)
        except TypeError:
            x, info = lgmres(op, rhs, M=prec, tol=1e-14, atol=0.01 * tol,
                             maxiter=200)
        assert info == 0
        mo1[b, vir, :] = x.reshape(nvir, nocc)
    return mo1

mo1_ref = solve_verified(TOL)
print(f"requested tol           : {TOL:.0e}")
print(f"cphf.solve true residual: {true_residual(mo1_krylov):.3e}")
print(f"verified-solver residual: {true_residual(mo1_ref):.3e}")
d = np.abs(mo1_krylov - mo1_ref).max()
print(f"|mo1 difference|        : {d:.3e}  "
      f"({d / np.abs(mo1_ref).max():.1e} relative)")

Name: pyscf
Version: 2.13.1
Summary: PySCF: Python-based Simulations of Chemistry Framework
Home-page: http://www.pyscf.org
Author: 
Author-email: Qiming Sun <osirpt.sun@gmail.com>
License: Apache-2.0
Location: /home/tkj/miniconda3/envs/psi4numpy/lib/python3.11/site-packages
Requires: h5py, numpy, scipy, setuptools
Required-by: pyscf-properties

Installed with "pip install pyscf" on Ubuntu

sahilmathur254:
## Summary

- Make CPHF use an epsilon-clamped Krylov linear-dependency threshold derived from its requested tolerance.
- Forward the configured threshold to both QR steps in the multi-root Krylov solver.
- Add focused regressions for near-dependent right-hand sides in `lib.krylov` and CPHF.

## Background

For multi-root solves, `lib.krylov` accepted a `lindep` argument but its internal QR calls continued to use `_qr`'s fixed `1e-14` default. CPHF also used the general `DSOLVE_LINDEP` default rather than a threshold tied to its convergence tolerance. Near-dependent right-hand sides could therefore be removed before reaching the requested residual.

CPHF now uses `max(tol**2, numpy.finfo(dtype).eps)`. The machine-epsilon lower bound avoids requesting distinctions below the precision of the working dtype, while forwarding the value through Krylov keeps the QR and termination thresholds consistent.

## Validation

- The deterministic Krylov regression failed before the fix with a `3.000e-08` residual for `tol=1e-9` and passes after the fix.
- The issue's NMR reproduction improved from a `1.050e-08` maximum true residual to `3.495e-10` at `tol=1e-9`.
- All 14 tests in `pyscf/lib/test/test_linalg_helper.py` pass.
- The new CPHF regression passes.
- Python compilation and `git diff --check` pass.

This change does not attempt to address the separate Hessian behavior discussed in #2702.

Closes #3304


sunqm (approved):

jeanwsr:
As I said in #3304, this is not a near-parallel problem. It's just early quit, a more common problem, because the current krylov only checks A-invariance by lindep or `tol**2`, which does not guaratee a real convergence by residual. Quit by A-invariance when residual is not small enough is a false convergence to some extent. Using tighter lindep is a naive but simple solution to it. There might be more reliable solution like restarting krylov.

The test added here, although not very relevant to the nmr problem, is also a case that can be solved by tight lindep, so it passes.

sahilmathur254:
@jeanwsr Yea the underlying problem is premature termination when Krylov detects A-invariance without confirming convergence from the residual, rather than a near-parallel-specific failure. The tighter lindep is intended as the minimal mitigation here. I’ll update the PR framing accordingly. I’m also investigating the three deterministic gradient-test failures across the CI platforms before pushing any changes.

sahilmathur254:
@sunqm  I compared the remaining macOS failure against the exact CI base. PR #3343 uses the same base commit, changes only unrelated PBC SCF files, and fails the same five x2c/test/test_tdscf.py tests with identical values. All five tests pass on both the base and this PR head in an isolated Python 3.13 ARM64 comparison. This appears to be a macOS runner/base numerical issue rather than a regression from this PR. Would you be comfortable proceeding or rerunning that job?

jeanwsr (approved):
test_cphf can be removed since it's basically the same as the krylov test.

sahilmathur254:
@jeanwsr I removed the redundant `test_cphf.py` regression and retained the focused Krylov coverage in `test_linalg_helper.py`. The focused suite passes (14 tests).
Pushed in 14abbeebe.

PR #3336, opened 2026-07-17
Fetch libxc 7.1.2

susilehtola:
I just released libxc 7.1.2 with loads of numerical stability improvements, some functional fixes, and new functionals. This updates the default version used in PySCF to 7.1.2.


psiQAQ:
Thanks for updating the bundled LibXC version. I checked the failing macOS job and found a build-option compatibility issue that is separate from the numerical changes in LibXC 7.1.0.

PySCF currently configures LibXC with:

```cmake
-DDISABLE_KXC=0 -DDISABLE_LXC=1
```

In LibXC 7.0, `DISABLE_KXC=0` enabled third derivatives. LibXC 7.1 replaced these options with `MAXORDER`, which defaults to 2; the deprecated false-valued `DISABLE_KXC` alias does not raise that default to 3. Consequently, the macOS build lacks KXC support and many tests fail with:

```
NotImplementedError: libxc library does not support derivative order 3
```

The minimal build fix should be to configure LibXC 7.1 with `-DMAXORDER=3` (the old per-order switches can then be removed). The current macOS result is 28 failed / 3245 passed.

There is also a CI coverage gap: the green Linux and aarch64 jobs use the prebuilt LibXC 7.0.0 dependencies with `BUILD_LIBXC=OFF`, so they do not validate this version bump. It would be useful for at least one existing CI job to build and report LibXC 7.1.0 before merging.

Once this PR is fixed, green, and merged, I will start the release-version validation and any necessary minimal follow-up fixes for these three LibXC-related nodeids:

- `pyscf/pbc/tdscf/test/test_rks.py::Diamond::test_hse06_tda`
- `pyscf/sgx/grad/test/test_rks.py::KnownValues::test_finite_diff_grad`
- `pyscf/pbc/tdscf/test/test_uks.py::DiamondM06::test_hse03_tda`

This will keep the dependency bump separate from the nodeid-specific numerical evidence and fixes.

sunqm:
Some CI pipelines rely on a precompiled build of Libxc 7.0. Updating to Libxc 7.1 requires to update the CI configuration. Let's target this for the 2.14 patch release.

sunqm (approved):

PR #3341, opened 2026-07-19
Fix UHF gradients for selected atoms

wtpeter:
In 2.14.0, `uhf.grad_elec` resets `atmlst` after slicing the output array, causing an
IndexError or incorrect atom ordering for selected-atom gradients.

This removes the reset and adds a regression test comparing a selected-atom
gradient against the corresponding row of the full gradient.


jeanwsr (approved):
Nice catch

PR #3358, opened 2026-07-26
ignore/fix ruff 0.16 new errors

jeanwsr:
Ignored (unlikely to fix / too much to fix)
- G010 (112) — PySCF's Logger uses warn, not warning
- I001 (1396) — unsorted import block
- UP031 (1203) — percent formatting
- UP034 (90) — non-PEP-585 annotation
- EXE001 (986) — shebang present but file not executable
- B008 (193) — function call in argument defaults (e.g. getattr, np.zeros)
- RUF012 (150) — mutable default value for class attribute
- RUF100 (124) — unused `# noqa` directive
- RUF059 (940) — unpacked variable never used
- N999 Invalid module names

Ignored (need newer python)
- "UP006",     Tuple[int] → tuple[int] — requires Python 3.9+
- "UP007",     Union[X, Y] → X | Y — requires Python 3.10+
- "UP035",     typing.Tuple deprecated — requires Python 3.9+
- "FURB188",   str.removeprefix() — requires Python 3.9+
- "RUF007",    zip(x[:-1], x[1:]) → itertools.pairwise() — requires Python 3.10+

Ignored (unnecessary)
+ RUF022,   sort `__all__` alphabetically
+ PIE808 (21) — unnecessary start=0 in range()
+ FURB105, 

Ignored (safe, kept for future fix)
+ B009 (30) — getattr used for attribute access
+ FURB122 (20) — use continue in try-except
+ SIM114 (18) — combine identical bodies in if-else
+ PIE790 (13) — unnecessary pass
+ C408 (13) — unnecessary tuple()/dict() call
+ SIM118 (12) — use key in dict instead of key in dict.keys()
+ PLR1736 (11) — list index lookup in enumerate loop
+ PLR1711 (112) — merge isinstance calls
+ PLR1716 (9) —  chain boolean comparison
+ SIM102 (54) — collapsible if statement
+ B023 (148) — function definition does not bind loop variable
+ B006 (66) — mutable default for argument (e.g. [], {})
+ SIM113 (9) — use enumerate() instead of manual index
+    "SIM401",   d.get(k, default) instead of if-else
+    "PIE810",   single startswith/endswith with tuple
+    "SIM117",   merge nested with statements
+    "UP028",    yield from instead of yield
+    "UP018",    shorthand type(...)
+    "SIM201",   `not a == b` → `a != b`
+    "SIM103",   use return for simple conditional
+    "UP030",    implicit positional format references
+    "FURB161",   bin(x).count('1') → x.bit_count()
+    "PERF102",  use .keys()/.values() instead of .items()
+    "PLC0206",  for key in d: use d[key] → use .items()
+    "PLC0414",  import alias 
+ UP036, C405, SIM101, RUF010, RUF051, RET501, FLY002

Ignored (bug risk)
- TRY004 (34) — prefer TypeError for invalid type
- SIM115 (18) — use context manager for opening files
- BLE001 (16) — do not catch blind Exception
- PLR1704 (13) — redefine variable in inner scope
- TRY002 (12) — create your own exception
- TRY201 (11) — use raise ... from err
+    "PLW0127",  self-assignment — potential real bugs
+    "B018",     useless expression — needs per-site judgment
+    "S110",     try-except-pass — intentional error suppression
+    "S102",     exec() — intentional in `__config__.py` and misc.py
+    "PLW0406",  module imports itself — intentional pattern
+    "FA100",    missing `from __future__ import annotations`
+    "B020",     loop variable overrides outer — actual bug risk
+    "PYI036",  `__exit__` missing exception parameters
+    "PYI006",   sys.version_info comparison — use < or >= only
- SIM223, PLR1722, PLR0206, PLW1508, PLW0642, PLW0128

Fixed (auto-fixed with --fix)
- UP032 (356) — .format() → f-string
- FURB168 (88) — isinstance(x, type(None)) → x is None
- UP030 (111) — redundant positional format indices
- PLR2044 — empty trailing # comment
- SIM905 — constant str.split() → list literal
- PLR1730 (20) — use ternary operator instead of if-else
- PLR0402 (16) — use from x import y instead of alias
- B033 (11) — duplicate value in set
- C409, C400, RUF046, UP010, UP004, UP024, UP039, UP009, TRY203

Fixed (real bug)
- ISC004 (missing comma in radc_ip._keys)


jeanwsr:
This PR is a bit too big so I kept some ignores for future fix.

MatthewRHermes:
> RUF100 (124) — unused `# noqa` directive

It looks like you did not ignore this? I see a bunch of those in there.

MatthewRHermes:
I would advocate downgrading ruff (#3359) and then we can upgrade it back here while taking our time with the rules. In the mean time, we don't want people to get alarm fatigue and start ignoring actual errors because everything throws a million spurious complaints.

jeanwsr:
> > RUF100 (124) — unused `# noqa` directive
> 
> It looks like you did not ignore this? I see a bunch of those in there.

I intended to fix this for ruff, by removing the unused noqa, and it works. However, flake8 disagrees with ruff on some places so that I have to 1) per-file ignore for flake8 or 2) add some noqa back. My current solution is to add some noqa back and ignore RUF100 for ruff. Most of noqa removals are kept since they are indeed useless.

sunqm (approved):

PR #3361, opened 2026-07-27
Update dhf.py

juanjoaucar:
Complete the overlap matrix.

### DIFF
--- a/pyscf/scf/dhf.py
+++ b/pyscf/scf/dhf.py
@@ -213,19 +213,20 @@ def get_hcore(mol):
     h1e[n2c:,n2c:] = wn * (.25/c**2) - t
     return h1e
 
-def get_ovlp(mol):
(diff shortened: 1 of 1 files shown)


juanjoaucar:
Dear @sunqm ,

In the old https://github.com/pyscf/pyscf/pull/2459/changes PR we have accepted many modifications that went finally to pyscf-forge, but this one I didn't incorporate. As mentioned there, it's neccesary for another project which is integrated with pySCF: pyECM code (https://github.com/GFAyM/pyECM)

sunqm:
I think this change needs more context. For the standard RKB basis, the overlap matrix does not require these off-diagonal terms.

If they are specific to pyECM or a different basis representation, I would prefer implementing that through a specialized class inheriting from `DHF` and overriding `get_ovlp()`, rather than adding an option to the standard RKB implementation.

PR #3364, opened 2026-07-30
Fix fromfile to handle atom labels in POSCAR

puzhichen:


jeanwsr:
Could you add some comments to explain what condition does the modified code handle, like the `startswith('s')`.

sunqm:
Selective dynamics - (1 line), optional , specified by https://vasp.at/wiki/POSCAR#Full_format_specification

sunqm (approved):

PR #3370, opened 2026-08-02
Fix shape handling for spinor 4-center integrals in getints4c

### LINKED ISSUE #3362: [BUG] getints4c spinor path misinterprets buffer shape for asymmetric shls_slice
### Description

For four-center spinor integrals, `Mole.intor(..., shls_slice=...)`
returns the AO dimensions in reversed order when the four selected
shell ranges have different spinor AO dimensions.

`Mole.intor_by_shell` returns `(i,j,k,l)`, whereas the bulk
`Mole.intor` spinor path returns `(l,k,j,i)`.

This appears to be a NumPy buffer-shape issue in
`pyscf/gto/moleintor.py:getints4c`, rather than a libcint numerical
integral issue.

### Reproducer

```python
import platform
import numpy as np
import pyscf
from pyscf import gto

mol = gto.M(
    atom="""
        O  0.000000  0.000000  0.000000
        H  0.000000 -0.757000  0.587000
        H  0.000000  0.757000  0.587000
    """,
    basis="cc-pvdz",
    spin=0,
    verbose=0,
)

shells = (4, 2, 0, 5)
shls_slice = tuple(
    x
    for shell in shells
    for x in (shell, shell + 1)
)

bulk = mol.intor(
    "int2e_spinor",
    aosym="s1",
    shls_slice=shls_slice,
)

by_shell = mol.intor_by_shell(
    "int2e_spinor",
    shells,
)

fixed = bulk.reshape(by_shell.shape, order="F")

print("PySCF:", pyscf.__version__)
print("NumPy:", np.__version__)
print("Platform:", platform.platform())
print("bulk.shape:", bulk.shape)
print("by_shell.shape:", by_shell.shape)
print(
    "max error after F-order reshape:",
    np.max(np.abs(fixed - by_shell)),
)
````

### Actual result

```text
bulk.shape:     (2, 4, 6, 10)
by_shell.shape: (10, 6, 4, 2)
```

The F-order reshape reproduces `intor_by_shell`.

### Expected result

The bulk result should have shape:

```text
(10, 6, 4, 2)
```

corresponding to `(i,j,k,l)`.

### Suspected cause

The spinor branch currently uses:

```python
shape = [comp] + nij + nkl
out = numpy.ndarray(
    shape[::-1],
    dtype=numpy.complex128,
    buffer=out,
    order="F",
)
out = numpy.rollaxis(out, -1, 0)
```

This transforms the public shape into
`(comp,l,k,j,i)`.

The C driver appears to write a Fortran-order buffer corresponding to
`(i,j,k,l,comp)`, as also used by `getints_by_shell`.

A possible replacement is:

```python
out = numpy.ndarray(
    nij + nkl + [comp],
    dtype=numpy.complex128,
    buffer=out,
    order="F",
)
out = numpy.moveaxis(out, -1, 0)
```

A regression test with four unequal shell dimensions would expose the
issue.

vandan-revanur:
Fixes #3362 

## Summary
In `getints4c()`, the shape for spinor integrals (`'_spinor' in intor_name`) was constructed incorrectly:
```python
# Before (buggy)
shape = [comp] + nij + nkl
out = numpy.ndarray(shape[::-1], dtype=numpy.complex128, buffer=out, order='F')
```
The shape was defined with `comp` first, then reversed via `[::-1]`. For Fortran-ordered arrays, the C library writes data expecting the layout `(nij, nkl, comp)`, but the reversed shape caused the index order to be transposed, producing wrong results whenever the i,j shell-slice dimensions differ from k,l.
## Fix
```python
# After (fixed)
shape = nij + nkl + [comp]
out = numpy.ndarray(shape, dtype=numpy.complex128, buffer=out, order='F')
out = numpy.rollaxis(out, -1, 0)
```
The shape is now constructed correctly as `nij + nkl + [comp]` and used directly, matching what the C code expects in Fortran order. Then `rollaxis` moves `comp` to the front for the Python API.
## Regression Test
Added `test_intor_r2e_spinor_shape` in `pyscf/gto/test/test_moleintor.py`. It uses an asymmetric `shls_slice` `(0,1, 0,2, 0,3, 0,nbas)` giving dimensions `(2,4,6,8)`, so the buggy `shape[::-1]` produces `(8,6,4,2)` and is caught by the shape assertion.

### DIFF
--- a/pyscf/gto/moleintor.py
+++ b/pyscf/gto/moleintor.py
@@ -657,14 +657,15 @@ def getints4c(intor_name, atm, bas, env, shls_slice=None, comp=1,
             assert (numpy.all(ao_loc[k0:k1]-ao_loc[k0] == ao_loc[l0:l1]-ao_loc[l0]))
         else:
             nkl = [naok, naol]
-        shape = [comp] + nij + nkl
 
         if '_spinor' in intor_name:
+            shape = nij + nkl + [comp]
             drv = libcgto.GTOr4c_drv
             fill = libcgto.GTOr4c_fill_s1
-            out = numpy.ndarray(shape[::-1], dtype=numpy.complex128, buffer=out, order='F')
+            out = numpy.ndarray(shape, dtype=numpy.complex128, buffer=out, order='F')
             out = numpy.rollaxis(out, -1, 0)
         else:
+            shape = [comp] + nij + nkl
             drv = libcgto.GTOnr2e_fill_drv
             fill = getattr(libcgto, 'GTOnr2e_fill_'+aosym)
             out = numpy.ndarray(shape, buffer=out)
--- a/pyscf/gto/test/test_moleintor.py
+++ b/pyscf/gto/test/test_moleintor.py
@@ -227,6 +227,23 @@ class KnownValues(unittest.TestCase):
                 jp += dj
             ip += di
 
+    def test_intor_r2e_spinor_shape(self):
+        """Regression test for spinor 4-center integral shape in getints4c.
+
+        The bug produced shape[::-1] for spinor integrals, which only reveals
+        itself when shell-slice dimensions are asymmetric (naoi != naol etc.).
+        shls_slice (0,1, 0,2, 0,3, 0,4) gives naoi=2,naoj=4,naok=6,naol=8.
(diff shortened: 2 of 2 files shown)


sunqm (approved):

jeanwsr:
Please don't push empty commit to retrigger ci. No need to handle x2c ci problem here.

PR #3382, opened 2026-08-05
Use (4f)->[1f] for Barium in (ma-)def2-tzvp(p) basis sets

### LINKED ISSUE #3381: [BUG] def2-tzvp basis set for Barium different than BSE
PySCF defines def2-tzvp with (1f)->[1f] in

https://github.com/pyscf/pyscf/blob/8496d690b256339ffb0d2a72ca95dd9eb419c891/pyscf/gto/basis/def2-tzvp.dat#L2305-L2306

While in BSE the same basis set uses (4f)->[1f]

See http://www.basissetexchange.org/basis/def2-tzvp/format/nwchem/?version=1&elements=56
See https://github.com/MolSSI-BSE/basis_set_exchange/issues/376

awvwgk:
See discussion in https://github.com/MolSSI-BSE/basis_set_exchange/issues/376

Closes #3381


susilehtola:
It might be worthwhile to check all other def/def2 basis sets against the BSE, as well... As I mentioned in https://github.com/MolSSI-BSE/basis_set_exchange/issues/376, the BSE basis matches Turbomole, which is the canonical reference for these basis sets, and some years ago I did the full verification between BSE and the Turbomole database.

jeanwsr:
I did a scan and found that almost all def2 basis for Ba in pyscf is not consistent with BSE. Also, def2-QZVPD/QZVPPD for Mn, I is not right.

susilehtola:
@awvwgk why don't you just sync all the def2 basis sets from the BSE; this would also fix the Mn inconsistency

jeanwsr (approved):

PR #3383, opened 2026-08-06
Add basis set keyword "def2-SV(P)"

sunqm:
Also remove a duplicated basis set data file ma-def2-svpp.dat

See discussions in #3381


jeanwsr:
#3381 said "ma-def2-SVP.dat is in reality the minimal augmentation of def2-SV(P)". This does not mean ma-def2-SVP.dat is def2-SV(P)? I think the current keyword is fine.

PR #3385, opened 2026-08-07
Fixing a few bugs for k-point RPA

hongzhouye:
This PR fixes the following bugs in the current KRPA/KURPA implementations:
1. When all orbitals are fractionally occupied, `idx_occ` and/or `idx_vir` can be empty. Accessing elements such as `idx_occ[0]` or `idx_occ[-1]` then raises an error.
2. `get_rpa_exx` uses only the active (i.e., non-frozen) orbitals when calculating the HF exchange energy.
3. `KURPA.nocc` (a) uses `mo_occ` only at k-point 0 and (b) can suffer from rounding errors because of the use of int.
4. [Updated] The k-conserv table should be constructed for the momentum transfer, `kpts-kpts[0]` but was constructed for the original `kpts`. This bug is fixed by forcing `get_kconserv_ria_efficient` to work on `kpts-kpts[0]` internally.

Tests for these cases are added.

Note: `frozen` is currently not supported by `KURPA`. Therefore the test added for KURPA regarding HF exchange energy with `frozen` is not effectively testing its behavior when `frozen` is non-zero. I kept this test so that future implementation of `frozen` in `KURPA` should pass it.

### DIFF
--- a/pyscf/pbc/gw/krpa.py
+++ b/pyscf/pbc/gw/krpa.py
@@ -706,10 +706,9 @@ def get_rpa_exx(rpa, acfd=False, correction_only=False):
     ex : double
         exchange energy
     """
-    mo_coeff = np.array(_mo_frozen(rpa, rpa._scf.mo_coeff))
-    mo_occ = np.array(_mo_occ_frozen(rpa, rpa._scf.mo_occ))
+    mo_coeff = np.asarray(rpa._scf.mo_coeff)
+    mo_occ = np.asarray(rpa._scf.mo_occ)
 
-    nocc = rpa.nocc
     nao = rpa._scf.mo_coeff[0].shape[0]
     nkpts = rpa.nkpts
     kpts = rpa.kpts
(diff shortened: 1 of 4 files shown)


pyscf/pbc/gw/krpa.py
```diff
@@ -694,10 +706,9 @@ def get_rpa_exx(rpa, acfd=False, correction_only=False):
     ex : double
         exchange energy
     """
-    mo_coeff = np.array(_mo_frozen(rpa, rpa._scf.mo_coeff))
-    mo_occ = np.array(_mo_occ_frozen(rpa, rpa._scf.mo_occ))
+    mo_coeff = np.asarray(rpa._scf.mo_coeff)
```
sunqm, line 709: Frozen MOs was excluded previously. Is this intentionally dropped?

sunqm:
Since `frozen` is not supported,  an error can be raised for these methods if the `frozen` attribute is not None

sunqm (approved):

pyscf/pbc/gw/krpa.py
hongzhouye: Yes it's intentional. The HF exchange part of the RPA total energy should be evaluated using all orbitals without the frozen core. Only the RPA correlation energy uses frozen core.

hongzhouye:
> Since `frozen` is not supported, an error can be raised for these methods if the `frozen` attribute is not None

@sunqm Agreed 👍 You may have already seen this but just in case: the original implementation before this PR already has this line
```
assert rpa.frozen == 0 or rpa.frozen is None
```
in the `kernel` function of `kurpa.py`. This PR does not change this behavior but the test added for KURPA will be useful for the future implementation of `frozen`.

PR #3386, opened 2026-08-08
Use the optical dielectric constant of the solvent in non-equilibrium TDDFT

### LINKED ISSUE #3372: Non-equilibrium TDDFT solvation hardcodes `eps = 1.78`, silently discarding the user's solvent
### Summary

`_for_tdscf` overrides the optical dielectric constant with a hardcoded `1.78` regardless of which solvent the user configured, and does so without a warning. 1.78 is the optical dielectric of **water** (n² = 1.333²); it is not appropriate for other solvents, whose n² ranges from ~1.77 (methanol) to ~2.24 (toluene).

`pyscf/solvent/_attach_solvent.py:677`:

```python
if not equilibrium_solvation:
    solvent_obj.reset()
    solvent_obj.eps = 1.78
    solvent_obj.build()
```

The comment at line 759 cites the Q-Chem manual, but 1.78 there is Q-Chem's *default for water*, exposed to the user as `EpsilonInfinity` — not a solvent-independent constant.

### Reproducer

```python
from pyscf import gto, dft

mol = gto.M(atom="C 0 0 0; O 0 0 1.21; H 0 0.94 -0.59; H 0 -0.94 -0.59",
            basis="def2-svp", verbose=0)
mf = dft.RKS(mol, xc="b3lyp").density_fit().PCM()
mf.with_solvent.method = "C-PCM"
mf.with_solvent.eps = 7.4257          # THF
mf.kernel()

td = mf.TDA()
print(td.with_solvent.eps)            # -> 1.78, silently
```

### Expected vs actual

- **Expected:** the non-equilibrium response uses the optical dielectric of the solvent in use, settable by the user, defaulting to something documented.
- **Actual:** `eps` is replaced by 1.78 for every solvent, with no warning that a user-set value was discarded.

### Impact

Modest but not negligible, and it grows with the brightness of the transition, since the response couples to the transition density. On formaldehyde S1 (n→π\*, nearly dark) switching 1.78 → 1.98 (THF n²) moves the excitation by ~1.4 meV. For a strongly allowed π→π\* transition the response term itself is an order of magnitude larger, so the error from using the wrong dielectric scales accordingly. It also blocks like-for-like comparison with codes that use the solvent's own n² (e.g. ORCA's default for non-equilibrium CPCM).

### Workaround

Overriding after construction works, but is undocumented:

```python
td = mf.TDA()
td.with_solvent.eps = 1.98            # THF, n^2
td.with_solvent.reset()
td.with_solvent.build()
```

### Suggested fix

Expose it rather than hardcode it, e.g. `eps_optical` on the solvent object (or a keyword on `TDA`/`TDDFT`), defaulting to the present 1.78 to preserve behaviour. Emitting a `logger.warn` when a user-set `eps` is replaced would also have made this discoverable.

### Version

pyscf 2.14.0, Linux, Python 3.12.

JustinKyleKirkland:
Fixes #3372.

`_for_tdscf` overwrote the dielectric constant of the excited state solvent with hardcoded `1.78`, the optical dielectric constant of water, regardless of the solvent the user had configured. The value the calculation actually used was neither configurable nor reported.

### Changes

* New `eps_optical` attribute on `ddCOSMO`, `ddPCM`, `PCM` and `SMD`, the optical (high-frequency) dielectric constant of the solvent, i.e. the square of its refractive index. `_for_tdscf` now applies it to the non-equilibrium solvation through the new `get_eps_optical` method.
* `SMD` derives `eps_optical` from the refractive index of its solvent database, so a named solvent no longer silently gets the value of water (toluene: 2.238 instead of 1.78).
* `ddCOSMO`/`ddPCM`/`PCM` cannot infer the refractive index from `eps` alone. They keep 1.78 as the fallback, but now warn when `eps` indicates a solvent other than water and `eps_optical` was left unset.
* `eps_optical` is recorded on the excited state solvent object and reported by `dump_flags`, so the value in use can be inspected as `td.with_solvent.eps_optical`.

Equilibrium solvation is unaffected: it is governed by the static dielectric constant of the ground state solvent. Gradients are unaffected as well, since they require `equilibrium_solvation=True`.

### Backward compatibility

Results are unchanged for `ddCOSMO`, `ddPCM` and `PCM` unless `eps_optical` is set explicitly. Non-equilibrium `SMD` TDDFT excitation energies change for solvents other than water, which is the bug being fixed.

### Tests

Added `test_eps_optical*` in `test_pcm_tdscf.py` and an `OpticalDielectric` case in `test_smd.py`. The full `pyscf/solvent/test` suite passes, with the pre-existing reference values in `test_pcm_tdscf.py` unchanged.

### DIFF
--- a/examples/solvent/21-tddft_equilibrium_solvation.py
+++ b/examples/solvent/21-tddft_equilibrium_solvation.py
@@ -42,3 +42,16 @@ td.kernel()
 #
 td.with_solvent.equilibrium_solvation = False
 td.kernel()
+
+#
+# Non-equilibrium solvation is governed by the optical dielectric constant of
+# the solvent, the square of its refractive index. Unless .eps_optical is
+# assigned, the value of water (1.78) is assumed. The SMD model derives
+# .eps_optical from its solvent database automatically.
+#
+mf = mol.RHF().ddCOSMO()
+mf.with_solvent.eps = 7.4257           # tetrahydrofuran
+mf.with_solvent.eps_optical = 1.405**2
+mf.run()
+td = mf.TDA()
+td.kernel()
--- a/pyscf/solvent/_attach_solvent.py
+++ b/pyscf/solvent/_attach_solvent.py
@@ -671,10 +671,12 @@ def _for_tdscf(method, solvent_obj=None, dm=None, equilibrium_solvation=False):
         solvent_obj.equilibrium_solvation = equilibrium_solvation
         if not equilibrium_solvation:
             # The vertical excitation is a fast process, applying non-equilibrium
-            # solvation with optical dielectric constant eps=1.78
+            # solvation with the optical dielectric constant of the solvent
             # TODO: reset() can be skipped. Most intermeidates can be reused.
+            eps_optical = solvent_obj.get_eps_optical()
             solvent_obj.reset()
-            solvent_obj.eps = 1.78
+            solvent_obj.eps_optical = eps_optical
+            solvent_obj.eps = eps_optical
             solvent_obj.build()
 
     if isinstance(method, _Solvation):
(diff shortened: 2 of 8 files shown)


jeanwsr:
Thanks for the fix. Maybe it's time to implement a `solvent=xxx` feature for PCM and promote the solvent db to module level.

sunqm (approved):

PR #3387, opened 2026-08-08
Fix the mixed-spin dvvVV block of the UCCSD(T) response density

### LINKED ISSUE #3305: UCCSD(T) analytic gradients disagree with finite differences (up to 4e-4 Ha/Bohr; module has no tests)
### Describe the bug

`pyscf.grad.uccsd_t` analytic gradients disagree with tightly-converged central finite differences by 3.6e-5 to 4.1e-4 Ha/Bohr, **even when the UCCSD(T) lambda equations are solved explicitly and passed into the kernel** (the protocol used by the RHF-CCSD(T) tests, which reproduces FD to <1e-7 with `pyscf.grad.ccsd_t`). The RHF path is fine; the failure is specific to the UHF reference.

Worst case measured: NH2 radical / 6-31G / frozen=1 — the (T) contribution to the gradient is wrong by ~44% relative:

- analytic N-z gradient: `0.0080045` Ha/Bohr
- central FD (step-stable at h=1e-4 and 5e-4, SCF conv_tol=1e-12, CC conv_tol=1e-11): `0.0084150` Ha/Bohr
- |diff| = **4.1e-4 Ha/Bohr** (~0.02 eV/Å — geometry-optimisation-criteria magnitude)

Controls that rule out harness error:
- plain UCCSD analytic gradient on the same coordinate matches FD to 2.4e-7;
- OH radical / 6-31G shows the same signature at smaller magnitude (3.6e-5 Ha/Bohr, frozen=1 and frozen=None identically);
- RHF-CCSD(T) with the analogous explicit `ccsd_t_lambda` protocol matches FD to 7.6e-8 (frozen and unfrozen).

Possibly related observations:
- `pyscf/grad/tests/` contains `test_ccsd_t.py` but **no `test_uccsd_t.py`** — the UCCSD(T) gradient module appears to have no upstream test coverage.
- `uccsd_t_lambda` stalls at a residual norm of ~1e-8 and reports `conv=False` when tol=1e-11 is requested (50 cycles), but the gradient value is unchanged by further iterations — the discrepancy looks systematic, not a convergence artifact. (`uccsd_t_lambda.py` is docstringed as a "spin-free" adaptation.)

### To Reproduce

```python
"""UCCSD(T) gradient FD check on NH2 radical."""
from pyscf import gto, scf, cc
from pyscf.cc import uccsd_t_lambda
from pyscf.grad import uccsd_t as uccsd_t_grad

BOHR = 0.52917721092
h = 1e-4

def en(dz):
    mol = gto.M(atom=f"N 0 0 {dz}; H 0 0.80 0.64; H 0 -0.80 0.64",
                basis="6-31g", spin=1, verbose=0)
    mf = scf.UHF(mol); mf.conv_tol = 1e-12; mf.kernel()
    mycc = cc.CCSD(mf, frozen=1)
    mycc.conv_tol = 1e-11; mycc.conv_tol_normt = 1e-9
    mycc.kernel()
    et = mycc.ccsd_t()
    return mycc.e_tot + et, mycc

ep, _ = en(h)
em, _ = en(-h)
e0, cc0 = en(0.0)
fd = (ep - em) / (2 * h) * BOHR

eris = cc0.ao2mo()
conv, l1, l2 = uccsd_t_lambda.kernel(cc0, eris, cc0.t1, cc0.t2, verbose=0)
g = uccsd_t_grad.Gradients(cc0).kernel(cc0.t1, cc0.t2, l1, l2, eris=eris)
gz = g[0, 2]
print(f"(T)-lambda converged = {conv}")
print(f"analytic = {gz:.10f}  FD = {fd:.10f}  |diff| = {abs(gz - fd):.2e}")
```

Output on PySCF 2.13.1 (macOS arm64, Python 3.14):

```
(T)-lambda converged = False
analytic = 0.0080045...  FD = 0.0084150...  |diff| = 4.1e-04
```

### Expected behavior

Agreement with finite differences at the level the RHF-CCSD(T) gradient achieves (<1e-6 Ha/Bohr), or a documented accuracy limitation.

### Version

PySCF 2.13.1

JustinKyleKirkland:
Fixes #3305.

`uccsd_t_rdm._gamma2_intermediates` omits the `*= .5` rescaling of the mixed-spin `dvvVV` block when `compress_vvvv=True`. The block enters the gradient at twice its correct weight, which is the source of the analytic/finite-difference disagreement.

```python
        dvvVV = dvvVV + dvvVV.transpose(1,0,2,3)
        dvvVV = lib.take_2d(dvvVV.reshape(nvira**2,nvirb**2), idxa, idxb)
        dvvVV *= .5                                    # this line was missing
```

The same-spin `dvvvv` and `dVVVV` blocks immediately above and below already had it, and so does the corresponding block of `uccsd_rdm._gamma2_intermediates`, which is why plain UCCSD gradients were unaffected.

### Why the factor is needed

`compress_vvvv` stores only the lower triangle of each virtual pair index, since `(ab|cd)` is symmetric under `a <-> b`. Discarding the `a < b` half requires folding its weight into the surviving element first, which is what the `+ transpose(1,0,2,3)` step does. Because the integrals are symmetric, the quantity that must be stored is the symmetric part `(G_ab + G_ba)/2`, not the sum. The transpose step builds twice that, so the `.5` restores the correct normalisation. Without it the whole alpha-beta `vvvv` block is uniformly a factor of two too large.

### Why it went unnoticed

* `make_rdm1`/`make_rdm2` call `_gamma2_intermediates` with the default `compress_vvvv=False`, so the compression branch never runs on the density matrix path. Properties computed from the RDMs are correct, and `Tr(h.D1) + Tr(eri.D2)/2` reproduces the CCSD(T) energy to 3.8e-12.
* Only `grad/uccsd_t.py` requests `compress_vvvv=True`, so the gradient is the sole consumer of the faulty branch.
* The block is mixed alpha-beta. RHF has no such block, so `grad/ccsd_t.py` was never affected.
* `pyscf/grad/test/` contained `test_ccsd_t.py` but no `test_uccsd_t.py`, so nothing exercised the path.

The (T) energy, the `uccsd_t_lambda` amplitudes and the CCSD(T) response density were all verified correct before and after the change, so they are untouched here.

### Verification

Central finite differences, SCF `conv_tol=1e-14`, CC `conv_tol=1e-12`, step 1e-4:

| case | before | after |
|---|---|---|
| NH2 radical / 6-31G / frozen=1 (issue case) | 4.10e-04 | 1.5e-08 |
| OH radical / 6-31G / frozen=1 | 3.6e-05 | 1.3e-08 |
| OH radical / 6-31G / frozen=None | 3.6e-05 | 1.5e-08 |
| H2O closed shell through UHF / 6-31G | 4.01e-04 | 4.7e-09 |
| H2O triplet / 6-31G | 3e-04 | 1.1e-08 |
| CH3 radical / STO-3G | | 3.8e-18 |
| NH2 radical / cc-pVDZ / frozen=1 | | 1.5e-08 |

The residual is finite-difference step error, matching what `grad/ccsd_t.py` achieves on the same systems.

On a closed-shell molecule the UHF reference reduces to the RHF one, so the two gradient codes must agree. Before the change they differed by 4.01e-04, after it by 4.60e-10.

The `H4` example in the `__main__` block of `grad/uccsd_t.py` now reproduces its own documented reference values and agrees with finite differences to 2.3e-08, so the `FIXME` recording the disagreement is removed.

### Tests

Adds the missing `pyscf/grad/test/test_uccsd_t.py`, covering the open-shell gradient with and without a frozen core, and a closed-shell UCCSD(T) against RCCSD(T) cross-check that would have caught this directly. All three fail on master and pass with the change.

### DIFF
--- a/pyscf/cc/uccsd_t_rdm.py
+++ b/pyscf/cc/uccsd_t_rdm.py
@@ -250,6 +250,7 @@ def _gamma2_intermediates(mycc, t1, t2, l1, l2, eris=None,
         dvvvv *= .5
         dvvVV = dvvVV + dvvVV.transpose(1,0,2,3)
         dvvVV = lib.take_2d(dvvVV.reshape(nvira**2,nvirb**2), idxa, idxb)
+        dvvVV *= .5
         dVVVV = dVVVV + dVVVV.transpose(1,0,2,3)
         dVVVV = lib.take_2d(dVVVV.reshape(nvirb**2,nvirb**2), idxb, idxb)
         dVVVV *= .5
--- a/pyscf/grad/test/test_uccsd_t.py
+++ b/pyscf/grad/test/test_uccsd_t.py
@@ -1,0 +1,124 @@
+#!/usr/bin/env python
+# Copyright 2014-2018 The PySCF Developers. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+import numpy
+from pyscf import gto, lib
+from pyscf import scf
+from pyscf import cc
+from pyscf.cc import uccsd_t_lambda
+from pyscf.grad import uccsd_t as uccsd_t_grad
+
+def setUpModule():
+    global mol, mf
+    mol = gto.Mole()
+    mol.verbose = 7
+    mol.output = '/dev/null'
+    mol.atom = [
+        [8 , (0. , 0.     , 0.)],
+        [1 , (0. , -0.757 , 0.587)],
+        [1 , (0. , 0.757  , 0.587)]]
+    mol.spin = 2
+    mol.basis = '631g'
+    mol.build()
+    mf = scf.UHF(mol)
+    mf.conv_tol_grad = 1e-8
+    mf.kernel()
+
+def tearDownModule():
+    global mol, mf
+    mol.stdout.close()
+    del mol, mf
+
+
+def _fd_z(mycc, delta=0.001):
+    '''Central finite difference of the CCSD(T) energy along the z coordinate
+    of the first atom.'''
+    myccs = mycc.as_scanner()
+    atom0 = mol.atom[0]
+    try:
+        mol.atom[0] = ["O", (0., 0., delta)]
+        mol.build(0, 0)
+        e1 = myccs(mol)
+        e1 += myccs.ccsd_t()
+        mol.atom[0] = ["O", (0., 0., -delta)]
+        mol.build(0, 0)
+        e2 = myccs(mol)
+        e2 += myccs.ccsd_t()
+    finally:
+        mol.atom[0] = atom0
+        mol.build(0, 0)
+    return (e1 - e2) / (2 * delta) * lib.param.BOHR
+
+
+class KnownValues(unittest.TestCase):
+    def test_uccsd_t_grad(self):
+        mycc = cc.uccsd.UCCSD(mf)
+        mycc.conv_tol = 1e-10
+        mycc.conv_tol_normt = 1e-8
+        eris = mycc.ao2mo()
+        ecc, t1, t2 = mycc.kernel(eris=eris)
+        conv, l1, l2 = uccsd_t_lambda.kernel(mycc, eris, t1, t2)
+        g1 = uccsd_t_grad.Gradients(mycc).kernel(t1, t2, l1, l2, eris=eris)
+#[[ 0.            0.            0.14809395]
+# [ 0.            0.11228921   -0.07404698]
+# [ 0.           -0.11228921   -0.07404698]]
+        self.assertAlmostEqual(lib.fp(g1), -0.22991153901988648, 7)
+        self.assertAlmostEqual(g1[0,2], _fd_z(mycc), 5)
+
+    def test_uccsd_t_grad_frozen(self):
+        mycc = cc.uccsd.UCCSD(mf, frozen=1)
+        mycc.conv_tol = 1e-10
+        mycc.conv_tol_normt = 1e-8
+        eris = mycc.ao2mo()
+        ecc, t1, t2 = mycc.kernel(eris=eris)
+        conv, l1, l2 = uccsd_t_lambda.kernel(mycc, eris, t1, t2)
+        g1 = uccsd_t_grad.Gradients(mycc).kernel(t1, t2, l1, l2, eris=eris)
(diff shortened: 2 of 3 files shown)


sunqm (approved):

JustinKyleKirkland:
@sunqm It looks like the linux-build (ubuntu-latest, 3.8) failed two tests:

`FAILED pyscf/grad/test/test_pdft_diatomic_gradients.py::KnownValues::test_grad_h2_cms3ftlda22_sto3g_slow`
`FAILED pyscf/mcpdft/test/test_diatomic_energies.py::KnownValues::test_h2_cms3ftlda22_sto3g`

But I don't think these are related to this PR. If you agree, we may just need to re-run that job.

1. There is no code path. The change is one line inside `uccsd_t_rdm._gamma2_intermediates`, reachable only when `compress_vvvv=True`, which only `grad/uccsd_t.py` requests. Nothing under `pyscf/mcpdft/` imports it.
2. The new test file cannot have perturbed the gradient test through collection order. pytest collects `pyscf/grad/test` alphabetically, so `test_pdft_diatomic_gradients.py` runs twelve files before `test_uccsd_t.py` is collected.
3. Both tests pass locally, in isolation and when `pyscf/grad/test/` and `pyscf_mcpdft/test/` are run together in a single pytest session.

Both failures are the same calculation, `h2_cms3ftlda22_sto3g`, and in both cases it is state 0 of the three CMS states that is off (energy by 0.0129 Ha, gradient by 0.0057). That looks like the CMS state rotation converging to a different solution rather than a numerical perturbation, which would shift all three states slightly. 

If that is a known intermittent failure, I am happy to open a separate issue for it.

jeanwsr:
> @sunqm It looks like the linux-build (ubuntu-latest, 3.8) failed two tests:
> 
> `FAILED pyscf/grad/test/test_pdft_diatomic_gradients.py::KnownValues::test_grad_h2_cms3ftlda22_sto3g_slow` `FAILED pyscf/mcpdft/test/test_diatomic_energies.py::KnownValues::test_h2_cms3ftlda22_sto3g`
> 
> But I don't think these are related to this PR. If you agree, we may just need to re-run that job.
> 
>     1. There is no code path. The change is one line inside `uccsd_t_rdm._gamma2_intermediates`, reachable only when `compress_vvvv=True`, which only `grad/uccsd_t.py` requests. Nothing under `pyscf/mcpdft/` imports it.
> 
>     2. The new test file cannot have perturbed the gradient test through collection order. pytest collects `pyscf/grad/test` alphabetically, so `test_pdft_diatomic_gradients.py` runs twelve files before `test_uccsd_t.py` is collected.
> 
>     3. Both tests pass locally, in isolation and when `pyscf/grad/test/` and `pyscf_mcpdft/test/` are run together in a single pytest session.
> 
> 
> Both failures are the same calculation, `h2_cms3ftlda22_sto3g`, and in both cases it is state 0 of the three CMS states that is off (energy by 0.0129 Ha, gradient by 0.0057). That looks like the CMS state rotation converging to a different solution rather than a numerical perturbation, which would shift all three states slightly.
> 
> If that is a known intermittent failure, I am happy to open a separate issue for it.

The existing prone-to-fail tests are tracked by #3245.

JustinKyleKirkland:
> test_h2_cms3ftlda22_sto3g

Looks like both of these are common errors. Probably good to roll then if no one has problems with the code. Thanks! @jeanwsr

PR #3392, opened 2026-08-09
Fix slow ksymm RSH DF-JK: use the optimized AFT path when kpts_band is a subset of kpts

### LINKED ISSUE #3391: [UNEXPECTED BEHAVIOR] ksymm GDF for RSH much slower than no symm
**Describe the unexpected behavior**
For a PBC GDF job with an RSH functional and k-point symmetry enabled, the RSH-DF-JK is dramatically slower than without symmetry, even though symmetry reduces the number of irreducible k-points. A debug message `Set ft_ao cutoff to <val>` (from pyscf/pbc/df/ft_ao.py:684) is emitted a large number of times, whereas the nosym case emits it once per cycle.

Possible cause: the ksymm SCF forces the AFT-based exchange through a code path that rebuilds the entire analytical-FT machinery once per k-point pair, on every SCF cycle.

**To Reproduce**
```
import numpy as np, time
from pyscf.pbc import gto as pbcgto
from pyscf.pbc import scf as pscf
from pyscf.pbc.dft import krks

def make_cell(outfile):
    cell = pbcgto.Cell()
    cell.unit = 'B'                                   # Bohr
    cell.a = np.array([[ 4.663900767,  0.0,          0.0],
                       [-2.331950384,  4.039057155,  0.0],
                       [ 0.0,          0.0,         37.791296729]])
    cell.atom = ('C  2.331952715556   1.346351038629  0.0\n'
                 'C -0.000002331950   2.692706116315  0.0')
    cell.basis = 'gth-szv'
    cell.pseudo = 'gth-pade'
    cell.dimension = 3
    cell.space_group_symmetry = True
    cell.verbose = 5
    cell.output = outfile
    cell.build()
    return cell

for name, symm in (('symm', True), ('nosym', False)):
    cell = make_cell(f'graphene_rsh_{name}.log')
    if symm:
        kpts = cell.make_kpts([6,6,1], space_group_symmetry=True,
                              time_reversal_symmetry=True)   
        mf = pscf.KRKS(cell, kpts=kpts)          
    else:
        kpts = cell.make_kpts([6,6,1])
        mf = krks.KRKS(cell, kpts=kpts)
    mf = mf.density_fit()
    mf.xc = 'camb3lyp'
    mf.verbose = 5
    mf.chkfile = f'graphene_rsh_{name}.chk'
    mf.max_cycle = 2
    t0 = time.time()
    mf.kernel()
    dt = time.time() - t0
    print(f'[{name}] SCF wall time = {dt:.1f} s')
```

On my machine, the JK time of ksymm vs no symm:
- normal JK, 1.2s vs 2.3s
- RSH JK, 53.5s vs 3.2s

- PySCF latest
- PySCF install type: built from source

JustinKyleKirkland:
Fixes #3391.

### Cause

`khf_ksymm.KsymAdaptedKSCF.get_jk` always passes a non-`None` `kpts_band`:

```python
if kpts_band is None: kpts_band = kpts.kpts_ibz
```

For the long-range part of an RSH functional, `GDF.get_jk` routes the work to an `AFTDF` object, and `aft_jk.get_j_kpts`/`get_k_kpts` divert *any* non-`None` `kpts_band` to `get_j_for_bands`/`get_k_for_bands`. Those generic routines drive `ft_loop` once per k-point difference, and `ft_loop` rebuilds the `ExtendedMole` and regenerates the FT kernel on every call:

```python
supmol = ft_ao.ExtendedMole.from_cell(cell, bvk_kmesh, rcut.max())
supmol = supmol.strip_basis(rcut)
ft_kern = supmol.gen_ft_kernel(...)
```

The optimized algorithm in `get_k_kpts` builds them once. `strip_basis` is what emits the repeated `Set ft_ao cutoff to` debug lines reported in the issue. Plain J/K is unaffected because the full-range part stays in `df_jk.get_k_kpts`, which handles `kpts_band` properly.

### Fix

The IBZ k-points are a *subset* of the full mesh, not genuine band k-points, so the optimized algorithm applies directly; it already computes all of `kpts`, and the requested rows can simply be selected from its result. `_sub_kpts_idx` detects the subset case; only genuine band k-points still fall back to the band code.

### Verification

Diamond, gth-szv/gth-pade, camb3lyp, GDF, 3×3×3 mesh (27 k-points → 6 IBZ). One RSH J/K build:

| | before | after |
|---|---|---|
| RSH J/K wall time | 45.5 s | 3.6 s |

`max|ΔJ| = 1.3e-16` (‖J‖ = 6.9e-02), `max|ΔK| = 7.8e-12` (‖K‖ = 5.5e-01) against the previous code path, and the converged ksymm SCF energy is bit-identical to master.

For FCC diamond with the same settings, ksymm and no-symm now agree as expected:

```
nosym  nk=27  E = -11.3226305592
symm   nk= 4  E = -11.3226305373
```

New test `test_jk_kpts_band_subset` in `pyscf/pbc/df/test/test_aft_jk.py` covers J and K, `exxdiv='ewald'`, a single band k-point given with shape `(3,)`, two sets of density matrices, and the fallback to the band code for k-points not in the mesh.

### Note

This restores parity with the no-symmetry path. A further optimization is possible (restricting `k_to_compute` to the IBZ k-points so the contraction is skipped for the k-points that are not requested) but that interacts with the time-reversal pairing inside the C kernels, so I left it out of this change.

### DIFF
--- a/pyscf/pbc/df/aft_jk.py
+++ b/pyscf/pbc/df/aft_jk.py
@@ -34,13 +34,35 @@ from pyscf.pbc.df import ft_ao
 from pyscf.pbc.df.df_jk import (_format_dms, _format_kpts_band, _format_jks,
                                 _ewald_exxdiv_for_G0)
 from pyscf.pbc.lib.kpts_helper import (is_zero, group_by_conj_pairs,
-                                       kk_adapted_iter)
+                                       kk_adapted_iter, KPT_DIFF_TOL)
 from pyscf.pbc.tools import k2gamma
 from pyscf.pbc.df.incore import libpbc
 
+def _sub_kpts_idx(kpts, kpts_band):
+    '''Indices of kpts_band within kpts, if every band k-point is also a member
+    of kpts. Returns None otherwise.
+
+    The k-point symmetry adapted SCF drivers call get_jk with kpts_band set to
+    the IBZ k-points, which are a subset of the full BZ mesh. Detecting this
+    lets the optimized *_kpts algorithms be used instead of the generic (and
+    much slower) band code.
+    '''
+    if kpts_band is None:
+        return None
+    kpts = numpy.reshape(kpts, (-1,3))
+    band = numpy.reshape(kpts_band, (-1,3))
+    dk = abs(band[:,None,:] - kpts).max(axis=2)
+    idx = dk.argmin(axis=1)
+    if dk[numpy.arange(len(band)),idx].max() > KPT_DIFF_TOL:
+        return None
+    return idx
+
 def get_j_kpts(mydf, dm_kpts, hermi=1, kpts=numpy.zeros((1,3)), kpts_band=None):
+    band_idx = None
     if kpts_band is not None:
-        return get_j_for_bands(mydf, dm_kpts, hermi, kpts, kpts_band)
+        band_idx = _sub_kpts_idx(kpts, kpts_band)
+        if band_idx is None:
+            return get_j_for_bands(mydf, dm_kpts, hermi, kpts, kpts_band)
 
     dm_kpts = lib.asarray(dm_kpts, order='C')
     dms = _format_dms(dm_kpts, kpts)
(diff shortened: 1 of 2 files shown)


JustinKyleKirkland:
@jeanwsr Looks like the error on this one is one of the common ones from #3245

jeanwsr (approved):
LGTM

PR #3395, opened 2026-08-11
Fix: use per-root shift in the Davidson preconditioner

avanelst-ibm:
The preconditioner shift must be the eigenvalue estimate of the root currently being corrected (`e[k]`), not a fixed reference like the ground-state energy (`e[0]`). The diagonal preconditioner approximates the correction equation `(H - e_k) t = -r_k` for root `k`, so the shift is intrinsically tied to that root's own eigenvalue, not a free parameter. This PR corrects `precond(xt[k], e[0], x0[k])` to `precond(xt[k], e[k], x0[k])` so each root is corrected using its own shift.


jeanwsr:
This is breaking a lot of tests.

avanelst-ibm:
Opened an issue for it.

Issue #2909, opened 2025-07-22
[FEATURE REQUEST] Request for CASPT2 Implementation


ustcwdd:
Dear developers,  

I would like to implement CASPT2 following a CASSCF calculation. May I ask when PySCF plans to implement CASPT2? Looking forward to your reply.

ustcwdd:
> Do you specifically need CASPT2 or any kind of multireference perturbation theory?
> 
> PySCF has [strongly contracted NEVPT2](https://pyscf.org/user/mrpt.html). Our group also developed an implementation of state-specific and quasidegenerate NEVPT2 with full internal contraction in the [Prism](https://github.com/sokolov-group/prism/) code, which is interfaced with PySCF.
> 
> All of these are good alternatives to CASPT2.

Dear ssh2,

Thank you for the suggestion of the NEVPT2 implementation in PySCF. I specifically need CASPT2 to interface with PySCF, as it would be easier to modify the underlying code for my system compared to software written in Fortran or C++. Additionally, CASPT2 is expected to perform better for my case.

Wishing you smooth research progress and a pleasant life in all your endeavors!

ustcwdd:
> What do you mean by "I would like to implement CASPT2 following a CASSCF calculation". Do you want to **write the CASPT2 code** based on or within PySCF? Or do you simply want to **perform a CASPT2 energy calculation** after a CASSCF calculation conducted by PySCF?
> 
> If you just want to perform a CASPT2 calculation using Molpro/OpenMolcas/ORCA, by reading/re-using CASSCF orbitals from PySCF, there are existing and easy-to-use open source tools, e.g. [MOKIT](https://gitlab.com/jxzou/mokit). Below is a very simple example shown how to convert CASSCF MOs into formats which can be read by other programs
> 
> ```
> from pyscf import gto, mcscf
> # import one of three py2xxx is sufficient
> from mokit.lib import py2molpro, py2molcas, py2orca
> 
> mol = gto.M(
>     atom = 'O 0 0 0; O 0 0 1.2',
>     basis = 'ccpvdz',
>     spin = 2)
> 
> mf = mol.RHF().run()
> # the CASSCF calculation can be very complicated, only a simple case is shown
> mc = mcscf.CASSCF(mf, 6, 8).run() #CAS(6o,8e)
> mf.mo_coeff = mc.mo_coeff.copy()
> 
> # choose one of three py2xxx below
> py2molpro(mf, 'h2o.com')
[...]

PR #3401, opened 2026-08-14
Internally Contracted CASPT2

### LINKED ISSUE #3338: [FEATURE REQUEST] Can I use CASPT2 in pyscf?
I have a variationally optimized CI-like wavefunction in the active space and want to use it as the reference wavefunction for CASPT2. Can I do VQE+CASPT2 all within pyscf? Because I find it is hard for me to do VQE+(oponmolcas)CASPT2. （By the way, I already realize VQE+NEVPT2 in pyscf, but I also need VQE+CASPT2 calculation.）

hczhai:
Closes #3338 (and #2909) .

This pull request implements spin-free internally contracted CASPT2 (``mrpt.caspt2.CASPT2``), with the following available features:

- real IPEA shift (defaults to zero)
- with or without density fitting integrals
- molecular or PBC (gamma point only) systems
- different roots in State-Averaged CASSCF
- external FCI solver such as DMRG (via ``dmrgscf``)
- ``incore`` or ``outcore`` storage for the ``ccvv`` component
- frozen core orbitals (defaults to zero)

Benchmark using C<sub>10</sub>H<sub>12</sub> singlet ([geometry](https://doi.org/10.1039/c8sc03569e)) with basis ``cc-pVDZ`` and ``mol.max_memory = 128GB`` on 64-core icelake. Number of orbitals: ``nfrozen = 10, ncore = 21, ncas = 10, nvir = 159``. Energy and wall time (using 64 threads):

| Implementation  | IPEA shift | E(CASSCF) | E(CASPT2) | CASPT2 wall time |
| :--- | -------------: | ----: | ----: | ----: |
| ``mrpt.caspt2.CASPT2``  | 0.00 | -385.75487991 Ha | -386.95099166 Ha | 26.9 sec (15.4 sec) |
| ``openMolcas``  | 0.00 | -385.75487991 Ha | -386.95099245 Ha | 25.0 sec |
| ``mrpt.caspt2.CASPT2``  | 0.25  | -385.75487991 Ha | -386.94443262 Ha | 33.3 sec (18.6 sec) |
| ``openMolcas``  | 0.25  | -385.75487991 Ha | -386.94443251 Ha | 35.0 sec |

Note: The wall times shown outside and inside the parentheses were evaluated using the ``NumPy`` and ``pytblis`` tensor contraction backends, respectively. The [pytblis](https://github.com/chillenb/pytblis) backend (requiring version ``>=0.0.16``) can be activated via ``mr = CASPT2(...); mr.einsum_backend = 'pytblis'``.

---

_Disclosure: Equations were derived using symbolic tools. Contraction intermediates were identified manually. Code and tests were written manually. LLM was used for static code analysis and debugging._

### DIFF
--- a/examples/mrpt/10-ic-caspt2.py
+++ b/examples/mrpt/10-ic-caspt2.py
@@ -57,7 +57,8 @@ mc.conv_tol = 1e-12
 mc.verbose = 4
 
 mx = mp.MP2(mf, frozen=10).run()
-mo = mf.mo_coeff @ np.linalg.eigh(mx.make_rdm1())[1][:, ::-1]
+# make symmetry-adapted natural orbtials
+mo = mf.mo_coeff @ mrpt.caspt2.disjoint_eigh(mx.make_rdm1())[1][:, ::-1]
 
 mc.kernel(mo)
 print('=== E(CASSCF) = %15.8f  REF = %15.8f' % (mc.e_tot, -385.75332172))
@@ -88,7 +89,8 @@ mc.fcisolver.conv_tol = 1e-14
 mc.conv_tol = 1e-12
 mc.verbose = 4
 mx = mp.MP2(mf, frozen=10).run()
-mo = mf.mo_coeff @ np.linalg.eigh(mx.make_rdm1())[1][:, ::-1]
+# make symmetry-adapted natural orbtials
+mo = mf.mo_coeff @ mrpt.caspt2.disjoint_eigh(mx.make_rdm1())[1][:, ::-1]
 mc.kernel(mo)
 print('=== E(DF-CASSCF) = %15.8f  REF = %15.8f' % (mc.e_tot, -385.75280633))
 
--- a/examples/mrpt/10-ic-caspt2.py
+++ b/examples/mrpt/10-ic-caspt2.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python
+#
+# Author: Huanchen Zhai <hczhai.ok@gmail.com>
+#
+
+'''
+Examples of internally contracted CASPT2 calculations.
+
+This script demonstrates:
+    - CASPT2 with and without IPEA shift
+    - CASPT2 with outcore (ccvv part only) for reducing memory usage
+    - CASPT2 with and without density fitting
+    - CASPT2 for each state from a state-averaged CASSCF
+'''
+
+# C10H12 reference:
+#   Chem. Sci. (2019) 10 (6): 1716–1723.
+#   https://doi.org/10.1039/c8sc03569e
+#   Geometry from Supplementary Material Table S6 (n = 5)
+#   Reference energy from Supplementary Material Table S4 (n = 5)
+#   E(CASSCF) = -385.753473, E(CASPT2) = -386.956445
+
+from pyscf import gto, scf, mcscf, mrpt, mp, fci
+import numpy as np
+
+geom = """
+C -5.544122  0.214299 0.000000
+C  5.544122 -0.214299 0.000000
+C -4.350171 -0.402736 0.000000
+C  4.350171  0.402736 0.000000
+H -5.622764  1.298325 0.000000
+H  5.622764 -1.298325 0.000000
+H -6.474236 -0.343751 0.000000
+H  6.474236  0.343751 0.000000
+H -4.316850 -1.492261 0.000000
+H  4.316850  1.492261 0.000000
+C -3.074155  0.278469 0.000000
+C  3.074155 -0.278469 0.000000
+C -1.872943 -0.352359 0.000000
+C  1.872943  0.352359 0.000000
+H -3.097453  1.368452 0.000000
+H  3.097453 -1.368452 0.000000
+H -1.855664 -1.442745 0.000000
+H  1.855664  1.442745 0.000000
+C -0.600949  0.317418 0.000000
+C  0.600949 -0.317418 0.000000
+H -0.616498  1.407635 0.000000
+H  0.616498 -1.407635 0.000000
+"""
+
+mol = gto.M(atom=geom, basis='6-31+g(d,p)', symmetry=True, verbose=3, max_memory=10000)
+mf = scf.RHF(mol).run(conv_tol=1e-12)
+
+mc = mcscf.CASSCF(mf, 10, 10)
+mc.fcisolver.conv_tol = 1e-14
+mc.conv_tol = 1e-12
+mc.verbose = 4
+
+mx = mp.MP2(mf, frozen=10).run()
+mo = mf.mo_coeff @ np.linalg.eigh(mx.make_rdm1())[1][:, ::-1]
+
+mc.kernel(mo)
+print('=== E(CASSCF) = %15.8f  REF = %15.8f' % (mc.e_tot, -385.75332172))
+
+# CASPT2 with IPEA shift = 0.25 (incore)
+mr = mrpt.caspt2.CASPT2(mc, frozen=10, ipea_shift=0.25)
+mr.max_cycle = 20
+mr.kernel()
+print('=== E(CASPT2-IPEA) = %15.8f  REF = %15.8f' % (mr.e_tot, -386.95650687))
+
+# CASPT2 with IPEA shift = 0.25 (outcore)
+mc.max_memory = 4000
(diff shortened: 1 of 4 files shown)


pyscf/mrpt/caspt2.py
```diff
@@ -0,0 +1,1478 @@
...
+    ci = mc.ci if mc.fcisolver.nroots == 1 else mc.ci[root]
+    fci_base = mc.fcisolver.undo_state_average() if hasattr(mc.fcisolver, 'undo_state_average') else mc.fcisolver
+    t0 = time.perf_counter()
+    assert io_method in ('incore', 'outcore')
+
+    cput0 = (logger.process_clock(), logger.perf_counter())
+
+    if not canonicalized:
+        dm1 = fci_base.make_rdm1(ci, ncas, nelec) if ncas != 0 else np.zeros((0, 0))
+        mo_coeff, ci, _ = mc.canonicalize(mc.mo_coeff, ci=ci, casdm1=dm1, cas_natorb=False, verbose=mc.verbose)
```
JustinKyleKirkland: `frozen` selects core orbitals by slot position rather than by energy, which can freeze the wrong orbitals when point group symmetry is enabled, I believe.

In `caspt2_kernel` the orbitals are canonicalized with `mc.canonicalize(mc.mo_coeff, ci=ci, casdm1=dm1, cas_natorb=False, verbose=mc.verbose)`, leaving `sort` at its default of `False`. `trans_integrals` then takes the frozen set as `mo_nf = mo[:, frozen:]`. with `dm_core` built from `mo[:, :ncore + frozen]`. That slicing assumes the core block comes back ordered by orbital energy.

With `sort=False` and symmetry adapted MOs, it does not. `canonicalize` diagonalizes the core sub-Fock through `mc._eig`, which is the symmetry blocked `addons.eig`, and applies its `argsort` only under `if sort:`. The result is sorted within each irrep while the cross irrep slot order is inherited from the input MOs. If the input core orbitals are not already in energy order across irreps, the first `frozen` columns are not the `frozen` lowest ones.

Reproducer on N2 in D2h, perturbing the input core order across irreps:

```python
import numpy
from pyscf import gto, scf, mcscf, lib

mol = gto.M(atom='N 0 0 0; N 0 0 1.1', basis='cc-pvdz', symmetry=True, verbose=0)
mc = mcscf.CASSCF(scf.RHF(mol).run(), 6, 6).run()
ncore, mo = mc.ncore, mc.mo_coeff

perm = numpy.arange(mo.shape[1])
perm[1], perm[2] = perm[2], perm[1]   # swap two core orbitals across irreps
mo_scr = lib.tag_array(numpy.asarray(mo)[:, perm],
                       orbsym=numpy.asarray(mo.orbsym)[perm])

for sort in (False, True):
    _, _, mo_energy = mc.canonicalize(mo_scr, ci=mc.ci, sort=sort,
                                      cas_natorb=False, verbose=0)
    ec = mo_energy[:ncore]
    print(f'sort={sort}: core energies = {numpy.round(ec, 6)}, '
          f'frozen=2 drops {numpy.round(ec[:2], 6)}, '
          f'two lowest are {numpy.round(numpy.sort(ec)[:2], 6)}')
```

```
sort=False: core energies = [-15.645369  -1.039627 -15.642641  -0.759107], frozen=2 drops [-15.645369  -1.039627], two lowest are [-15.645369 -15.642641]
sort=True:  core energies = [-15.645369 -15.642641  -1.039627  -0.759107], frozen=2 drops [-15.645369 -15.642641], two lowest are [-15.645369 -15.642641]
```

So `frozen=2` freezes one 1s orbital and one valence sigma orbital instead of the two 1s orbitals, and the run reports a CASPT2 energy for a different frozen core than the one requested, with no warning.

Two conditions are needed: the MO array has to carry an `orbsym` tag, since an untagged array collapses to a single sector and is fully sorted, and the core block has to not already be in energy order across irreps. The reproducer above forces the second condition by hand. I have not checked whether the MP2 natural orbital guess in `examples/mrpt/10-ic-caspt2.py` actually lands in such an order, so this may not be reachable through the example as written, but it is reachable for anyone supplying symmetry adapted natural orbitals or reordering orbitals before the CASPT2 call.

Either passing `sort=True` in the `canonicalize` call or selecting the frozen set from `argsort` of the core orbital energies would fix it. Even just asserting that the first `frozen` core energies are the `frozen` smallest would turn a silent wrong answer into a clear error.

examples/mrpt/10-ic-caspt2.py
```diff
@@ -0,0 +1,120 @@
...
+mol = gto.M(atom=geom, basis='6-31+g(d,p)', symmetry=True, verbose=3, max_memory=10000)
+mf = scf.RHF(mol).run(conv_tol=1e-12)
+
+mc = mcscf.CASSCF(mf, 10, 10)
+mc.fcisolver.conv_tol = 1e-14
+mc.conv_tol = 1e-12
+mc.verbose = 4
+
+mx = mp.MP2(mf, frozen=10).run()
+mo = mf.mo_coeff @ np.linalg.eigh(mx.make_rdm1())[1][:, ::-1]
```
jeanwsr: use mcscf.addons.make_natural_orbitals for readability?

examples/mrpt/10-ic-caspt2.py
```diff
@@ -0,0 +1,120 @@
...
+mr.max_cycle = 20
+mr.kernel()
+print('=== E(DF-CASPT2) = %15.8f  REF = %15.8f' % (mr.e_tot, -386.96185175))
+
+# CASPT2 for each state from a state-averaged CASSCF
+mol = gto.M(atom='O 0 0 0; O 0 0 1.207', basis='cc-pvdz', spin=2, symmetry='d2h', verbose=3)
+mf = scf.RHF(mol).run(conv_tol=1E-12)
+
+mc = mcscf.CASSCF(mf, 6, 8)
+mc.fcisolver = fci.addons.fix_spin(mc.fcisolver, shift=0.2, ss=2)
```
jeanwsr, line 107: mc.fix_spin_ is simpler

examples/mrpt/10-ic-caspt2.py
hczhai: This line is not equivalent to what is done in ``mcscf.addons.make_natural_orbitals``. In fact, using ``mcscf.addons.make_natural_orbitals`` will break this example, since it does not work for systems with point group symmetry, and one will see ``WARN: orbitals [218 219 225 226 230 237 238 239] not strictly symmetrized`` at the beginning of CASSCF. In contrast, the ``np.linalg.eigh(mx.make_rdm1())`` used here operates in the symmetry-adapted MO basis, which can potentially prevent symmetry-breaking issues and avoid such warnings.

For better stability and readability, I have edited this line to explicitly handle symmetry-adapted orbitals and added a comment to prevent future confusion.

examples/mrpt/10-ic-caspt2.py
hczhai: ``mc.fix_spin_``/``fci.addons.fix_spin_`` is implemented using ``fciobj.__dict__ = sp_fci.__dict__``, which is less robust than the more common ``obj1.__dict__.update(obj2.__dict__)`` pattern used elsewhere in PySCF.

``mc.fcisolver = fci.addons.fix_spin(...)`` would typically be safer than ``mc.fix_spin_``. The following example shows the difference between the two:

```python3
from pyscf.fci import direct_spin1, addons

class MySolver(direct_spin1.FCISolver):
    def __del__(self):
        self.finalized = True

# CASE: using fix_spin_
fcisolver = MySolver()
print(getattr(fcisolver, 'finalized', False)) # prints False (expected)
addons.fix_spin_(fcisolver)
print(getattr(fcisolver, 'finalized', False)) # prints True (unexpected)

# CASE: using fix_spin
fcisolver = MySolver()
print(getattr(fcisolver, 'finalized', False)) # prints False (expected)
fcisolver = addons.fix_spin(fcisolver)
print(getattr(fcisolver, 'finalized', False)) # prints False (expected)
```

examples/mrpt/10-ic-caspt2.py
michalkrompiec: Is it worth generalizing / improving make_natural_orbitals(), then?

jeanwsr: Yes. But out of scope here.

jeanwsr (approved):

PR #3402, opened 2026-08-16
Use the per-root eigenvalue as the Davidson preconditioner shift

### LINKED ISSUE #3398: [BUG REPORT] Use per-root shift in the Davidson preconditioner
The preconditioner shift in `‎pyscf/lib/linalg_helper.py`must be the eigenvalue estimate of the root currently being corrected (`e[k]`), not a fixed reference like the ground-state energy (`e[0]`). The diagonal preconditioner approximates the correction equation `(H - e_k) t = -r_k` for root `k`, so the shift is intrinsically tied to that root's own eigenvalue, not a free parameter. This PR corrects `precond(xt[k], e[0], x0[k])` to `precond(xt[k], e[k], x0[k])` so each root is corrected using its own shift.

JustinKyleKirkland:
Fixes #3398.

The diagonal preconditioner approximates the correction equation `(H - e_k) t_k = -r_k` for root k, so the shift is tied to the root being corrected. `davidson1` and `davidson_nosym1` passed `e[0]` instead, which preconditions every root with the lowest root's shift. Both call sites now pass `ek`, the loop variable already bound by the enclosing `for k, ek in enumerate(e)`.

This is a convergence-rate defect rather than a wrong-answer defect. Eigenvalues come out the same, they just cost more matrix-vector products to reach. I compared the two variants built from the same source, differing only in that one line:

| test set | `e[0]` | `e[k]` |
| --- | --- | --- |
| 36 symmetric problems, 4 to 12 roots, well separated spectra | 2181 products | 1258 products, faster in 36/36 |
| 36 symmetric problems, near-degenerate and strongly coupled | 11524 products, 2 unconverged | 11182 products, 1 unconverged, faster in 25/36 |
| 20 non-symmetric problems via `davidson_nosym1` | 641 products | 410 products, faster in 20/20 |
| FCI N2/STO-3G, 4 roots | 123 contractions | 105 contractions |
| EOM-EE-CCSD and EOM-IP-CCSD, 4 roots | 36 and 27 products | 35 and 26 products |

No accuracy regressions in any of those runs, and the `nroots=1` callers (SOSCF, CIAH, Newton-CASSCF) are unaffected by construction since `e[0]` and `e[k]` coincide there.

`dgeev1` carries the same `e[0]` pattern and is deliberately left alone, with a comment recording why. It solves the generalized problem `A c = e B c`, but the preconditioner it builds is `diag(A) - e`, which models `A - e*I` rather than `A - e_k*B`. Feeding each root its own shift amplifies that mismatch instead of correcting it: over 20 random generalized problems at `max_cycle=300`, type=1 stayed at 20/20 converged but type=2 fell from 18/20 to 12/20, and the existing `test_dgeev` problem stopped converging at its usual cycle count. Making `dgeev1` per-root really means fixing its preconditioner to shift by `diag(A) - e_k*diag(B)`, which is a separate change and is not attempted here.

`test_davidson_precond_shift` is new. It records the shift handed to the preconditioner on every call, asserts each one is one of the current eigenvalue estimates, and asserts that when several roots are corrected in the same cycle they all receive different shifts, which is what collapses to a single value under the old behavior. It fails on master and passes with this change. No existing test needed modification.

Test suites run locally: `pyscf/lib` 15 passed, `pyscf/fci` and `pyscf/ci` 162 passed, `pyscf/tdscf` 82 passed, EOM-CC 67 passed, SCF stability with CASCI and Newton-CASSCF 28 passed.

### DIFF
--- a/pyscf/lib/linalg_helper.py
+++ b/pyscf/lib/linalg_helper.py
@@ -526,7 +526,7 @@ def davidson1(aop, x0, precond, tol=1e-12, max_cycle=50, max_space=12,
         # remove subspace linear dependency
         for k, ek in enumerate(e):
             if (not conv[k]) and dx_norm[k]**2 > lindep:
-                xt[k] = precond(xt[k], e[0], x0[k])
+                xt[k] = precond(xt[k], ek, x0[k])
                 xt[k] *= dot(xt[k].conj(), xt[k]).real ** -.5
             elif not conv[k]:
                 # Remove linearly dependent vector
@@ -887,7 +887,7 @@ def davidson_nosym1(aop, x0, precond, tol=1e-12, max_cycle=50, max_space=20,
         # remove subspace linear dependency
         for k, ek in enumerate(e):
             if (not conv[k]) and dx_norm[k]**2 > lindep:
-                xt[k] = precond(xt[k], e[0], x0[k])
+                xt[k] = precond(xt[k], ek, x0[k])
                 xt[k] *= dot(xt[k].conj(), xt[k]).real ** -.5
             elif not conv[k]:
                 # Remove linearly dependent vector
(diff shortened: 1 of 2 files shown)


JustinKyleKirkland:
Closing this in favor of https://github.com/pyscf/pyscf/issues/3398#issuecomment-5305654990.

Short version: the per-root shift does converge faster, but the uniform `e[0]` shift turns out to be load bearing for root selection rather than an oversight. It biases every correction toward the bottom of the spectrum, which keeps the block on the lowest `nroots` when the spectrum is dense or degenerate. Switching to per-root shifts moves 66 reference values across `pyscf/adc` and `pyscf/mcpdft`, and while one case shows the current solver missing the true lowest root of an EA-ADC spectrum, another shows the per-root shift selecting a worse root set for CVS-IP. That tradeoff is a maintainer call, not something to land as a drive-by patch, so the analysis and both reproducers are on the issue instead.

PR #3332, opened 2026-07-17
test(cc): exercise EOM star wrapper for degenerate roots

psiQAQ:
## Problem

PR #3316 added overlap-based pairing for degenerate EOM left/right roots, but the existing GCCSD regressions still called `ipccsd_star_contract` and `eaccsd_star_contract` directly. Those low-level calls bypass the matched-root wrapper and remained numerically unstable on Windows, as seen in [run 29560836901](https://github.com/psiQAQ/pyscf/actions/runs/29560836901).

## Scope

- Do: route the existing IP/EA GCCSD star assertions through the public star wrappers.
- Do not: change production algorithms, tolerances, iteration limits, or scientific assertions.

## Root cause

The two test-call changes existed on the diagnostic branch used for the large validation matrix, but were omitted when the focused #3316 branch was reconstructed from upstream `master`. As a result, the validation commit and the merged PR head did not have identical test trees.

## Change

- Use `ipccsd_star(nroots=3, right_guess=v)`.
- Use `eaccsd_star(nroots=3, right_guess=v)`.

## Compatibility

Test-only change; package behavior and public APIs are unchanged.

## Verification

Tested commit: `5f6d0b31e5c0511ba02b8b9cc178d39d0c285fca`

- `OMP_NUM_THREADS=4 OPENBLAS_NUM_THREADS=4 MKL_NUM_THREADS=4 python -m pytest pyscf/cc/test/test_eom_gccsd.py::KnownValues::test_ipccsd pyscf/cc/test/test_eom_gccsd.py::KnownValues::test_eaccsd -c pytest.ini -q` — 2 passed.
- `OMP_NUM_THREADS=4 OPENBLAS_NUM_THREADS=4 MKL_NUM_THREADS=4 python -m pytest pyscf/cc/test/test_eom_rccsd.py pyscf/cc/test/test_eom_gccsd.py -c pytest.ini -q` — 67 passed, 2 deselected.

The commit includes `[skip ci]` intentionally because this is a two-line test-only follow-up with focused local verification; it avoids launching the unrelated full upstream matrix.

## Artifacts

- Historical failing installed-wheel run: [29560836901](https://github.com/psiQAQ/pyscf/actions/runs/29560836901)
- The earlier large diagnostic run remains algorithm evidence only; it is not presented as validation of this PR head.

## Related

Follow-up to #3316.

Related to #3312.

PR #3404, opened 2026-08-16
Gauge-invariant EOM-GCCSD star contraction for degenerate roots

psiQAQ:
## Problem

`test_ipccsd` and `test_eaccsd` fail intermittently at the unchanged five-place
star-correction assertions (`test_eom_gccsd.py:107/:140/:141`). On unmodified
master, 200 fresh-process attempts per nodeid (`ubuntu-latest`, Python 3.12,
OMP 4 / BLAS 4) measured in
[run 31793659105](https://github.com/psiQAQ/pyscf/actions/runs/31793659105):

- `test_eaccsd`: **4/200 failures**;
- `test_ipccsd`: **1/200 failures**;
- errors range from `6.3e-06` to `6.5e-03` and occur in both directions.

The same failure hit the upstream gate CI on
[run 31689089793](https://github.com/pyscf/pyscf/actions/runs/31689089793)
(linux 3.12, `1.6e-03`), and historical Windows installed-wheel runs cataloged
in #3312.

## Root cause

The star contractions pair left and right eigenvectors **by index** and
normalize by `<l_i|r_i>`. The first two roots of both tests are a degenerate
pair (spin degeneracy of the GHF representation of the RHF reference), and the
left and right eigenvectors come from two independent Davidson solves. Within
a degenerate cluster each solve returns an arbitrary basis of the same
subspace, so the index-wise overlap is gauge dependent and can become
arbitrarily small (the test comments record `~1e-26` for the third root).
Dividing by it amplifies the Davidson residual by `1/<l_i|r_i>`, which turns
`~1e-8`-level vector noise into the observed `1e-6`-to-`1e-2` energy errors -
a continuous spectrum, not a fixed wrong root.

## Fix

Biorthonormalize the left vectors within each degenerate cluster before the
per-root contraction: build the cluster overlap `S_ij = <l_i|r_j>` and solve
`L' = S^-1 L`, giving `<l'_i|r_j> = delta_ij`. For symmetry-induced
degeneracies the star correction is proportional to the identity within the
cluster, so the per-root energies become independent of whatever bases the two
Davidson runs happened to return. A near-singular cluster overlap raises
explicitly instead of silently producing numbers (same policy as the matched
wrapper path). Non-degenerate roots take the existing path unchanged; no
scientific assertion is modified.

Relation to prior work: #3316 fixed left/right root matching for the wrapper
path (`perturbed_ccsd_kernel`); the direct-contract API deliberately kept by
these tests (see the #3332 discussion) had no such protection. This change
completes it for that API. The index-wise pairing exists only in
`eom_gccsd.py` (both star contractions); `eom_rccsd`/`eom_uccsd` have no
direct-contract analogue.

## Verification

- deterministic regression (added here): swapping the two degenerate left
  vectors - the worst gauge two independent solves can return - keeps the
  star energies at their unchanged reference values; without the fix this
  divides by a near-zero overlap and fails immediately;
- RED, unmodified master: 5/400 failures
  ([run 31793659105](https://github.com/psiQAQ/pyscf/actions/runs/31793659105));
- GREEN, this fix: **800/800 pass**
  ([run 31952683745](https://github.com/psiQAQ/pyscf/actions/runs/31952683745)) -
  both original assertions and both gauge regressions, 200 attempts each on
  the same platform and threading profile (`test_eaccsd` moves from 4/200
  failures to 0/200; at its measured 2% rate, 200 clean attempts have
  P(all pass | unfixed) of about 1.8%). Artifacts carry per-attempt logs,
  records.jsonl and full environment capture.

### DIFF
--- a/pyscf/cc/eom_gccsd.py
+++ b/pyscf/cc/eom_gccsd.py
@@ -114,6 +114,51 @@ def ipccsd_diag(eom, imds=None):
     vector = eom.amplitudes_to_vector(Hr1, Hr2)
     return vector
 
+def _biorthonormalize_degenerate_left_evecs(eom, evals, evecs, levecs,
+                                            nmo, nocc, deg_tol=1e-6):
+    """Rotate left eigenvectors within each degenerate cluster so that the
+    left/right amplitude overlap matrix becomes the identity.
+
+    The left and right eigenvectors come from two independent Davidson
+    solves.  Within a degenerate cluster each solve returns an arbitrary
+    basis of the same subspace, so the index-wise overlap <l_i|r_i> used by
+    the star contraction is not well defined and can become arbitrarily
+    small, amplifying the residual error of the Davidson vectors.  Solving
+    L' = S^{-1} L with S_ij = <l_i|r_j> gives <l'_i|r_j> = delta_ij, which
+    makes the per-root contraction independent of the arbitrary bases (for
+    symmetry-induced degeneracies the star correction is proportional to
+    the identity within the cluster).  Non-degenerate roots are returned
+    unchanged.
+    """
+    evals = np.atleast_1d(np.asarray(evals))
+    evecs = np.atleast_2d(np.asarray(evecs))
+    levecs = np.atleast_2d(np.asarray(levecs)).copy()
+    nroots = len(evals)
+
+    def overlap(lvec, rvec):
+        l1, l2 = eom.vector_to_amplitudes(lvec, nmo, nocc)
+        r1, r2 = eom.vector_to_amplitudes(rvec, nmo, nocc)
+        return np.dot(l1.ravel(), r1.ravel()) + 0.5 * np.dot(l2.ravel(), r2.ravel())
+
+    i0 = 0
+    while i0 < nroots:
+        i1 = i0 + 1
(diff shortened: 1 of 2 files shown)


psiQAQ:
Note on the red gate CI ([run 31953724084](https://github.com/pyscf/pyscf/actions/runs/31953724084)): the only failure is `sgx/test/test_dm_screening.py::KnownValues::test_uhf_dm_screening` on linux 3.8, which failed with a `nan` - unrelated to this PR (the change touches only the degenerate-cluster star-contraction path in `cc/eom_gccsd.py`; every EOM test including the two new gauge regressions passed inside that same job).

To rule out any interaction I ran a focused comparison on `ubuntu-latest` / Python 3.8 / `omp4-blas4`, 200 fresh-process attempts per branch:

- base = current master without this PR: [run 31956879785](https://github.com/psiQAQ/pyscf/actions/runs/31956879785) - **200/200 pass**;
- head = master + this PR: [run 31956881150](https://github.com/psiQAQ/pyscf/actions/runs/31956881150) - **200/200 pass**.

Neither branch reproduces the `nan` in isolation, so it appears to be an environment-specific flake of the gate run itself (first observation of this test failing; it is not among the #3312 entries). A re-run of the failed job should come back green.

sunqm (approved):

Issue #3379, opened 2026-08-05
[BUG REPORT] Incorrect energy minimization when smearing used


esoteric-ephemera:
There appears to be a bug in the energy quantity which is minimized / converged when smearing is turned on. In the [`energy_tot`](https://github.com/pyscf/pyscf/blob/8496d690b256339ffb0d2a72ca95dd9eb419c891/pyscf/scf/smearing.py#L256) method of smearing classes, the free energy $E(\sigma) - \sigma S(\sigma)$ should be returned, but only set as `scf.e_free`. 

When smearing is used, the free energy is the actual quantity to be [minimized](https://journals.aps.org/prb/abstract/10.1103/PhysRevB.107.195122), and the forces which are minimized correspond to the free energy functional.

This could lead to false convergence if the smearing term in the free energy is unstable. More, users who expect to get the zero-smearing extrapolated value as `scf.e_tot` will be mistaken, as this quantity is stored as `scf.e_zero`.

I'm happy to submit a PR to modify the behavior of this, but opening an issue here to permit discussion

PR #3410, opened 2026-08-19
Use e_free for conv check for smear scf

jeanwsr:
This was originally for debugging #3331 and I found it finally partly address #3379.

The Fe2 smeared UHF in #3331 has a free-energy-degenerate solution family: E_tot varies around 7e-6 across runs, while e_free is pinned to 1e-8 -- the well-defined quantity.

So I made smearing SCF to check e_free for convergence, and now the test unconvergence rate is 0/1000.

Notes for changes in this PR:
- Because the previous extra-cycle conv-check is inlined, I turned it into a hook function for cleaner code. The new check_extra_convergence function is used like check_convergence.
- The second run changing `mu` in the test looks like no-op so I removed it. The smearing SCF has no `mu` but only `mu0`.

For #3331: the e_tot -243.086994 state still occurs, but in the new scheme it is considered as correct since its e_free does not violate test assert. The assertion of e_tot is still kept, but with a looser criteria, to let this condition pass.

For #3379: the request of changing e_tot definition is not considered here, because I think this is backward compatibility breaking and needs more discussion.

### DIFF
--- a/pyscf/scf/smearing.py
+++ b/pyscf/scf/smearing.py
@@ -133,6 +133,7 @@ class _SmearingSCF:
         self.entropy = None
         self.e_free = None
         self.e_zero = None
+        self._e_free_prev = None
 
     def undo_smearing(self):
         obj = lib.view(self, lib.drop_class(self.__class__, _SmearingSCF))
@@ -143,6 +144,25 @@ class _SmearingSCF:
         del obj.e_free
         del obj.e_zero
         return obj
+
+    def pre_kernel(self, envs):
+        self._e_free_prev = self.e_free
+
+    def check_extra_convergence(self, envs):
+        '''Convergence check of the extra (conv_check) cycle.
+
+        The envs dict carries the relaxed tolerances (conv_tol*10,
+        conv_tol_grad*3).  Smearing minimizes the free energy
+        e_free = E - sigma*S, so the extra cycle certifies convergence on
+        e_free instead of the plain energy (issue #3379).  When smearing is
+        inactive the default OR logic is replicated.
+        '''
+        if (self.sigma and self.smearing_method
+                and self.e_free is not None and self._e_free_prev is not None):
+            return (abs(self.e_free - self._e_free_prev) < envs['conv_tol']
+                    and envs['norm_gorb'] < envs['conv_tol_grad'])
+        return (abs(envs['e_tot'] - envs['last_hf_e']) < envs['conv_tol']
+                or envs['norm_gorb'] < envs['conv_tol_grad'])
 
     def get_occ(self, mo_energy=None, mo_coeff=None):
         '''Label the occupancies for each orbital
@@ -256,6 +276,7 @@ class _SmearingSCF:
     def energy_tot(self, dm=None, h1e=None, vhf=None):
         e_tot = self.energy_elec(dm, h1e, vhf)[0] + self.energy_nuc()
         if self.sigma and self.smearing_method and self.entropy is not None:
+            self._e_free_prev = self.e_free
             self.e_free = e_tot - self.sigma * self.entropy
             self.e_zero = e_tot - self.sigma * self.entropy * .5
             logger.info(self, '    Total E(T) = %.15g  Free energy = %.15g  E0 = %.15g',
--- a/pyscf/scf/smearing.py
+++ b/pyscf/scf/smearing.py
@@ -148,6 +148,22 @@ class _SmearingSCF:
     def pre_kernel(self, envs):
         self._e_free_prev = self.e_free
 
+    def check_convergence(self, envs):
+        '''Convergence check of the main SCF loop.
+
+        Smearing minimizes the free energy e_free = E - sigma*S, so the main
+        loop requires both the plain energy and the free energy to be
+        converged (issue #3379).
+        '''
+        if (self.sigma and self.smearing_method
+                and self.e_free is not None and self._e_free_prev is not None):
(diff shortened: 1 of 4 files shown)


psiQAQ:
Independent cross-validation with the harness used in #3331/#3312, on top of this PR's head (`ce71ff65`; the verification branch `psiQAQ:codex/investigate/pr3410-efree-verify` adds only CI tooling plus the Windows wheel build/loader infrastructure - no change to any scientific path):

- **Linux source-tree** (ubuntu-latest, Python 3.12, `omp4-blas4`): **1000/1000 pass** ([run 32223176764](https://github.com/psiQAQ/pyscf/actions/runs/32223176764));
- **Windows installed wheel** (windows-latest, Python 3.13, `omp4-blas4`): **200/200 pass** ([run 32224026926](https://github.com/psiQAQ/pyscf/actions/runs/32224026926)) - the other historically failing platform (#3312 samples), not covered in the PR description.

RED baselines on unmodified master for contrast, same harness and profile: 1/200 at `conv_tol=1e-7` ([run 31689528256](https://github.com/psiQAQ/pyscf/actions/runs/31689528256)) and 3/1000 at `conv_tol=1e-8` ([run 31772613521](https://github.com/psiQAQ/pyscf/actions/runs/31772613521)).

The e_free criterion resolves the flake on both platforms. Per-attempt logs, records.jsonl and full environment capture are in the run artifacts.

pyscf/scf/smearing.py
```diff
@@ -144,6 +145,41 @@ def undo_smearing(self):
...
+        if (self.sigma and self.smearing_method
+                and self.e_free is not None and self._e_free_prev is not None):
+            dE = abs(envs['e_tot'] - envs['last_hf_e'])
+            dF = abs(self.e_free - self._e_free_prev)
+            return (dE < envs['conv_tol'] and dF < envs['conv_tol']
+                    and envs['norm_gorb'] < envs['conv_tol_grad'])
+        return (abs(envs['e_tot'] - envs['last_hf_e']) < envs['conv_tol']
+                and envs['norm_gorb'] < envs['conv_tol_grad'])
+
+    def check_extra_convergence(self, envs):
```
sunqm, line 166: why not merging this with the check_convergence?

The convergence check just needs to test the e_free. e_tot is not variational in this case. The strict convergence threshold for e_tot is not necessary.

lijiachen417:
Can this change be applied to PBC calculations as well?

pyscf/scf/smearing.py
jeanwsr: The main conv check did not use e_free only because I found that mode cannot provide a stable enough e_tot (with 1e-4 fluctuation) to pass the test. 
Now I realized that tighter conv_tol (like 1e-9) solves the problem, so the check_convergence now uses e_free only in the updated code. But the extra cycle grad is a little not stable enough in this mode, which causes 5% rate unconvergence. So I have to change check_extra_convergence to dF OR |g| (matches normal SCF, but still different from smearing's check_convergence) .

jeanwsr:
> Can this change be applied to PBC calculations as well?

It is affecting PBC smearing because of pyscf's inheriting logic. Currently no regression found.

pyscf/scf/test/test_addons.py
```diff
@@ -475,7 +475,7 @@ def test_rohf_smearing1(self):
         mf = addons.smearing(mf, sigma=0.1)
         mf.kernel()
         self.assertAlmostEqual(mf.mo_occ.sum(), 15, 8)
-        self.assertAlmostEqual(mf.e_tot, -106.9310800402, 8)
+        self.assertAlmostEqual(mf.e_tot, -106.9310800142, 8)
```
sunqm: e_tot is not variational. Checking 8 decimal places could easily break in the future. We can check e_free with 8 digits and e_tot with 4 or 5 digits, like other tests below

pyscf/scf/smearing.py
sunqm: Thanks for the explanation. I still think introducing a separate `check_extra_convergence` API is confusing and unnecessary. By overloading check_convergence, the variational quantity (e_free when smearing is enabled) together with the corresponding orbital gradient |g|shpuld be enough.

I don't think `dF OR |g|` is the right criterion. Since |g| is the orbital gradient of the Lagrangian, it should converge to the required threshold regardless of whether smearing is enabled. In contrast, e_tot is not variational, so I don't think it needs a strick check for its convergence in this case.

pyscf/scf/test/test_addons.py
jeanwsr: Updated.

pyscf/scf/smearing.py
jeanwsr: Hmm, by `dF OR |g|` I mean de_free or |g|, so e_tot is not checked here, matching your request. So the only remaining controversy is the AND/OR I think. My preference of OR mainly comes from the normal SCF's extra cycle, which does use OR. For extra cycle (with or without smearing), `10*de_tol AND 3*g_tol` is a bit hard to fulfill for our current program. Using OR is a lazy strategy to reduce extra-cycle-fail. Alternative plan may be looser tol with AND, like `30*de_tol AND 9*g_tol`.

Regarding the separate `check_extra_convergence` API, it is introduced because the current behavior of normal SCF is indeed different convergence checker for main loop and extra cycle. It does not change anything for normal SCF, and smearing SCF uses quite similar checker.

PR #3412, opened 2026-08-19
Bug fix and update `pbc.tools.precompute_exx`

hongzhouye:
This PR fixes several bugs in `precompute_exx` for periodic HF calculations with `exxdiv="vcut_ws"` and updates the implementation to more closely follow Sundararaman and Arias, Phys. Rev. B 87, 165122 (2013). In particular:

- The old implementation constructs the BvK lattice using `kcell.a = cell.lattice_vectors() * Nk`, which scales the *columns* of the lattice-vector matrix, whereas the k-mesh should scale its *rows*. The new implementation fixes this. This fix does not affect isotropic k-meshes but now correctly handles anisotropic k-meshes.
- The old minimum-image search considers only the eight translations in `[0, 1]^3`, which is insufficient for some anisotropic or skewed BvK supercells. The new implementation searches `[-nimgs, nimgs]` in each direction, with a default `nimgs=[3, 3, 3]`, and checks whether the supplied range is sufficient. Expert users can change the default through `__config__` as documented in the docstring.
- `Rin` is now the exact inradius of the BvK Wigner-Seitz cell, i.e. half the length of the shortest nonzero BvK lattice vector.
- `alpha` is determined from an accuracy threshold `precision`, which defaults to `min(cell.precision, 1e-11)`, following the criterion in the paper but also allows refinement with tightened `cell.precision`.
- The FFT mesh is determined from the Nyquist-frequency criterion in the paper. A separate `precision_fft` can be set through either the function argument or `__config__` to tune the FFT mesh without changing `alpha`.

Old tests in `pbc/tools/test/test_pbc.py` related to "vcut_ws" are updated. One new test regarding the `Rin` construction is added.

### A scientific note

The new FFT mesh is substantially smaller than the old default. For very small k-meshes, this introduces a small difference from the old, much denser numerical construction because of the quadrature error associated with the Wigner-Seitz boundary. The difference decreases rapidly as the k-mesh, and therefore the BvK supercell, increases. This was also noted by Sundararaman and Arias in their original implementation (see the end of Appendix 1):

> The cusps in the periodic repetition of the integrand at the boundaries of the Wigner-Seitz cell cause an additional error in the kernel, but this error does not contribute in the Coulomb energy of charge distributions that are confined to the half-sized Wigner-Seitz cell and are resolvable on the chosen Fourier grid.

The following table compares the HF exchange energies of diamond (2-atom primitive cell) in the GTH-cc-pVDZ basis set with the GTH-HF-rev pseudopotential. `cell.precision` is set to `1e-12`. The old FFT mesh is `[120,120,120]` while the new one is `[35,35,35]`. One can see that the difference between the new and old implementations becomes physically negligible for anything but very small k-meshes.

| k-mesh | New exchange energy (Ha) | Old exchange energy (Ha) | Error (microhartree) |
|:------:|--------------------:|--------------------:|---------------------:|
| 1x1x1  | -3.4252007880 | -3.4252832174 | 82.4 |
| 2x2x2  | -3.2658658746 | -3.2658693694 | 3.5 |
| 3x3x3  | -3.2386490504 | -3.2386495625 | 0.5 |
| 4x4x4 | -3.2307226222 | -3.2307225656 | 0.06 |

### Update (08/21)

The original implementation assumes that the input k-points form a (possibly twisted) uniform mesh. In `get_bands`, however, `q = kpti - band_kptj` generally does not lie on the reciprocal BvK lattice. The original code rounds `q` to the nearest reciprocal BvK grid point, resulting in unphysical band structures.

There are two possible solutions:

1. Calculate `vq` independently for every `kpti - band_kptj` pair, requiring up to `Nk * Nk_band` additional FFTs but no extra memory.
2. For each unique band-k-point shift, calculate `vq` once and cache it. This requires up to `Nk_band` additional FFTs and cached `vq` arrays.

For a fixed band k-point, all sampled k-points have the same fractional shift on the reciprocal BvK grid. This PR therefore uses the second approach. The additional memory is modest with the substantially reduced FFT mesh size. For example, a diamond calculation with a 6x6x6 SCF k-mesh and 100 band k-points requires about 40 MB for the cache (with default `1e-11` precision). The resulting `vcut_ws` band structure is now physical and agrees with `vcut_sph`, as illustrated below for silicon. An example for HF band-structure calculation using `vcut_ws` or `vcut_sph` is added in `examples/pbc/50-band_vcut.py`.

A related aliasing bug was also found in the reciprocal-grid bounds check. For a nonorthogonal lattice, a reciprocal vector can lie inside the Cartesian bounding box of the precomputed grid while its integer BvK coordinates lie outside the FFT index range. The subsequent modulo indexing,
```python
gxyz = (gxyz + mesh) % mesh
```
can then alias a high-q vector accidentally onto an unrelated low-q vector, causing sizable energy errors for large k-meshes. The new implementation checks the integer BvK coordinates directly against the FFT bounds before applying the modulo indexing.

Simple tests have been added to `test_pbc.py` for both bugs.

band

### DIFF
--- a/pyscf/pbc/tools/pbc.py
+++ b/pyscf/pbc/tools/pbc.py
@@ -486,65 +486,186 @@ def get_coulG(cell, k=np.zeros(3), exx=False, mf=None, mesh=None, Gv=None,
             coulG[G0_idx] += Nk*cell.vol*madelung(cell, kpts, omega=0)
     return coulG
 
-def precompute_exx(cell, kpts=None):
+
+def precompute_exx(cell, kpts=None, precision=None, precision_fft=None, nimgs=None):
+    '''Precompute the Wigner-Seitz truncated EXX kernel.
+
+    The long-range part of the kernel is constructed with the minimum-image
+    convention and range separation of Eq. (A4) in Phys. Rev. B 87, 165122
+    (2013). The short-range part is evaluated analytically in :func:`get_coulG`.
+
+    Args:
+        cell : :class:`pyscf.pbc.gto.Cell`
+            Primitive cell.
+        kpts : (nkpts, 3) array_like
+            Complete regular k-point mesh. Defaults to the Gamma point.
+        precision : float
+            Accuracy threshold used to set the range-separation parameter ``alpha``.
+            Defaults to ``min(cell.precision, 1e-11)``, where the default value
+            1e-11 follows the PRB paper above.
+        precision_fft : float
+            Accuracy threshold used to set the FFT mesh for the numerical
+            long-range kernel. Defaults to ``__config__.pbc_tools_pbc_vcut_ws_precision_fft``
+            if set, and to ``precision`` otherwise. Smaller values produce denser FFT
+            meshes without changing ``alpha``.
+        nimgs : (3,) array_like of int
+            Number of lattice images searched in each direction on both sides
+            of the Born-von Karman cell. Defaults to [3,3,3], which can be overwritten
+            by setting the `__config__` attribute "pbc_tools_pbc_vcut_ws_nimgs".
+
+    Returns:
+        dict
+            Range-separation parameter, Born-von Karman cell, reciprocal
+            vectors, and the numerical long-range kernel.
+    '''
     from pyscf.pbc import gto as pbcgto
-    from pyscf.pbc.dft import gen_grid
+    from pyscf.pbc.lo.base import get_kmesh
+
     log = lib.logger.Logger(cell.stdout, cell.verbose)
-    log.debug("# Precomputing Wigner-Seitz EXX kernel")
-    Nk = get_monkhorst_pack_size(cell, kpts)
-    log.debug("# Nk = %s", Nk)
+    log.debug('# Precomputing Wigner-Seitz EXX kernel')
+
+    if kpts is None: kpts = np.zeros((1, 3))
+    kpts = np.reshape(kpts, (-1, 3))
+    kmesh = np.asarray(get_kmesh(cell, kpts), dtype=int)
+    scaled_kpts = cell.get_scaled_kpts(kpts - kpts[0])
+    scaled_kpts = np.rint(scaled_kpts * kmesh).astype(int) % kmesh
+    if len(np.unique(scaled_kpts, axis=0)) != len(kpts):
+        raise RuntimeError('Input k-points do not form a complete regular mesh')
+    log.debug('# kmesh = %s', kmesh)
+
+    if precision is None:
+        precision = min(cell.precision, 1e-11)
+    else:
+        precision = float(precision)
+    assert 0 < precision < 1
+
+    if precision_fft is None:
+        precision_fft = getattr(__config__, 'pbc_tools_pbc_vcut_ws_precision_fft', None)
+        if precision_fft is None:
+            precision_fft = precision
+    precision_fft = float(precision_fft)
+    assert 0 < precision_fft < 1
+
+    if nimgs is None:
+        nimgs = getattr(__config__, 'pbc_tools_pbc_vcut_ws_nimgs', [3, 3, 3])
+    nimgs = np.asarray(nimgs, dtype=int)
+    assert nimgs.shape == (3,)
+    assert np.all(nimgs > 0)
 
     kcell = pbcgto.Cell()
     kcell.atom = 'H 0. 0. 0.'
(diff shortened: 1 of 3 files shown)


hongzhouye:
An additional note: because the old implementation scales the columns rather than the rows of the lattice-vector matrix, calculations using anisotropic k-meshes (i.e., `[n1, n2, n3]` with unequal values) are generally incorrect.

For example, consider the same diamond system with a `[1,1,2]` k-mesh. The old implementation gives an HF exchange energy of `-3.4778660490 Ha`, corresponding to a finite-size error of about `7 eV` relative to the nearly TDL value of `-3.2273368014 Ha` from a `[6,6,6]` calculation. The new implementation gives `-3.3364212966 Ha`, reducing the finite-size error to about `3 eV`. This result is qualitatively consistent with Fig. 2(c) of the Sundararaman–Arias paper, which shows a finite-size error of approximately `3 eV` for the `[1,1,2]` mesh of diamond.

pyscf/pbc/tools/pbc.py
```diff
@@ -486,67 +503,207 @@ def get_coulG(cell, k=np.zeros(3), exx=False, mf=None, mesh=None, Gv=None,
...
+        precision_fft = getattr(__config__, 'pbc_tools_pbc_vcut_ws_precision_fft', None)
+        if precision_fft is None:
+            precision_fft = precision
+    precision_fft = float(precision_fft)
+    assert 0 < precision_fft < 1
+
+    log.debug('# precision_fft = %.15g', precision_fft)
+
+    if nimgs is None:
+        nimgs = getattr(__config__, 'pbc_tools_pbc_vcut_ws_nimgs', [3, 3, 3])
```
sunqm, line 568: cell.nimgs indicates the boundary of lattice sum. does it make sense to use cell.nimgs as the initial guess of ws nimgs?

pyscf/pbc/tools/pbc.py
```diff
@@ -486,67 +503,207 @@ def get_coulG(cell, k=np.zeros(3), exx=False, mf=None, mesh=None, Gv=None,
...
+        precision = float(precision)
+    assert 0 < precision < 1
+
+    log.debug('# precision = %.15g', precision)
+
+    if precision_fft is None:
+        precision_fft = getattr(__config__, 'pbc_tools_pbc_vcut_ws_precision_fft', None)
+        if precision_fft is None:
+            precision_fft = precision
+    precision_fft = float(precision_fft)
```
sunqm: can the same value to be used for `precision` and `precision_fft`? It looks more reasonable to use the consistent precision settings for the long- and short-range

pyscf/pbc/tools/pbc.py
```diff
@@ -486,67 +503,207 @@ def get_coulG(cell, k=np.zeros(3), exx=False, mf=None, mesh=None, Gv=None,
...
-    r = np.min([lib.norm(rs-c, axis=1) for c in corners], axis=0)
+    log.debug('# kcell kngs = %d', kngs)
+
+    images_coord = lib.cartesian_prod([
+        range(-n, n + 1) for n in nimgs
+    ])
+    images = np.dot(images_coord, kcell.a)
+    r = np.full(kngs, np.inf)
+    r_mic = np.empty_like(rs)
+    for image in images:
```
sunqm, line 602: why not directly using the `nimgs_ref` here to determine `r` and `r_mic` ?

pyscf/pbc/tools/pbc.py
hongzhouye: I originally kept them separately for benchmarking the convergence with the FFT mesh size. But you are right it looks confusing. I've removed `precision_fft` so now there's only a single `precision` parameter linked to `cell.precision`.

pyscf/pbc/tools/pbc.py
hongzhouye: Agreed. I've updated the logic so now `nimgs` only provides an initial guess to determine `nimgs_ref`. `r` and `r_mic` are now determined from `nimgs_ref`.

pyscf/pbc/tools/pbc.py
hongzhouye: I would keep them separately because `cell.nimgs` are determined from the decay of AO basis, while the `nimgs` needed here are purely geometric and only depend on the BvK cell shape.

sunqm (approved):

PR #3415, opened 2026-08-20
Add density matrix keyword to response functions

jonas-greiner:
This PR adds a `dm0` keyword to the `_gen_response` functions (for RHF/UHF/GHF/DHF/X2C-HF) so the response function can be built directly from a density matrix instead of requiring `mo_coeff` and `mo_occ`. This is useful for callers that only have a density matrix on hand and would otherwise have to diagonalize it back into an MO basis just to call this function. Internally this reuses the existing `cache_xc_kernel1` helper (already in `numint.py`) instead of `cache_xc_kernel` when `dm0` is passed. For the NLC contribution, `get_vnlc_resp` in `pyscf/hessian/rks.py `is split into `get_vnlc_resp` (which builds `dm0` from `mo_coeff/mo_occ`) and a new `get_vnlc_resp1` function (which takes dm0 directly), so the `dm0` path can call the latter. No new numerical machinery is introduced. I also added tests for the equivalence of the implementations employing `mo_coeff` vs. `dm0` for RHF/UHF/GHF/DHF/X2C-HF.

### DIFF
--- a/pyscf/hessian/rks.py
+++ b/pyscf/hessian/rks.py
@@ -1838,17 +1838,6 @@ def _get_vnlc_deriv1(hessobj, mo_coeff, mo_occ, max_memory):
     return vmat
 
 def get_vnlc_resp(mf, mol, mo_coeff, mo_occ, dm1s, max_memory):
-    """
-        Equation notation follows:
-        Liang J, Feng X, Liu X, Head-Gordon M. Analytical harmonic vibrational frequencies with
-        VV10-containing density functionals: Theory, efficient implementation, and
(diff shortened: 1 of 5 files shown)


jeanwsr:
But in what condition the caller "only have a density matrix"?

jonas-greiner:
In my case I am optimizing the density matrix directly in an external code, and I need the response function from that density matrix alone without any MO coefficients ever computed. Constructing `mo_coeff`/`mo_occ` just to call `gen_response` that the function immediately turns back into a density matrix seems like a redundant round trip forced by the current interface. That said, I get that this may be a niche use case, and I'm fine if the community doesn't see enough value in supporting it.

jeanwsr (approved):

PR #3416, opened 2026-08-23
Support specifying the solvent name in PCM

### LINKED ISSUE #3413: [FEATURE REQUEST] Support specifying solvent name in PCM
**Describe the feature you'd like to see implemented**
PCM should support `solvent=xxx` like SMD. This makes the user-input more convenient and avoids incomplete logic of inferring eps_optical in #3386.

JustinKyleKirkland:
Fixes #3413.

PCM only accepted a dielectric constant, and it had no way to know the refractive index of the solvent, so the optical dielectric constant applied by the non-equilibrium solvation of excited states fell back to the value of water for every solvent (see the discussion in #3386).

### Changes

* The solvent database is moved from `smd.py` into a new `pyscf.solvent._solvent_data` module, so that the PCM family models can use it as well. `smd.solvent_db` still refers to the same dictionary.
* `PCM(mol, 'toluene')`, `mf.PCM('toluene')`, `solvent.PCM(mol, 'toluene')` and `pcmobj.solvent = 'toluene'` assign `eps` and `eps_optical` from that database. Both attributes can still be overwritten afterwards, and changing the solvent after `build()` discards the outdated intermediates.
* Solvent names are matched case insensitively, ignoring spaces and hyphens (`hexane` and `n-hexane` both work), and common abbreviations such as `DMSO`, `THF` or `MeCN` are recognized.
* `SMD` inherits the same name matching (`mf.SMD('DMSO')`), but keeps its own behavior of reading every descriptor from the database at build time, so `eps` stays `None` there and a custom `sol_desc` is not overridden.
* `solvent` is reported by `dump_flags`.

```python
mf = mol.RHF().PCM('toluene')
mf.with_solvent.eps          # 2.3741
mf.with_solvent.eps_optical  # 1.4961**2, no longer the value of water
```

### Backward compatibility

Results are unchanged unless a solvent name is specified. The one behavior change is the exception raised for an unknown solvent name in the `SMD` constructor: it now comes from the shared lookup, with a message pointing at the database. It is still a `RuntimeError`.

Note that `PCM(mol, 'water')` uses `eps=78.355` from the database, rather than the `EPS_WATER=78.3553` default of the module. I kept the database as the single source of truth instead of special casing water, but I am happy to change that if you prefer the two to agree exactly.

`ddCOSMO` and `ddPCM` are untouched. The same mechanism would apply to them, so let me know if it should be extended in this PR.

### Tests

Added `NamedSolvent` in `test_pcm.py` and `test_smd.py`, and `test_eps_optical_from_solvent_name` in `test_pcm_tdscf.py`, covering the name matching, the assignment of `eps`/`eps_optical`, the energy equivalence with a hand-set `eps`, the rebuild after a solvent change, and the error for unknown names. The full `pyscf/solvent/test` suite passes (112 passed, 53 skipped, the skips being the tests that need the SMD CDS library).


jeanwsr:
I think we can use 78.3553 in the db, since it's more accurate and used in other programs. The db comment can be "adapted from database ...".

JustinKyleKirkland:
> I think we can use 78.3553 in the db, since it's more accurate and used in other programs. The db comment can be "adapted from database ...".

Done!

sunqm (approved):

JustinKyleKirkland:
CI failure has nothing to do with my code change.

jeanwsr (approved):

jeanwsr:
Will be great if you update the [doc](https://github.com/pyscf/pyscf.github.io) as well.

JustinKyleKirkland:
> Will be great if you update the [doc](https://github.com/pyscf/pyscf.github.io) as well.

Done @jeanwsr

Issue #3414, opened 2026-08-20
[UNEXPECTED BEHAVIOR] Error of state-specific excited state DMRG-PDFT calculation


1234zou:
**Describe the unexpected behavior**
I'm trying to calculating the S1 state using CASCI-based DMRG-PDFT via `mc.state_specific_(1)`. However, the following error in found
```python
CASCI E = -688.645328292437  E(CI) = -31.7177562469612  S^2 = 0.0000000
Traceback (most recent call last):
  File "/public/home/jxzou/vip_test/tet_dmrgci1818_mcpdft_r1/test.py", line 58, in <module>
    mc.kernel()
  File "/public/home/jxzou/software/pyscf-2.14.0/pyscf/mcpdft/mcpdft.py", line 543, in kernel
    self.compute_pdft_energy_(otxc=otxc, grids_attr=grids_attr,
  File "/public/home/jxzou/software/pyscf-2.14.0/pyscf/mcpdft/mcpdft.py", line 510, in compute_pdft_energy_
    epdft = [self.energy_tot(mo_coeff=self.mo_coeff, ci=self.ci, state=ix,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/public/home/jxzou/software/pyscf-2.14.0/pyscf/mcpdft/mcpdft.py", line 510, in <listcomp>
    epdft = [self.energy_tot(mo_coeff=self.mo_coeff, ci=self.ci, state=ix,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/public/home/jxzou/software/pyscf-2.14.0/pyscf/mcpdft/mcpdft.py", line 804, in energy_tot
    e_tot, e_ot = energy_tot(self, mo_coeff=mo_coeff, ot=ot, ci=ci,
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/public/home/jxzou/software/pyscf-2.14.0/pyscf/mcpdft/mcpdft.py", line 66, in energy_tot
    casdm1s = mc.make_one_casdm1s(ci, state=state)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #3419, opened 2026-08-23
Fix state-specific DMRG-PDFT calculations (#3414)

anandvenugopal-tech:
Fixes an issue where `mc.state_specific_(n)` fails in DMRG-PDFT calculations with a `TypeError`. 

The `StateSpecificFCISolver` sets `nroots = state + 1` internally, but its `kernel()` only returns a single CI vector. The PDFT code previously used `nroots` to determine the number of CI vectors, which caused it to incorrectly index into a single CI vector object. 

This PR fixes the issue by replacing `nroots`-based checks with `isinstance(ci, (list, tuple))` to correctly determine whether multiple CI vectors are present. This fix was applied to `_get_fcisolver` in `_dms.py`, and `compute_pdft_energy_` in `mcpdft.py`.


sunqm (approved):

jeanwsr:
This is clearly wrong. Your `compute_pdft_energy_` is not even computing the correct state.

Please verify the result before PR.

PR #3420, opened 2026-08-24
Analytic nuclear gradients for GHF and GKS

JustinKyleKirkland:
## Motivation

`GHF.nuc_grad_method` and `GKS.nuc_grad_method` are both a bare `NotImplementedError` (`pyscf/scf/ghf.py:550`, `pyscf/dft/gks.py:179`), so neither generalized reference can be geometry-optimized. GHF is the reference for non-collinear spin, for spin-orbit coupling via `x2c1e()`, and for symmetry-broken solutions, and four-component DHF has had analytic gradients all along (`pyscf/grad/dhf.py`).

## Approach

A GHF density matrix is Hermitian over the spin-blocked AO basis and is complex once the solution is not collinear. Two observations keep the whole gradient on the existing real integral drivers rather than requiring a complex one.

The Coulomb term only sees the charge density `dm_aa + dm_bb`, whose antisymmetric imaginary part cancels against the symmetric two-electron integrals. The same cancellation removes the imaginary part from the (symmetric) hcore derivative and from the bra-plus-ket sum of the overlap derivative, so those three terms use real matrices.

Hermiticity of the full density matrix, `dm_ts[nu,mu] = dm_st[mu,nu].conj()`, splits the exchange energy into real and imaginary parts,

```
E_K = -1/2 sum_st sum_mn ( K[Re dm_st]_mn Re dm_st_mn + K[Im dm_st]_mn Im dm_st_mn )
```

so the four spin blocks enter as eight real density matrices through the ordinary `get_k` driver.

`grad/gks.py` reuses `grad/uks.py` for the exchange-correlation part. A collinear functional, which is the default, sees only the diagonal spin blocks exactly as UKS does. Hybrids and range-separated hybrids are handled.

## Scope

Implemented: GHF gradients; GKS gradients for `collinear='col'` including hybrids and range separation, with grid response.

Explicitly refused rather than silently approximated: the `ncol` and `mcol` schemes raise `NotImplementedError`, since they need spin-density XC derivatives on the grid that `numint2c` does not currently provide (it has `_gks_mcol_vxc` for the potential but nothing gradient-side). NLC functionals raise for the same reason. There is a test asserting these refuse.

## Validation

| check | agreement |
| --- | --- |
| closed-shell GHF vs RHF gradient | 1e-8 |
| collinear GHF vs UHF gradient | 3e-10 |
| finite difference, complex non-collinear H3 | 3.5e-9 |
| finite difference, heteronuclear with d functions | 1.3e-8 |
| global spin-rotation invariance (exact) | 3e-15 |
| GKS finite difference: LDA, PBE, B3LYP, TPSS | 1e-8 or better |
| closed-shell GKS vs RKS gradient | 1e-14 |
| translational invariance, `atmlst` subsetting | exact |

The spin-rotation check is the strongest of these. Rotating the spin quantization axis maps a solution onto an exactly degenerate one with complex off-diagonal spin blocks, so the gradient has to be identical, with no finite-differencing error to hide in. It holds to machine precision, and the test asserts the rotated density really does have `|Im dm|` and `|dm_ab|` well above zero so it cannot pass vacuously.

For GKS, `grid_response=True` improves finite-difference agreement by roughly two orders of magnitude over `False`, which is the expected signature of a correct grid response term.

`pyscf/grad/test/test_ghf.py` adds 8 tests, running in about 13 seconds. Reference-free where possible: the checks are invariances and finite differences rather than hard-coded numbers.

## Notes

* `grad/__init__.py` now exports `GHF` and `GKS` alongside the existing references.
* Lint is clean under all four checks the Lint workflow runs (ruff with the repo config at the pinned `ruff<0.16.0`, ruff NPY, flake8, pycodestyle).
* Not included, and worth separate work: GKS Hessians, and excited-state gradients for TD-GHF/GKS/DHF/DKS, which remain energy-only.

### DIFF
--- a/pyscf/grad/ghf.py
+++ b/pyscf/grad/ghf.py
@@ -1,0 +1,179 @@
+#!/usr/bin/env python
+# Copyright 2014-2026 The PySCF Developers. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+'''
+Non-relativistic generalized Hartree-Fock analytical nuclear gradients
+'''
+
+import numpy
+from pyscf import lib
+from pyscf.lib import logger
+from pyscf.grad import rhf as rhf_grad
+
+
+def _spin_block_dms(dm):
+    '''Real density matrices that the derivative J and K integrals are
+    contracted with.
+
+    A GHF density matrix is Hermitian over the spin-blocked AO basis and is
+    complex whenever the solution is not collinear,
+
+        dm = [[dm_aa, dm_ab],
+              [dm_ba, dm_bb]]
+
+    The Coulomb term only sees the charge density dm_aa + dm_bb. Its
+    imaginary part is antisymmetric and cancels against the symmetric
+    two-electron integrals, so only the real part is kept.
+
+    The exchange term sees each spin block on its own. Using the hermiticity
+    of the full density matrix, dm_ts[nu,mu] = dm_st[mu,nu].conj(), the real
+    exchange energy separates into real and imaginary parts,
+
+        E_K = -1/2 sum_st sum_mn ( K[Re dm_st]_mn Re dm_st_mn
+                                 + K[Im dm_st]_mn Im dm_st_mn )
+
+    which lets the whole gradient run through the existing real integral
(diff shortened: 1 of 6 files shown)


pyscf/grad/ghf.py
```diff
@@ -0,0 +1,179 @@
...
+        mf_grad : grad.ghf.Gradients or grad.gks.Gradients object
+    '''
+    mf = mf_grad.base
+    mol = mf_grad.mol
+    if mo_energy is None: mo_energy = mf.mo_energy
+    if mo_occ is None:    mo_occ = mf.mo_occ
+    if mo_coeff is None:  mo_coeff = mf.mo_coeff
+    log = logger.Logger(mf_grad.stdout, mf_grad.verbose)
+
+    hcore_deriv = mf_grad.hcore_generator(mol)
```
sunqm, line 103: raise an error if `mf.with_soc` is enabled.

JustinKyleKirkland: Done! @sunqm

sunqm (approved):

Issue #2925, opened 2025-08-05
[FEATURE REQUEST] Sigma potentials in solvation module?


DazlerNL:
Hello,

Thank you for your great software, it is very convenient to use.

I wanted to ask if it's possible to extract sigma-profiles and sigma-potentials from the solvation functionality in its current state, or if not, if that could be added in the future.

So far, I have been able to extract sigma-profiles by using other software on COSMO files generated in PySCF, but I've not been able to generate sigma-potentials. Sadly, my python skills are not sufficient to solve the necessary equations as described by Klamt. See for example:
- https://www.researchgate.net/publication/221876222_COSMO-RS_An_Alternative_to_Simulation_for_Calculating_Thermodynamic_Properties_of_Liquid_Mixtures 
Or for a simplified single descriptor version:
- https://chemrxiv.org/engage/api-gateway/chemrxiv/assets/orp/resource/item/60c73f19469df44eb0f429e0/original/[cosmoplex-self-consistent-simulation-of-self-organizing-inhomogeneous-systems-based-on-cosmo-rs.pdf]

Kind regards,

John van de Moosdijk

sunqm:
I'm not particularly sure how the sigma profile is calculated. If you have a formula for the sigma profile, I guess it can be computed from "segments" section of the output of the cosmo-rs 
```
from pyscf.solvent.cosmors import get_pcm_parameters
ps = get_pcm_parameters(mf, step)
print(ps['segments'])
```
The elements of `segments` are defined here.
https://github.com/pyscf/pyscf/blob/fafef00a03b28a9ddbe23cad2d99ce0f1222180c/pyscf/solvent/cosmors.py#L249

PR #3421, opened 2026-08-24
Implement the outlying charge correction in the COSMO-RS module

JustinKyleKirkland:
## The problem

Every corrected field in `cosmors.py` is currently a copy of the uncorrected value:

```
cosmors.py:238  'e_tot_corr':  float(E_tot),               # TODO: implement OCC
cosmors.py:240  'e_diel_corr': float(E_diel)               # TODO: implement OCC
cosmors.py:257  'charge_corr': charges.tolist(),           # TODO: implement OCC
cosmors.py:260  'sigma_corr':  (charges / areas).tolist(), # TODO: implement OCC
```

plus `screening_charge['correction']` hardcoded to `0.0`.

`write_cosmo_file` writes those into the `$screening_charge` and `$cosmo_energy` blocks and uses `charge_corr` for the per-segment charges. So a COSMO file produced by PySCF tells COSMOtherm, ORCA's reader, and anything else consuming the format that uncorrected charges have been corrected. The docstrings did note the correction was missing, but a file parser does not read docstrings. This is closer to a correctness bug than a missing feature, which is why I picked it up.

## The correction

Part of the solute electron density lies outside the cavity, so the screening charges fall short of the total surface charge Gauss's law requires. `get_outlying_charge_correction` restores the two sum rules

```
sum(q_nuc)  = -f_eps * Q_nuc
sum(q_elec) = -f_eps * Q_elec
```

by scaling the nuclear and electronic screening charges separately, then regenerating the surface potential from the corrected charges. Correcting the potential and not only the charges is what makes the corrected dielectric energy accurate; see Klamt and Jonas, J. Chem. Phys. 105, 9972 (1996), and the recent discussion in the NWChem COSMO paper (J. Chem. Theory Comput., 2026).

The COSMO equations are linear in the surface potential, so the split into nuclear and electronic charges is exact and comes straight from `v_grids` and `v_grids_n`.

I did not implement the Klamt-Jonas double-cavity variant. It projects only the contact spherical portions of the surface, leaves the reentrant portion uncorrected, and has been criticised for producing arbitrary changes in the multipole moments of the surface charge distribution. The scaling form is better defined and is what NWChem moved to.

## Scope

Restricted to C-PCM and COSMO. Their response matrix is a multiple of the identity, `R = -f_eps * I`, so the screening charges are a uniform `f_eps` scaling of the conductor charges and the sum rules above are the model's own. IEF-PCM and SS(V)PE raise `NotImplementedError` rather than report a wrong correction. That is not much of a restriction in practice: `write_cosmo_file` already requires `f_eps = 1`, and in that conductor limit IEF-PCM reduces to C-PCM exactly (verified numerically, the two agree to machine precision).

## It is not cosmetic

| system | sum(q) before | sum(q) after | exact | dE_diel |
| --- | --- | --- | --- | --- |
| H2O / 6-31G* | -0.002846 | 0.000000 | 0 | -0.030 kcal/mol |
| H2O / aug-cc-pVTZ | -0.009265 | 0.000000 | 0 | -0.042 kcal/mol |
| NH4+ / 6-31G* | -1.002679 | -1.000000 | -1 | +0.444 kcal/mol |
| OH- / aug-cc-pVDZ | +0.963934 | +1.000000 | +1 | **-6.425 kcal/mol** |

The anion in a diffuse basis is the case that matters: its density leaks badly, the uncorrected screening charge is off by 3.6%, and the dielectric energy moves by 6.4 kcal/mol. The correction also scales with basis diffuseness, as it should.

## Validation

* the nuclear/electronic charge split reproduces the full solution to 1e-15;
* the corrected total screening charge equals `-f_eps * Q` to 1e-15 for neutral, anionic and cationic solutes;
* the regenerated potential reproduces PySCF's own `e_solvent` **exactly** (difference 0.0) when the scaling factors are set to one, which confirms the energy expression is PySCF's own relation rather than a reimplementation;
* the scaling factors come out at 1.0013 (nuclear) and 1.0016 (electronic) for H2O/6-31G*. The nuclear factor is the pure cavity-discretisation deficit, since point nuclei are strictly inside the cavity; the extra amount on the electronic factor is the genuine outlying charge. Scaling them separately is what separates the two.

Four tests added to `test_cosmors.py`, which now runs 8 tests in under 4 seconds. The four existing tests assert only uncorrected quantities and are unaffected.

## Notes

* `cosmors.py` has CRLF line endings; the diff preserves them, so it stays at 95 insertions rather than a whole-file rewrite.
* The stale "not implemented yet" note in `examples/solvent/07-cosmors.py` is updated.
* This unblocks #2925 (sigma-profiles and sigma-potentials), where the profile is a histogram over `segments['sigma']` and now has corrected charges to work from.
* Lint is clean under all four checks the Lint workflow runs.

### DIFF
--- a/examples/solvent/07-cosmors.py
+++ b/examples/solvent/07-cosmors.py
@@ -31,12 +31,15 @@ mol = gto.M(atom=coords, basis='6-31G*', verbose=1)
 cm = pcm.PCM(mol)
 cm.eps = float('inf') # f_epsilon = 1 is required for COSMO-RS
 cm.method = 'C-PCM' # or COSMO, IEF-PCM, SS(V)PE, see https://manual.q-chem.com/5.4/topic_pcm-em.html
+                    # outlying charge correction requires C-PCM or COSMO
 cm.lebedev_order = 29 # lebedev grids on the cavity surface, lebedev_order=29  <--> # of grids = 302
 
 
 #%% COSMO-files
 
-# Please note, that outlying charge correction is not implemented yet
+# The corrected charges and energies in the COSMO-file carry the outlying
+# charge correction. It is available for C-PCM and COSMO; the other models
+# raise NotImplementedError.
 
 # run DFT SCF (any level of theory is OK, though DFT is optimal)
 mf = dft.RKS(mol, xc='b3lyp')
--- a/pyscf/solvent/cosmors.py
+++ b/pyscf/solvent/cosmors.py
@@ -178,10 +178,88 @@ def _lot(mf):
 
 
 
+#%% Outlying charge correction
+
+def get_outlying_charge_correction(mf):
+    '''Outlying charge correction for the COSMO screening charges.
+
+    Part of the solute electron density lies outside the cavity, so the
+    computed screening charges fall short of the total surface charge that
+    Gauss's law requires,
+
+        sum(q_nuc)  = -f_eps * Q_nuc
+        sum(q_elec) = -f_eps * Q_elec
+
+    where Q_nuc and Q_elec are the total nuclear and electronic charges of the
+    solute. The correction restores both sum rules by scaling the nuclear and
+    electronic screening charges separately, and then regenerates the surface
+    potential from the corrected charges. Correcting the potential as well as
+    the charges is what makes the corrected dielectric energy accurate.
+
+    Only C-PCM and COSMO are supported. Their response matrix is a multiple of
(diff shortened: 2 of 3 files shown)


JustinKyleKirkland:
The CI error has nothing to do with my code-change on this one. Should be good to merge, pending review. @jeanwsr

PR #3426, opened 2026-08-27
Analytic nuclear gradients for DF-CCSD and DF-CCSD(T)

### LINKED ISSUE #3425: [BUG REPORT] CCSD/CCSD(T) analytic nuclear gradients incorrect if density fitting is used
**Version:** pyscf 2.14.0 · numpy 2.4.5 · Linux x86-64 

## Summary

`pyscf.grad.ccsd.grad_elec` always contracts the relaxed
two-particle density matrix against **exact four-index ERI derivatives**, with no branch on whether
the underlying mean field is density-fitted. `pyscf.grad.ccsd_t.grad_elec` reuses the same assembly
and contains no DF handling either, so we expect it to be affected identically -- though the
reproducer below only demonstrates CCSD. When the reference is built with
`scf.RHF(mol).density_fit()`, the returned analytic gradient therefore does not differentiate the
energy that was actually computed.

There is no warning and no `NotImplementedError` — the calculation completes and returns a
plausible-looking gradient. That silence is the main problem: the result is wrong by an amount that
is small enough to look reasonable and large enough to matter.

## Reproducer

[df_ccsd_grad_bug.py](https://github.com/user-attachments/files/31522137/df_ccsd_grad_bug.py) (attached, ~45 lines, runs in ~2 minutes). It compares pyscf's analytic
gradient against central finite differences of **pyscf's own energy**, so no external reference is
involved and no convention mismatch is possible. The conventional-ERI case is a control that
validates the finite-difference harness.

H2O / cc-pVDZ, h = 1e-3 Bohr, `mf.conv_tol=1e-12`, `mycc.conv_tol=1e-10`:

| reference | max abs(analytic - FD) | |
|---|---|---|
| conventional ERIs (control) | 2.1e-07 Ha/Bohr | 0.01 meV/Ang — finite-difference noise |
| **`density_fit()`** | **3.1e-05 Ha/Bohr** | **1.60 meV/Ang — 147x the control** |

The control agreeing to FD noise establishes that the harness, step size and convergence settings
are adequate, so the DF discrepancy is not an artefact of the comparison.

## Where it comes from

`pyscf/grad/ccsd.py`, in `grad_elec`:

```python
106:  eri0 = mol.intor('int2e', aosym='s2kl', shls_slice=shls_slice)
110:  eri1 = mol.intor('int2e_ip1', comp=3, aosym='s2kl', ...)
```

These are unconditional. `grep -c 'with_df\|density_fit'` over the executable paths of
`grad/ccsd.py` and `grad/ccsd_t.py` returns nothing relevant (the one `with_df` mention at line 208
is a docstring note about scanner side effects), and there are no `warn` or `NotImplementedError`
occurrences in the module. So a DF reference is accepted and silently treated as if the ERIs were
exact: the code returns the conventional-ERI gradient evaluated at DF amplitudes.

## Impact

Measured DF error, same protocol:

| system | nbf | max abs(analytic - FD) |
|---|---|---|
| H2O | 24 | 1.60 meV/Ang (0.059 mHa/Ang) |
| water dimer | 48 | 1.91 meV/Ang (0.070 mHa/Ang) |

So over this range the error is roughly size-insensitive (1.19x for 2x nbf); we are not claiming a
growth trend from two points. In absolute terms it is modest -- below, for example, a 0.2 mHa/Ang
force-accuracy target used for machine-learned force fields.

The reason we think it still warrants a fix is that it is **systematic and silent**:

- It is ~150x the finite-difference noise floor, so it is not lost in numerics, and being systematic
  it does not average out over a trajectory or a dataset.
- The analytic gradient is inconsistent with the energy being minimised, so geometry optimisations
  converge to a point that is not a stationary point of the DF energy surface.
- Most practically for us: **pyscf cannot be used as a reference to validate another DF gradient
  implementation.** We were developing a DF-aware CCSD(T) gradient and spent significant time on a
  phantom discrepancy before establishing that the pyscf side was the inconsistent one. A correct DF
  implementation agrees with finite differences to ~0.001 meV/Ang, so a 1.6 meV/Ang reference error
  is three orders of magnitude above the level at which such work is validated.

## Suggested resolutions, in increasing order of effort

1. **Raise `NotImplementedError`** when `getattr(mycc._scf, 'with_df', None) is not None`. 
2. **Warn loudly** if raising is considered too disruptive for existing users.
3. **Implement the DF path**: contract the 2-PDM through the three-index DF tensors and add the
   auxiliary-basis derivative terms (`int3c2e_ip1`, `int3c2e_ip2`, `int2c2e_ip1`, plus the
   derivative of the fitting metric).

JustinKyleKirkland:
Fixes #3425.

## The bug

`dfccsd.RCCSD` inherits `nuc_grad_method` from `ccsd.CCSD`, so `cc.CCSD(mf.density_fit())` returned `pyscf.grad.ccsd.Gradients`. That routine contracts the relaxed 2-particle density matrix with the exact four-index ERI derivatives, so the result is not a derivative of the density-fitted energy the calculation minimised. No warning, no error.

## The fix

`pyscf/df/grad/ccsd.py` mirrors `pyscf/grad/ccsd.py`, following the pattern already used by `pyscf/df/grad/casscf.py`. Two substitutions:

1. The `int2e`/`int2e_ip1` loop is replaced by a DF contraction. It solves `(P|Q) d_Quv = (P|kl) Gamma_uv,kl` in the AO pair basis, then contracts `d` against `int3c2e_ip1` (orbital-basis derivative), `int3c2e_ip2` (auxiliary-basis derivative) and `int2c2e_ip1` (fitting-metric response). It returns both the gradient term and the `Imat` intermediate.
2. The separable `vhf1` term goes through `mf_grad.get_jk(mol, (hf_dm1, dm1p))`. The hand-rolled per-atom `vhf1` contraction is algebraically identical to `-(veff[dm1p]:hf_dm1 + veff[hf_dm1]:dm1p)` over the atom's AO slice, and routing it through `get_jk` picks up the DF auxiliary-basis response as `0.5*(vaux[0,1] + vaux[1,0])`.

`pyscf/df/grad/ccsd_t.py` feeds the (T) relaxed densities into the same assembly.

DF-UCCSD has the identical bug and only the restricted path is implemented here, so `dfuccsd.UCCSD.nuc_grad_method` now raises `NotImplementedError` rather than staying silently wrong. `pyscf.grad.ccsd.grad_elec` warns when handed a density-fitted reference, which covers callers that build `grad.ccsd.Gradients` directly.

## Benchmark

H2O / cc-pVDZ, `mf.conv_tol=1e-12`, `mycc.conv_tol=1e-10`, central differences with h = 1e-3 Bohr against pySCF's own energy, exactly the protocol in the issue.

| reference | assembly | max abs(analytic - FD) | |
|---|---|---|---|
| conventional ERIs | `grad.ccsd` | 1.40e-07 Ha/Bohr | 0.0072 meV/Ang, FD noise floor |
| `density_fit()` | `grad.ccsd` (before) | 3.11e-05 Ha/Bohr | 1.599 meV/Ang |
| `density_fit()` | `df.grad.ccsd` (after) | 1.39e-07 Ha/Bohr | 0.0071 meV/Ang |

The 3.11e-05 row reproduces the reporter's 3.1e-05 / 1.60 meV/Ang. The new path lands on the same noise floor as the conventional-ERI control.

## Tests

New `pyscf/df/test/test_df_ccsd_grad.py`, 6 tests, 8.6 s: finite-difference checks for CCSD, frozen-core CCSD and CCSD(T); a dispatch check; a check that the DF and conventional assemblies actually differ; and the DF-UCCSD guard.

Also verified by hand and not in the test file, to keep it fast: `as_scanner` across geometries (including that `with_df.reset` clears the stale `auxmol`), `mol.symmetry`, `auxbasis_response = False`, and `only_dfj=True` (1.26e-07 against FD).

`grad/test/test_ccsd.py`, `grad/test/test_ccsd_t.py`, `cc/test/test_dfccsd.py`, `cc/test/test_dfuccsd.py` and `df/test/test_df_grad.py` all still pass.

## Not addressed

Two holes of the same kind, left out to keep this focused. `ci/__init__.py` warns and falls back to conventional CISD on a DF reference, so `grad/cisd.py` differentiates the wrong functional in its HF part. And `mp2.RMP2(df_mf)` built directly, rather than through `mp.MP2` which correctly returns the already-guarded `DFMP2`, has the same problem. Happy to fold either in if preferred.

### DIFF
--- a/pyscf/df/grad/ccsd.py
+++ b/pyscf/df/grad/ccsd.py
@@ -1,0 +1,318 @@
+#!/usr/bin/env python
+# Copyright 2014-2025 The PySCF Developers. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+'''
+Density-fitting CCSD analytical nuclear gradients
+
+Copied from pyscf.grad.ccsd.py.  The four-index ERI derivatives are replaced by
+the derivatives of the three-index integrals and of the fitting metric, so that
+the gradient differentiates the DF energy expression that the calculation
+actually minimised.
+'''
+
+from functools import reduce
+import numpy
+from pyscf import lib
+from pyscf import df
+from pyscf.lib import logger
+from pyscf.cc import ccsd_rdm
+from pyscf.ao2mo.outcore import balance_partition
+from pyscf.grad import rhf as rhf_grad
+from pyscf.grad import ccsd as ccsd_grad
+from pyscf.grad.mp2 import _index_frozen_active, has_frozen_orbitals
+from pyscf.df.grad.rhf import _int3c_wrapper, _gen_metric_solver
+
+
+def _auxmol(mycc):
+    with_df = getattr(mycc, 'with_df', None)
+    if with_df is None:
+        with_df = mycc._scf.with_df
+    auxmol = with_df.auxmol
+    if auxmol is None:
+        auxmol = df.addons.make_auxmol(with_df.mol, with_df.auxbasis)
+    return with_df, auxmol
+
+def _pair_weights(nao):
+    # Undo the double counting of the off-diagonal elements of a symmetric
+    # matrix stored in the lower-triangular packed layout.
+    w = numpy.ones(nao*(nao+1)//2)
+    diag = numpy.arange(nao)
+    w[diag*(diag+1)//2 + diag] = .5
+    return w
+
+def solve_df_rdm2(cc_grad, fdm2, max_memory=None):
+    '''Solve (P|Q) d_Quv = (P|kl) Gamma_uv,kl for the AO-basis DF 2-PDM.
+
+    Args:
+        cc_grad : DF-CCSD gradients method object
+        fdm2 : HDF5 group holding the AO 2-PDM 'dm2' written by
+            :func:`pyscf.grad.ccsd._rdm2_mo2ao`
+
+    Returns:
+        (int3c, dferi, dfdm2), each of shape (naux, nao_pair) in the
+        lower-triangular AO pair layout:
+            int3c[P,uv] = (P|uv)
+            dferi[P,uv] = (P|Q)^-1 (Q|uv)
+            dfdm2[P,uv] = (P|Q)^-1 (Q|kl) Gamma_uv,kl
+    '''
+    mol = cc_grad.mol
+    auxmol = _auxmol(cc_grad.base)[1]
+    nao, nbas = mol.nao, mol.nbas
+    naux = auxmol.nao
+    nao_pair = nao * (nao+1) // 2
+    aux_loc = auxmol.ao_loc
+    if max_memory is None:
+        max_memory = cc_grad.max_memory
+
+    int3c = numpy.empty((naux, nao_pair))
+    get_int3c = _int3c_wrapper(mol, auxmol, 'int3c2e', 's2ij')
+    blksize = max(1, min(int(max_memory*.2e6/8/max(nao_pair, 1)), naux))
+    for shl0, shl1, nL in balance_partition(aux_loc, blksize):
+        p0, p1 = aux_loc[shl0], aux_loc[shl1]
+        int3c[p0:p1] = get_int3c((0, nbas, 0, nbas, shl0, shl1)).T
+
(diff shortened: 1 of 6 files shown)


pyscf/df/grad/ccsd.py
```diff
@@ -0,0 +1,318 @@
...
+        dm1mo[OA[:,None],OF] = dco.T
+        dm1mo[VA[:,None],VA] = dvv + dvv.T
+        dm1mo[VF[:,None],VA] = dfv
+        dm1mo[VA[:,None],VF] = dfv.T
+    else:
+        dm1mo[:nocc,:nocc] = doo + doo.T
+        dm1mo[nocc:,nocc:] = dvv + dvv.T
+
+    dm1 = reduce(numpy.dot, (mo_coeff, dm1mo, mo_coeff.T))
+    vhf = mycc._scf.get_veff(mycc.mol, dm1) * 2
```
sunqm, line 227: ensure `mycc._scf._scf` is an instance of density fitting SCF method

sunqm (approved):

pyscf/df/grad/ccsd.py
JustinKyleKirkland: Yeah, nice catch, thanks! I fixed in aac556740.

Issue #3339, opened 2026-07-18
[BUG REPORT] Analytic meta-GGA Hessian overestimates heavy-atom force constants (M06-2X C-Br stretch off by +220 cm^-1, OCBr2 modes off by up to +674 cm^-1)


MaroGol:
## Summary

The analytic Hessian for tau-dependent (meta-GGA) functionals produces force constants that are systematically too large for modes with heavy-atom amplitude.
A central-finite-difference Hessian built from the analytic gradients of the same mf object gives correct results (matching ORCA 6.0 and Gaussian 16 to ~1 cm^-1), so the analytic gradients are fine and the error is
confined to the analytic second derivatives.

## Reproduction

CH3Br, M06-2X/def2-TZVP, RKS, grids.level=6, conv_tol=1e-12
(script attached: repro_mgga_hessian_br.py; `--cpu` runs it on CPU PySCF, default is GPU4PySCF).

Sorted harmonic frequencies (cm^-1):

| source | C-Br stretch | remaining modes |
|---|---|---|
| analytic Hessian, CPU pyscf 2.13.1 release | **834.5** | match FD to ~5 cm^-1 |
| analytic Hessian, CPU pyscf master @ 7e665e9 | **834.5** | match FD to ~5 cm^-1 |
| analytic Hessian, gpu4pyscf 1.7.6 (also 1.7.4) | **834.5** | match FD to ~5 cm^-1 |
| finite-diff of analytic gradients (all versions) | 614.5 | - |
| ORCA 6.0 (TightSCF DEFGRID3) | 612.9 | agree with FD |
| Gaussian 16 (Int=UltraFine) | 618.8 | agree with FD |

Only the mode with Br amplitude is wrong. The other eight agree between all methods and both external codes.

## Scaling behavior
FD means Finite Difference

- Grows with atomic number: H2O fine, SO2 +14 cm^-1 on the S-dominated modes, CH3Br +220 on the C-Br stretch, OCBr2 (dibromophosgene) all five Br-involving modes 1.6-3x too stiff (freq MAE 336 cm^-1 vs FD/ORCA/Gaussian 0.8) while its C=O stretch is correct to ~1 cm^-1.
[...]

PR #3428, opened 2026-08-28
Harden the finite difference driver and use it for Hessians

JustinKyleKirkland:
Reworked after @jeanwsr's suggestion. This started as a separate `pyscf/hessian/numeric.py`; I had missed that `pyscf/tools/finite_diff.py` (#2774) already does this. It does, so the whole thing now lives there and the new module is gone.

## Why

`tools/finite_diff.py` already differences analytic gradients into a Hessian, which is the only route to a Hessian for the many methods with no analytic second derivative: TDHF and TDDFT, MP2, CISD, CCSD, CCSD(T), CASSCF. Four things kept it from being usable for that in practice, all measured on H2O/STO-3G RHF against the analytic Hessian.

**Displacement.** The 1e-2 default is dominated by truncation.

| displacement | Hessian | gradients |
|---|---|---|
| 1e-2 | 5.51e-05 | 1.47e-05 |
| 5e-3 | 1.37e-05 | 3.67e-06 |
| 1e-3 | 9.10e-07 | 1.47e-07 |

Both cases improve, so the shared default moves to 1e-3. Below that the Hessian starts picking up gradient noise, so this is the crossover rather than just "smaller is better".

**Convergence.** The Hessian guard was `conv_tol > displacement**3`. At displacement 1e-3 that is `1e-9 > 1e-9`, false, so it stays silent on exactly the pyscf default that is 500x wrong:

| SCF conv_tol | Hessian error at 1e-3 |
|---|---|
| 1e-9 (default) | 3.41e-04 |
| 1e-11 | 5.84e-05 |
| 1e-12 | 2.42e-06 |
| 1e-14 | 7.08e-07 |

Differencing amplifies error by `1/(2*displacement)`, so each check needs one more power of the displacement. The mean field is now found by walking `_scf`/`base`, so the check also reaches post-SCF and excited-state methods rather than only the object passed in.

**Symmetry.** Point group detection has a finite tolerance that the displacement falls inside, so a displaced C2v molecule is still assigned C2v and the wavefunction is symmetry-adapted to a group the geometry has lost.

| displaced-geometry handling | Hessian error |
|---|---|
| no symmetry, baseline | 9.10e-07 |
| symmetry on, as before | 7.79e-06 |
| symmetry on, pinned `'C1'` | 4.98e-07 |

Rebuilding with `symmetry=False` is not an option, since a symmetry-adapted SCF class rejects a symmetry-free Mole. The displaced Mole is now also built out of place rather than mutating the caller's, and in Bohr, which stops `set_geom_` announcing a unit change on every displacement.

**Grids.** The RKS/PBE0 error against the analytic Hessian is independent of the displacement (3.58e-04, 3.71e-04, 3.73e-04 at 5e-3, 1e-3, 5e-4) and `grid_response` on the gradient does not help. It is grid incompleteness: 3.71e-04 on the default pruned level-3 grid, 9.36e-06 at level 5 unpruned. It exceeds the finite difference error itself, so it is worth a warning.

Also symmetrises the Hessian, since the exact one is symmetric and the differencing error is not, and warns when an excited-state root stops being the best amplitude match under displacement, the root flip `grad/tdrhf.py` already carries a TODO about.

## One bug outside the module

`grad/tdrhf.py` line 237: the gradient scanner tested `converged[self.state]` with a 1-based `state`, so it read the convergence flag of the *next* root and would `IndexError` on the last one. Line 230 just above already uses `state-1`. The driver relies on that flag to refuse an unconverged displacement, which is how I hit it. Happy to split it out if you would rather it went separately.

## Tests

Six added to `pyscf/tools/test/test_finite_diff.py`, 13 total, ~17 s: Hessians against the analytic RHF and RKS ones, symmetry pinning, the symmetrisation, a TDDFT Hessian with stored frequencies plus a check that the scanner really pins its root, and the warnings firing. All seven existing tests pass unchanged. `examples/tools/07-finite_difference.py` gains the TDDFT case.

## Note on the default change

Moving `displacement` from 1e-2 to 1e-3 changes results for anyone relying on the default. It is a 60x accuracy improvement for Hessians and 100x for gradients, but say the word if you would rather keep 1e-2 and only document the recommendation.

### DIFF
--- a/examples/tools/07-finite_difference.py
+++ b/examples/tools/07-finite_difference.py
@@ -37,6 +37,21 @@ print('Analytical Hessian:')
 print(mol.RHF().run().Hessian().kernel())
 
 #
+# Hessian for a method that has an analytic gradient but no analytic second
+# derivative. TDDFT is the common case: excited-state vibrational frequencies
+# are only reachable this way.
+#
+mf = mol.RKS(xc='pbe0')
+mf.grids.level = 5      # a Hessian needs a finer grid than the default
+mf.grids.prune = None
+mf.run(conv_tol=1e-13)
+td = mf.TDA().run(nstates=5)
+# Build the scanner to pin the root being differentiated
+H = finite_diff.kernel(td.nuc_grad_method().as_scanner(state=1))
+print('Finite difference Hessian of the first excited state:')
+print(H)
+
+#
 # Finite difference Gradients as a PySCF builtin Gradients object
 #
 mf = mol.RHF()
--- a/pyscf/grad/tdrhf.py
+++ b/pyscf/grad/tdrhf.py
@@ -234,7 +234,7 @@ class TDSCF_GradScanner(lib.GradScanner):
     def converged(self):
         td_scanner = self.base
         return all((td_scanner._scf.converged,
-                    td_scanner.converged[self.state]))
+                    td_scanner.converged[self.state-1]))
 
 
 class Gradients(rhf_grad.GradientsBase):
--- a/pyscf/tools/finite_diff.py
+++ b/pyscf/tools/finite_diff.py
@@ -15,6 +15,16 @@
 
 '''
 Finite difference driver
+
+Differencing energies gives gradients, differencing analytic gradients gives
+Hessians.  The latter is the only route to a Hessian for the many methods that
+have an analytic gradient but no analytic second derivative: TDHF and TDDFT,
+MP2, CISD, CCSD, CCSD(T) and CASSCF among them.
+
+Excited states are differentiated at a fixed root.  Build the gradient scanner
+to pin the state and hand it over:
+
+    >>> H = finite_diff.kernel(td.nuc_grad_method().as_scanner(state=2))
 '''
 
 import numpy as np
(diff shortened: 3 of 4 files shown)


JustinKyleKirkland:
Error unrelated to my changes.

jeanwsr:
What about implement this in pyscf/tools/finite_diff.py

JustinKyleKirkland:
Good call, thank you. I had missed `pyscf/tools/finite_diff.py` entirely when I looked for prior art, so I went and built a parallel module in `pyscf/hessian/`. That is now deleted and everything lives in `tools/finite_diff.py`.

While folding it in I found that the four issues I had solved separately all apply to the existing driver too, so the PR is now a set of fixes to it rather than new code. Measured on H2O/STO-3G RHF against the analytic Hessian:

- the 1e-2 default displacement gives 5.5e-05, against 9.1e-07 at 1e-3
- the `conv_tol > displacement**3` guard is exactly non-triggering at the pyscf default conv_tol of 1e-9, which is where the Hessian is off by 3.4e-04
- a displaced C2v molecule is still detected as C2v, so it is symmetry-adapted to a group it no longer has: 7.8e-06, against 5.0e-07 with the group pinned to C1
- the DFT grid error is independent of displacement and larger than the finite difference error, 3.7e-04 on the default grid against 9.4e-06 at level 5 unpruned

Details and the full tables are in the updated description. One thing worth your attention separately: `grad/tdrhf.py:237` indexes `converged[self.state]` with a 1-based state, so it reads the next root's flag and would IndexError on the last one. Fixed here since the driver depends on it, but happy to split it out.

jeanwsr:
> the DFT grid error is independent of displacement and larger than the finite difference error, 3.7e-04 on the default grid against 9.4e-06 at level 5 unpruned

Is this TDDFT hessian or ground state DFT hessian?

JustinKyleKirkland:
> > the DFT grid error is independent of displacement and larger than the finite difference error, 3.7e-04 on the default grid against 9.4e-06 at level 5 unpruned
> 
> Is this TDDFT hessian or ground state DFT hessian?

That's the ground state. It's H2O/STO-3G RKS/PBE0: analytic RKS gradients differenced into a Hessian. Compared against the analytic RKS Hessian.

One thing I should flag on the back of your question: test_tddft_hessian runs on the default grid and asserts stored values, so it's a regression test rather than a grid-converged reference, which sits awkwardly next to the recommendation. Happy to move it to level 5 unpruned and re-store the frequencies. I can also measure how far the TDA/PBE0 Hessian shifts between the default grid and level 5 unpruned and add it to the description (that's the nearest excited-state analogue to the number you asked about).

jeanwsr:
The analytic hessian has no grid response (see #3339), so there's no way to test numeric hessian against it precisely. There are some tests (https://github.com/pyscf/pyscf/blob/master/pyscf/hessian/test/test_rks.py), but with low precision.

susilehtola:
FWIW grid response and other things are now trivial to add in PySCF, see https://arxiv.org/abs/2608.26440 and https://github.com/susilehtola/libxckernel

Issue #3431, opened 2026-09-02
[BUG REPORT] L-PDFT StateAverageMix CI vectors use the wrong rotation convention


JangidBhavnesh:
### Describe the bug

For L-PDFT, `mc.ci` contain CI vectors in the eigenbasis of the L-PDFT Hamiltonian. Reconstructing the Hamiltonian with these vectors should therefore yield `diag(mc.e_states)`.  This holds for ordinary same-spin state averaging, but fails for StateAverageMix: the returned CI vectors are rotated with si_pdft instead of si_pdft.T.

### To Reproduce
```python
import numpy as np
from scipy.linalg import block_diag
from pyscf import fci, gto, mcpdft, scf

mol = gto.M( atom="Li 0 0 0; H 1.5 0 0",
    basis="6-31G", verbose=4,
)
mf = scf.RHF(mol).run()

def spin_solver(ss):
    solver = fci.direct_spin1.FCI(mol)
    solver.spin = ss
    solver = fci.addons.fix_spin(solver, ss=ss)
    solver.nroots = 2
    return solver

def reconstruct_mixed_ham(mc):
    return block_diag(*mc.make_lpdft_ham_(ci=mc.ci))

# Two singlet and two triplet roots are needed: each spin block must have
# dimension > 1 for the transpose error to be visible.
mixed = mcpdft.CASSCF(mf, "tPBE", 5, 2, )
mixed = mixed.multi_state_mix([spin_solver(0), spin_solver(2)],
                            weights=[0.25] * 4,
                            method="lin",).run()

h_expected = np.diag(mixed.e_states)
h_reconstructed = reconstruct_mixed_ham(mixed)

diff = h_expected - h_reconstructed

print(np.linalg.norm(diff)) 
```

On the other hand, if you try to reconstruct the Hamiltonian using the ordinary same-spin state averaging then it reconstructs correctly:

```python
import numpy as np
from pyscf import fci, gto, mcpdft, scf

mol = gto.M( atom="Li 0 0 0; H 1.5 0 0",
    basis="6-31G", verbose=4,
)
mf = scf.RHF(mol).run()

same_spin = mcpdft.CASSCF(mf, "ftLDA,VWN3", 5, 2, grids_level=1)
same_spin.fix_spin_(ss=0)
same_spin = same_spin.multi_state([0.5, 0.5], method="lin").run()

h_expected = np.diag(same_spin.e_states)
[...]

PR #3432, opened 2026-09-02
Fixing the LPDFT CI vector for the State-average mix solvers

JangidBhavnesh:
Addressing #3431 

Basically adding the transpose which was missing in the L-PDFT CI vector construction. Added unit test to make sure this is checked.

@MatthewRHermes 

Thanks.


MatthewRHermes (approved):
tyvm!

PR #3435, opened 2026-09-06
Give the DF gradient auxiliary basis the range-separation parameter of mol

### LINKED ISSUE #3434: [BUG REPORT] `df.grad` honours `mol.with_range_coulomb` for the three-centre integrals but not for the fitting metric
**Version:** pyscf 2.14.0
**Module:** `pyscf/df/grad/rhf.py` (`get_jk` / `get_k`)

## Summary

Inside a `with mol.with_range_coulomb(omega)` block, the density-fitted gradient's `get_k` returns a
matrix that matches **neither** the attenuated nor the unattenuated exchange derivative. The
three-centre integrals `(P|munu)` are rebuilt from `mol` and do see the attenuation, while the
fitting metric `(P|Q)` is built from a separate `auxmol` that does not, so the two halves of the
fitted operator disagree. No warning is raised.

The `omega=` keyword argument is handled correctly, so a correct route exists; the problem is that
the context manager silently produces a third, wrong answer instead of either working or refusing.

## Reproducer

```python
import numpy as np
from pyscf import df, gto, scf

mol = gto.M(atom="O 0 0 0; H 0 0.757 0.587; H 0 -0.757 0.587", basis="def2-svp", verbose=0)
w = 0.3

mf = scf.RHF(mol).density_fit()
mf.kernel()
g = mf.nuc_grad_method()
dms = np.stack([mf.make_rdm1()])

k_arg = np.asarray(g.get_k(mol, dms, omega=w))        # attenuated, via the keyword
k_full = np.asarray(g.get_k(mol, dms))                # full range
with mol.with_range_coulomb(w):                       # attenuated, via the context manager
    k_ctx = np.asarray(g.get_k(mol, dms))

print("|ctx - arg| ", np.abs(k_ctx - k_arg).max())    # 3.031e-01
print("|ctx - full|", np.abs(k_ctx - k_full).max())   # 1.270e+01

# The same comparison on the conventional (non-DF) gradient agrees exactly:
mf2 = scf.RHF(mol)
mf2.kernel()
g2 = mf2.nuc_grad_method()
d2 = np.stack([mf2.make_rdm1()])
a2 = np.asarray(g2.get_k(mol, d2, omega=w))
with mol.with_range_coulomb(w):
    c2 = np.asarray(g2.get_k(mol, d2))
print("direct path |ctx - arg|", np.abs(c2 - a2).max())   # 0.000e+00
```

Output on pyscf 2.14.0:

```
|ctx - arg|  3.031e-01
|ctx - full| 1.270e+01
direct path |ctx - arg| 0.000e+00
```

So the contract holds on the direct path and breaks on the DF path.

## Mechanism

The attenuation reaches one factor of the fitted operator and not the other. Measured on the
integrals the DF **gradient** itself builds:

```python
auxmol = df.addons.make_auxmol(mol, "def2-universal-jkfit")

ip1 = df.incore.aux_e2(mol, auxmol, intor="int3c2e_ip1", aosym="s1", comp=3)
with mol.with_range_coulomb(w):
    ip1_ctx = df.incore.aux_e2(mol, auxmol, intor="int3c2e_ip1", aosym="s1", comp=3)

m1 = auxmol.intor("int2c2e_ip1", aosym="s1", comp=3)
with mol.with_range_coulomb(w):
    m1_molctx = auxmol.intor("int2c2e_ip1", aosym="s1", comp=3)   # mol's context: no effect
with auxmol.with_range_coulomb(w):
    m1_auxctx = auxmol.intor("int2c2e_ip1", aosym="s1", comp=3)   # auxmol's own: takes effect
```

```
int3c2e_ip1 under mol's context:     4.857e+00   (scale 1.096e+01)   -- attenuated
int3c2e_ip2 under mol's context:     3.593e+00   (scale 3.602e+00)   -- attenuated
int2c2e_ip1 under mol's context:     0.000e+00   (scale 1.305e+01)   -- NOT attenuated
int2c2e_ip1 under auxmol's context:  6.886e+00                       -- would be, if asked
```

`mol.with_range_coulomb` sets the range-omega in `mol._env`, which `auxmol` does not share, and the
DF gradient's metric comes from `auxmol`. So the derivative of the fitted exchange is assembled from
attenuated three-centre derivatives contracted against an unattenuated metric derivative, which is
the derivative of no operator at all. The same asymmetry holds for the plain `int3c2e` / `int2c2e`
pair, so it is not specific to the derivative integrals.

## Impact

Silent wrongness in a call that looks supported. The context manager is the documented way to
attenuate the Coulomb operator, it works as expected on the conventional path, and on this path it
returns a matrix that is the derivative of no operator. Nothing warns, and the result is plausible
in magnitude, so the only way to notice is an independent reference: a finite difference, or the
`omega=` route for comparison.

The discrepancy is not marginal. In the reproducer above the context result differs from the
correct attenuated one by 0.303 against a term of magnitude 1.04, and for a range-separated hybrid
the resulting error in the long-range exchange contribution to a nuclear gradient is several
percent.

## Suggested fix

Either propagate the range-omega to `auxmol` where the DF gradient builds its metric, so the
context manager and the `omega=` argument agree as they do on the direct path, or detect a nonzero
`mol` range-omega in `df/grad/rhf.py` and raise, directing the caller to `omega=`.

## Note on the SCF paths

For completeness, since the numbers invite the question: under the same context manager both the DF
and the conventional **SCF** `get_k` return the full-range result exactly, because each holds
integrals built at full range (`with_df._cderi`, `mf._eri`) that the context manager cannot
invalidate:

```
[SCF-conventional]  |ctx - arg| 9.111e+00   |ctx - full| 1.776e-15
[SCF-density-fitted]|ctx - arg| 9.111e+00   |ctx - full| 0.000e+00
```

That is a separate, arguably-expected caching effect, and at least it is self-consistent. The
gradient is different in kind: it rebuilds its integrals on the fly and attenuates only one of the
two factors, so its answer corresponds to no operator at all.

JustinKyleKirkland:
Fixes #3434.

In `df/grad/rhf.py` the three-center integrals are evaluated in the concatenated environment of `mol` and `auxmol` (`_int3c_wrapper`), so they pick up omega from `mol`, while the metric `(P|Q)` is evaluated with `auxmol` alone. Inside a `mol.with_range_coulomb(w)` context only one factor of the fitted operator gets attenuated, and the result is the derivative of no actual operator. The `omega=` keyword was never affected: it goes through `with_df.range_coulomb`, which sets omega on both molecules.

`auxmol` is now resolved through a helper that copies omega over from `mol` when the two disagree, which is what `make_auxmol` already does when it copies `mol._env[:PTR_ENV_START]`. The copy only happens on a mismatch, so the `omega=` path is untouched.

On the reproducer from the issue, `max|ctx - arg|` goes from 3.03e-01 to 3.9e-16, and the auxbasis response term from 4.19e-04 to 0. Added a regression test; `test_df_grad`, `test_df_hessian` and `test_df_jk` pass.

Not addressed here: `get_j` with a nonzero omega raises `LinAlgError` out of `scipy.linalg.solve(int2c, rhoj.T, assume_a='pos')` on the long-range metric. That happens on master through the `omega=` path as well, so it is a separate issue.


jeanwsr:
The get_j might need something similar to get_jk's `_cho_solve_rhojk` to prevent LinAlgError. 

Also, there might be same missing-omega bug in df/hessian.

PR #3444, opened 2026-09-11
Optimization of SA-CASSCF Nuclear Gradients

JangidBhavnesh:
In this PR, I have optimized the `SA-CASSCF` nuclear gradient computation, mainly the `CP-MCSCF` procedure. Both for the non-DF and DF versions. 

The main changes are as follows:

- Currently, the `CP-MCSCF` equation, i.e. `Hx = -G` is solved with `scipy.linalg.cg` and for logging, the current code does unnecessary Hessian-vector calls. In the modified code, logging info is extracted from the `CG` callback. For `niterations,` it saved `n-Hessian-vector` calls.

- Additionally, at various stages of gradient computation,  typical `MCSCF` `eris` is not passed, because of that it was recomputing the integral on the fly, in the modified code, I have passed the `eris`.

No new unit test or example added.

### $\mathrm N_2 \text{SA(2)-CASSCF, (10e,8o)}$
$$
\begin{array}{c|cc}
\text{Basis} & \text{Non-DF speedup} & \text{DF speedup} \\
\hline
\mathrm{cc\text{-}pVDZ} & 1.6\times & 4.78\times \\
\mathrm{cc\text{-}pVTZ} & 1.2\times & 4.13\times \\
\mathrm{cc\text{-}pVQZ} & 1.5\times & 5.1\times
\end{array}
$$

### $\mathrm {Fe(CO)}_5 \text{SA(10)-CASSCF, (10e,10o) def2-TZVP and DF variant only}$
$$
\begin{array}{c|c}
\text{Implementation} & \text{Gradient time} \\
\hline
\text{Previous} & 430\ \mathrm{s} \\
\text{Optimized} & 155\ \mathrm{s} \\
\text{Speedup} & 2.77\times
\end{array}
$$


JangidBhavnesh:
@MatthewRHermes I know you said keep optimization-3, separately, but in this code, I have made sure previous implementation is as it is and available at higher verbosity. The other methods based on the SA-CAS would automatically uses the previous implementation if they don't have the combined function as I defined for the `SA-CAS`.

Can you please have a look?, and even after that if not convinced, I will make this into two separate PRs.

pyscf/grad/sacasscf.py
```diff
@@ -685,6 +927,39 @@ def get_LdotJnuc (self, Lvec, state=None, atmlst=None, verbose=None, mo=None, ci
...
+
+        de = Lorb_Lci_dot_dgorb_dgci_dx(
+            Lorb, Lci, self.weights, self.base, mo_coeff=mo, ci=ci,
+            atmlst=atmlst, mf_grad=mf_grad, eris=eris, verbose=verbose,
+            fcasscf=fcasscf, ci_state=ci[state])
+        de += self.grad_nuc(atmlst=atmlst)
+        if self.mol.symmetry:
+            de = self.symmetrize(de, atmlst)
+        return de
+
```
MatthewRHermes, line 1045: So SA-CASSCF, and only SA-CASSCF, has this different, separate function which we have to maintain separately from everything else? This is precisely the thing that scares me.

JangidBhavnesh: I have created the same hook for the all the PDFT variants (MC-PDFT, L-PDFT, and MS-PDFT) which depends on this. Those hooks are placeholder in current PR. In [upcoming PR](https://github.com/JangidBhavnesh/pyscf/pull/10): I have added the same optimization for the all the PDFT variants.

MatthewRHermes (changes requested):
I'm sorry, but I just don't think it's responsible to put your `get_nuc_response` refactor into the codebase in its current state. The hierarchical dependencies SA-CASSCF -> MC-PDFT -> MS-PDFT -> L-PDFT are safety and readability features that ensure that any modification to SA-CASSCF must play nice with all child classes. This PR goes to some length to cordon off its modifications to SA-CASSCF alone, disabling those safety features (using code that scrambles the natural Python inheritance logic) in exchange for a 2-3x speedup. *This is not a good exchange.* We maintainers will be responsible for this code for many years, and in the Python ecosystem, routine updates to things like scipy and numpy frequently oblige us to make minor edits to everything. Under these circumstances, reusing code between methods is worth the performance hit. This is still true if you assume that LLMs will be doing everything in a few weeks, since LLMs still have token budgets and finite context windows.

JangidBhavnesh:
> I'm sorry, but I just don't think it's responsible to put your `get_nuc_response` refactor into the codebase in its current state. The hierarchical dependencies SA-CASSCF -> MC-PDFT -> MS-PDFT -> L-PDFT are safety and readability features that ensure that any modification to SA-CASSCF must play nice with all child classes. This PR goes to some length to cordon off its modifications to SA-CASSCF alone, disabling those safety features (using code that scrambles the natural Python inheritance logic) in exchange for a 2-3x speedup. _This is not a good exchange._ We maintainers will be responsible for this code for many years, and in the Python ecosystem, routine updates to things like scipy and numpy frequently oblige us to make minor edits to everything. Under these circumstances, reusing code between methods is worth the performance hit. This is still true if you assume that LLMs will be doing everything in a few weeks, since LLMs still have token budgets and finite context windows.

Thanks for the comment.

I have removed the `get_nuc_response` interface and overrides from SA-CASSCF, the PDFT methods, and NAC, including the DF variants. I also removed the combined-response functions.

The `SA-CAS` gradient now uses the `get_ham_response` and `get_LdotJnuc` methods through normal inheritance.

Note that the custom `CG` implementation is in the shared `lagrange` solver and applies to the derived methods as well. I will make another PR by passing ERIs, and benchmarking.

MatthewRHermes (approved):
Thanks. I'm willing to consider the `get_nuc_response` type of thing in the future if you can make it something a little more general. For instance, if you just combine the orb and ci parts of the LdotJnuc, that will probably inherit down cleanly.

PR #3448, opened 2026-09-15
Fix missing orbital-rotation terms in the ROHF Hessian

wtpeter:
`gen_g_hop_rohf` previously reused the UHF occupied–virtual Hessian blocks. However, ROHF shares spatial orbitals between spins: a rotation that is occupied–virtual for one spin can be occupied–occupied or virtual–virtual for the other. The UHF construction omits the corresponding Fock contributions.

This PR replaces the UHF-block construction with the complete orbital-gradient response, using `fock @ kappa - kappa @ fock` for each spin and retaining the existing density response.

Add a finite-difference test at random, overlap-orthonormal orbitals. For a random rotation direction x, it compares the central difference of the orbital gradient at `C @ exp(±eps * kappa)` with `h_op(x)`. The previous implementation fails with an error norm of approximately 4.0613.

Validation: 32 tests passed, including the new regression test and existing Newton and stability tests.


jeanwsr (approved):
Looks correct, but pbc/scf/newton_ah.py still has the old UHF-delegation construction, which may need a note.

wtpeter:
Thanks for pointing this out. I've updated the PBC implementation as well and added a finite-difference test. Both new tests pass in CI. The Windows CI failure is in a UCCSD zero-beta-electron test; I'm not sure what caused it. Could you please rerun the failed job?

hczhai:
This PR correctly fixes the missing term in the current ``gen_g_hop_rohf`` Hessian. However, there is one further subtlety in the definition of the ROHF Hessian. The operator computed in this PR differentiates the orbital-gradient components in the moving MO frame. Away from a stationary point, this differs from the Hessian of the fixed exponential parameterization $E(Ce^{\kappa})$ at $\kappa=0$. One can verify that the Hessian operator computed in this PR is not symmetric at a non-converged point:

```python3
# !pip install pyscf==2.14.0
# !wget https://raw.githubusercontent.com/wtpeter/pyscf/d04916e446894b5025673f43ce1591c4c193ec75/pyscf/soscf/newton_ah.py -O newton_ah_pr3448.py

from pyscf import gto, scf
from newton_ah_pr3448 import gen_g_hop_rohf
import numpy as np

mol = gto.M(atom="H 0.3 0.2 0; H 0 -0.2 0.6; H 0 0.8 0.7", basis='6-31g', spin=1, verbose=3)
mf = scf.ROHF(mol)
mf.max_cycle = 1
mf.kernel()

g, h_op, h_diag = gen_g_hop_rohf(mf, mf.mo_coeff, mf.mo_occ)

hess = np.zeros((len(g), len(g)))
for i in range(len(g)):
    v = np.eye(1, len(g), k=i).ravel()
    hess[:, i] = h_op(v)

print('symm error =', np.linalg.norm(hess - hess.T))
# symm error = 0.21079487641440495, should be ~ machine precision
```

The tests used in this PR compute the finite difference as ``(mf.get_grad(cp, occ) - mf.get_grad(cm, occ)) / (2 * eps)``. This finite difference correctly checks the derivative of the gradient components in the moving MO frame. However, it is not necessarily the same as the Hessian of the fixed exponential parameterization (especially for the ROHF case), because the frame changes with the orbitals. The proper definition for the ROHF Hessian is the covariant derivative:

$$
\mathrm{Hessian}[\kappa] = \nabla_\kappa \mathrm{grad}\ E=\lim_{t\rightarrow 0} \frac{\mathcal{T}^{-1}_{t\kappa} (\mathrm{grad}\ E(Ce^{t\kappa})) - \mathrm{grad}\ E(C)}{t}
$$

where $\mathcal{T}^{-1}_{t\kappa}$ is the parallel transport map. This is discussed in [*J. Phys. Chem. A* 2024, 128, 6601−6612](https://doi.org/10.1021/acs.jpca.4c03213):

> For RHF and RKS, this problem is not apparent, because the optimization takes place in a Grassmann manifold, for which the parallel transport map is trivial in the right parametrization (see [eq 23]). This is not the case for the flag manifolds on which ROHF and CASSCF optimization problems are set (see [eq 27]).

Eq. 27 in that paper gives the definition of parallel transport associated with the Levi-Civita connection for the metric used in that paper:

$$
\mathcal{T}_{C\kappa}(C\kappa') = Ce^{\kappa}e^{-\phi_\kappa}(\kappa'),\quad \phi_\kappa(\kappa') = \frac{1}{2} \mathrm{proj}_{\mathbb{k}}([\kappa,\kappa']).
$$

Adding the term induced by $\phi_\kappa$ in the ``h_op`` gives the expected symmetric Hessian:

```python3
g, h_op, h_diag = gen_g_hop_rohf(mf, mf.mo_coeff, mf.mo_occ)

def new_h_op(x):
    gm = scf.hf.unpack_uniq_var(g, mf.mo_occ)
    km = scf.hf.unpack_uniq_var(x, mf.mo_occ)
    return h_op(x) + 0.5 * scf.hf.pack_uniq_var(km @ gm - gm @ km, mf.mo_occ)

new_hess = np.zeros((len(g), len(g)))
for i in range(len(g)):
    v = np.eye(1, len(g), k=i).ravel()
    new_hess[:, i] = new_h_op(v)

print('new symm error =', np.linalg.norm(new_hess - new_hess.T))
# new symm error = 2.987183110301738e-15
```

This correction can also be checked directly against the mixed second derivative of $E(Ce^{s\kappa+t\kappa'})$ at $s=t=0$, without explicitly constructing parallel transport:

```python3
import scipy.linalg

mo, occ = mf.mo_coeff, mf.mo_occ
uk = lambda v: scf.hf.unpack_uniq_var(v, occ)
E  = lambda k: mf.energy_tot(mf.make_rdm1(mo @ scipy.linalg.expm(k), occ))

def fd(x, y, t=1e-4):
    kx, ky = uk(x), uk(y)
    return (E(t*(kx+ky)) - E(t*(kx-ky)) - E(t*(ky-kx)) + E(-t*(kx+ky))) / (4*t**2)

rng = np.random.default_rng(12)
x, y = rng.normal(size=len(g)), rng.normal(size=len(g))
# the factor 1/2 is pyscf's convention for g and h_op
print(fd(x, y) / 2, x @ new_hess @ y, x @ hess @ y)
# -5.209535330896742 -5.209535852153961 -5.15133134040563
```

wtpeter:
@hczhai Thanks for pointing this out🤩. I've added the the correction to both the molecular and PBC implementations. The tests now check symmetry and compare against mixed second derivatives of the energy using a four-point finite difference. Both previous implementations fail these checks; the corrected version passes.

